PR #2, opened 2020-08-07
Added welcome messages

bartgol:
## Motivation

Make EKAT more welcoming when a new user opens a PR/issue.

Also, tested that the pull-request template works as expected, and that the link to the contributing file shows up at the bottom.

### DIFF
--- a/.github/config.yml
+++ b/.github/config.yml
@@ -1,0 +1,18 @@
+# Configuration for new-issue-welcome - https://github.com/behaviorbot/new-issue-welcome
+
+# Comment to be posted to on first time issues
+newIssueWelcomeComment: >
+  Thanks for opening your first issue here! Be sure to follow the issue template!
+
+# Configuration for new-pr-welcome - https://github.com/behaviorbot/new-pr-welcome
+
+# Comment to be posted to on PRs from first time contributors in your repository
+newPRWelcomeComment: >
+  Thanks for opening this pull request! Please check out our contributing guidelines.
+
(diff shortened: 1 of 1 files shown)


.github/config.yml
```diff
@@ -0,0 +1,18 @@
+# Configuration for new-issue-welcome - https://github.com/behaviorbot/new-issue-welcome
+
+# Comment to be posted to on first time issues
+newIssueWelcomeComment: >
+  Thanks for opening your first issue here! Be sure to follow the issue template!
```
jeff-cohere, line 5: Is it practical to hyperlink the template?

.github/config.yml
```diff
@@ -0,0 +1,18 @@
...
+
+# Configuration for new-pr-welcome - https://github.com/behaviorbot/new-pr-welcome
+
+# Comment to be posted to on PRs from first time contributors in your repository
+newPRWelcomeComment: >
+  Thanks for opening this pull request! Please check out our contributing guidelines.
```
jeff-cohere, line 11: We could hyperlink the guidelines here (if it's supported)

bartgol: Somehow, they get automatically hyperlinked by github anyways. I think it's coded in the app that creates the message (if the contributing guidelines is present in the repo, that is).

bartgol: Well, the guidelines are linked when opening an issue. Since I already opened a PR, I can't check that.

.github/config.yml
bartgol: Well, this message should appear after they picked the template. And the template is already populated in the issue.

PR #1, opened 2020-08-07
Remove scream refs

bartgol:
There should be no occurrence of the word "scream" in any file name/content.

PR #5, opened 2020-08-08
Massive refactor of EKAT

bartgol:
This PR is mainly targeting the removal of single/double precision as well as pack size information from Ekat configuration.

The idea is that one should build ekat just once, and then use it in multiple ways across a project as large as E3SM. For instance, one component (or even subset of a component, like an atm parametrization in SCREAM) might want to use single precision, while another double. Furthermore, pack size specs might vary depending on algorithmic differences. Most of Ekat is already templated anyways, and the precision/packsize info are mostly needed only in tests.

Speaking of tests, we obviously keep precision/packsize configurability for tests. In particular, options like `EKAT_DOUBLE_PRECISION` or `EKAT_PACK_SIZE` have been turned into `EKAT_TEST_DOUBLE_PRECISION` and `EKAT_TEST_PACK_SIZE`. These are then used to configure the testsuite. When it comes to tests, Ekat can test both single and double precision in the same build, but it is limited to one configuration option for the pack size. This should not be limiting, since in production run we won't bother building ekat's testsuite.

Furthermore, I am removing some stuff that does not belong in Ekat, such as the `RepoState` enum, or the TimeStamp class, in response to #1 (I am also pondering about stuff like `Units`, suggestions are welcome). Additionally, some stuff may be moved around in folders or across files, with the goal (hope) to increase ease of navigation in the code, and the ability to find what is needed.


bartgol:
This PR is complete, and ready for review. I am still having issues setting up the jenkins jobs for testing on GPU, so we'll have to wait to hook up the autotester (and I'd like this PR to be handled by the AT, so we know it's hooked up correctly).

However, we can go ahead and start the review.

bartgol:
@jeff-cohere @jgfouca , this is a _massive_ PR. Most of it, however, boils down to a replace of `scream` with `ekat`, or to an uppercasification of ekat's macros. If it's too much, let me know. We can split the review process by area (cmake, test directory, src directory...).

jeff-cohere:
Hey @bartgol ,

This is a big one! I'll take a look at it tomorrow. Thanks for your detailed description of the PR. That makes it much easier to know what to look for.

jeff-cohere:
One preliminary comment: can we add support for the NVidia Turing75 architecture? It's supported by Kokkos. I just got a Quadro RTX 4000 card for local building and have verified that it works for SCREAM.

bartgol:
Re: Turing. Do we have a testbed where we can run tests? Also, AFAIK, no big HPC cluster (at least among E3SM targets) will use Turing, they will jump straight to Ampere. Do we "need" this?

It's not really a big effort, just one file to add, so we could do it, even if there's no HPC cluster, but I'm a bit hesitant adding a config file that we can't test.

bartgol:
Notice, however, that the version of kokkos that is in e3sm (and scream) is quite old, and won't have Turing arch settings. We should also update the kokkos submodule in e3sm, by the way. But that should be done in e3sm. I am not 100% sure how big of an impact it will have on ekat. Besides mach files, we might have to change some kokkos-related cmake vars throughout ekat.

So, before we add Turing support, I'd like to see the kokkos submodule updated in e3sm. It's true that ekat does not necessarily use that version (one simply needs to provide a src dir for kokkos); nevertheless, that's the version of kokkos that I use to test ekat, which is reasonable, I think.

bartgol:
The AT framework seems to work almost perfectly. I say "almost", cause if you click on "Pull Request Auto Testing has PASSED (click to expand) ", the build parameters seem wrong. They seem to be the parameters used by SCREAM's testing jobs. I'm working with AT developers to see if this is a bug (related to multiple repo testing via the same AT job) or if I'm not configuring the AT correctly on my end.

cmake/Ekat.cmake
```diff
@@ -1,64 +1,46 @@
-# This cmake utility allows you to build ekat just by adding
+# This cmake utility allows you to build ekat just by doing
```
jeff-cohere, line 1: "adding" still seems better here.

cmake/Ekat.cmake
```diff
@@ -170,45 +138,62 @@ macro (setVars PREFIX SET_DEFAULTS)
...
+    if (EKAT_CUDA_BUILD)
+      set (EKAT_TEST_PACK_SIZE 1 CACHE STRING "")
+    else ()
+      set (EKAT_TEST_PACK_SIZE 16 CACHE STRING "")
+    endif()
   endif()
 
-  if (DEFINED ${PREFIX}_POSSIBLY_NO_PACK_SIZE)
-    set (EKAT_POSSIBLY_NO_PACK_SIZE ${${PREFIX}_POSSIBLY_NO_PACK_SIZE} CACHE STRING "")
+  if (DEFINED ${PREFIX}_TEST_SMALL_PACK_SIZE)
```
jeff-cohere, line 177: Just curious: what's a "small pack size"? Is this for GPU stuff that doesn't use vectorization, as suggested by POSSIBLY_NO_PACK_SIZE?

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -119,22 +103,27 @@ struct ExeSpaceUtils<Kokkos::Cuda> {
 #endif
 
 /*
- * TeamUtils contains utilities for getting concurrency info for
- * thread teams. Don't use _TeamUtilsCommonBase directly, use
- * TeamUtils.
+ * TeamUtils contains utilities for getting concurrency info for thread teams.
+ * You cannot use it directly (protected c-tor. You must use TeamUtils.
```
jeff-cohere, line 107: close parens around "protected c-tor"

src/ekat/util/ekat_factory.hpp
```diff
@@ -1,13 +1,13 @@
 #ifndef EKAT_FACTORY_HPP
 #define EKAT_FACTORY_HPP
 
+#include "ekat/ekat_assert.hpp"
```
jeff-cohere, line 4: Another question for my own curiosity: did you have to move this above the other includes to satisfy the compiler? It's more conventional for "home-made" headers to go below standard ones.

src/ekat/util/ekat_math_utils.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+template <typename Integer, typename Integer1> KOKKOS_INLINE_FUNCTION
+void set_min_max (const Integer& lim0, const Integer& lim1,
+                  Integer& min, Integer& max, const Integer1& vector_size) {
+  min = util::min(lim0, lim1) / vector_size;
+  max = util::max(lim0, lim1) / vector_size;
+}
+
+template <typename Real> KOKKOS_INLINE_FUNCTION
+Real reldif (const Real& a, const Real& b) {
```
jeff-cohere, line 76: Can I suggest `ref_diff` or `reldiff` here?

src/ekat/util/ekat_math_utils.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+template <typename T>
+KOKKOS_INLINE_FUNCTION
+const T& max (const T& a, const T& b) {
+  return a > b ? a : b;
+}
+
+template <typename T>
+KOKKOS_INLINE_FUNCTION
+bool isfinite (const T& a) {
```
jeff-cohere, line 32: is_finite, maybe?

src/ekat/util/ekat_test_utils.cpp
```diff
@@ -1,13 +1,18 @@
 #include <ekat/util/ekat_test_utils.hpp>
-#include <ekat/ekat_assert.hpp>
+#include "ekat/ekat_assert.hpp"
```
jeff-cohere, line 2: Any reason for not changing the line 1 to use quotes, too? Also: are you sure you want to use angle braces for Kokkos_Core.hpp?

src/ekat/util/ekat_type_traits.hpp
```diff
@@ -24,8 +26,8 @@ namespace ekat {
  * especially useful for bfb tests agaisnt fortran,
  * to ensure that literals are not a source of round-off differences.
  */
-template<typename T> KOKKOS_INLINE_FUNCTION
-constexpr typename std::enable_if<std::is_arithmetic<T>::value,Real>::type
+template<typename T, bool doublePrecision> KOKKOS_INLINE_FUNCTION
```
jeff-cohere, line 29: We're not considering half-precision floating point numbers, are we? If we are, using a bool to determine whether something is double precision or not might not give enough info.

tests/io/array_io_tests.f90
```diff
@@ -1,27 +1,31 @@
...
-
     integer(kind=c_int) :: nerr
     integer ::  i, j
     real(kind=c_real), target :: a(10,3), b(10,3)
     logical :: ok
 
     character(kind=c_char, len=128), parameter :: &
-         filename = c_char_"unit_test_f90_array_io.dat"//C_NULL_CHAR
+#ifdef EKAT_DOUBLE_PRECISION
+         filename = c_char_"unit_test_f90_array_io_dp.dat"//C_NULL_CHAR
```
jeff-cohere, line 25: Do you intend this as a C string in Fortran, or a Fortran string using C-compatible characters? My understanding is that it's the latter. Last I checked, C strings map to Fortran arrays of character(kind=c_char,len=1). But if you've had success using this, that would be interesting.

src/ekat/ekat_assert.hpp
```diff
@@ -29,32 +29,32 @@
...
     if ( ! (condition) )                        \
       Kokkos::abort(#condition "\n" msg);       \
   } while (0)
 
 #ifndef NDEBUG
-#define ekat_assert(condition)                impl_throw(condition, "",  std::logic_error)
-#define ekat_assert_msg(condition, msg)       impl_throw(condition, msg, std::logic_error)
-#define ekat_kassert(condition)               impl_kthrow(condition, "")
-#define ekat_kassert_msg(condition, msg)      impl_kthrow(condition, msg)
+#define EKAT_ASSERT(condition)                      IMPL_THROW(condition, "",  std::logic_error)
```
jgfouca, line 39: What's the benefit of making all these names all caps?

allevin:
@bartgol @jgfouca I strongly recommend that you hold off on merging this PR till Luca can confirm that the test was run correctly on the Autotester.  The AT parameters issue mentioned above should be fairly easy to fix, but may have sent un-used data to the job causing it to use the default parameters.  I have an email into Luca on how to resolve.

cmake/Ekat.cmake
bartgol: So, this is linked to the fact that, IIRC, in the pack seminal work, Jim and Andrew found it is actually ok (or maybe even a tad better) to use a pack size that is larger than the arch's vector unit (of course, it must be a multiple of that). However, some loops might benefit from having smaller packs (like a single vector unit). Therefore, scream introduced a "small" pack size (like a single VU), as opposed to the "normal" pack size (a multiple of a single VU).

bartgol: POSSIBLY_NO_PACK_SIZE is honestly a bit confusing. It was introduced for the LinInterp mini app, where, on SKX, Jim observed that a pack size of 1 performs better than a pack size of 8 (or more). This is somewhat counterintuitive, especially considering pack size 8 is better than 1 on other arch (like KNL).

So that pack size is for algorithms such as lin interp which on SKX seem to prefer no pack, but on other archs still want a pack size larger than 1.

bartgol: Imho, scream should NOT use a pack size at the SCREAM level, but rather have each individual parametrization use their own pack size. That's one of the reasons I purged the pack size from the configuration of the library (the config options here above are only used in the testing of ekat).

src/ekat/kokkos/ekat_kokkos_utils.hpp
bartgol: Wow, you really are reading every line! Thanks! =)

src/ekat/util/ekat_factory.hpp
bartgol: I usually follow the approach of going from local to global (which I thought was more common), to avoid the scenario where the code compiles just because you happen to include things in the right order. If, say, `<string>` is needed in `ekat_type_traits.hpp`, I want it to be included there. If I put std headers first, that lack of include directive in `ekat_tyep_traits.hpp` might go undetected.

src/ekat/util/ekat_math_utils.hpp
bartgol: Yes, I like that better.

src/ekat/util/ekat_test_utils.cpp
bartgol: You're right, ekat's headers should use quotes rather than braces. For kokkos, I am leaning on braces, since, even if we build it as part of the project, it semantically is a TPL. We don't develop it, and we don't test it. We only build it on the fly cause it's cheap, and avoids the headache of having it pre-installed (and maintained) on all machines.

src/ekat/util/ekat_math_utils.hpp
bartgol: Ah, I wanted to change this too. The problem is that the STL has `isfinite`. A few lines below, you can see that if the exec space is not cuda we simply do `using std::isfinite`. In order to have `is_finite`, we'd have to impl a wrapper fcn around `std::isfinite`. It's doable, maybe a bit tedious. Do you think the benefit of snake case outweigh the benefit of using the same fcn name as the STL? I don't have a strong opinion either way.

src/ekat/util/ekat_type_traits.hpp
bartgol: I thought about this. I even started writing an enum for the precision. However, I refrained myself, since I wasn't sure supporting half precision was ever going to be useful in the E3SM ecosystem.

That said, supporting it with an enum would not be that hard. But if we show it in the enum, we should make sure all our utils and algorithms either support it, or throw errors (possibly at compile time) if one tries to use them with half precision.

tests/io/array_io_tests.f90
bartgol: It's the latter. A Fortran array of T is interpreted as a ptr to T in C, so this works. However, notice that in the interoperable subroutines that use such strings, you need to declare them with an unspecified length (see ekat_array_io_mod.f90).

bartgol: Argh, that file is still called `array_io_mod.f90`, while I thought I prepended `ekat_` to all files...

src/ekat/kokkos/ekat_kokkos_utils.hpp
jeff-cohere: Skimming. :-) Just happened to scan this one.

src/ekat/util/ekat_factory.hpp
jeff-cohere: That makes sense.

src/ekat/ekat_assert.hpp
bartgol: When you see `ekat_assert_msg(cond, "blah");` you are not sure if it is a fcn or a macro. All caps are almost a universal sign of either a macro or a compile-time constant. I think it's beneficial to let the user know that these are macros, not function.

This is an unlikely scenario for these macros, but still: taking the address of a fcn is ok, but if you try the same with a macro you might get a cryptic compilation errror.

src/ekat/util/ekat_math_utils.hpp
jeff-cohere: I forgot about the standard function. Yeah, no need to be over-consistent with snake case (C++ certainly isn't!). I think it's fine as is.

src/ekat/util/ekat_type_traits.hpp
jeff-cohere: Half precision has been discussed on the EAGLES project, but it's still just an idea. I didn't know if other E3SM subprojects were toying with it too. What do you think? Should we wait to hear that someone needs it and change in a later PR if/when we decide to, maybe?

PeterCaldwell: I'm definitely interested in half precision but think we should just focus on "get something working" now. Thus if it is easy to generalize things now to make half precision easier later, we should do it. Otherwise, we should save these things for a future PR.

bartgol: Let's open an issue in Ekat, to keep track of the idea (and the needs for that idea). But for now, I'm also for getting stuff going.

But please, feel free to formally request it on behalf of EAGLES (or whatever project), if you move forward with the idea.

jeff-cohere: Opened a feature request.

src/ekat/util/ekat_array_io_mod.f90
```diff
@@ -0,0 +1,39 @@
+module array_io_mod
```
jeff-cohere, line 1: Just checking: we don't want to name this module `ekat_array_io_mod`, do we?

src/ekat/util/ekat_test_utils.hpp
```diff
@@ -72,6 +60,15 @@ void catch2_req_pk_sensitive(const Scalar lhs, const Scalar rhs)
   }
 }
 
+// Check whether a string (from argv) mathces a predefined option name,
```
jeff-cohere, line 63: mathces -> matches (not a big deal... in case you're wondering, I was a writer for a couple of years, so my eyes catch this stuff almost automatically.)

src/ekat/util/ekat_array_io_mod.f90
bartgol: Mmm, good point. I think it's probably best to do it. The `ekat_` prefix in the module name will serve as an analogue of the C++ namespace, in the quite rare case that there already is an array_io_mod defined elsewhere. What do you think?

jeff-cohere: I agree it's probably a good idea. Unlikely that omitting the ekat_ prefix would cause a collision, but it's a very easy thing to add.

jeff-cohere (dismissed):
Looks great to me! I only found a couple very minor things that I've mentioned. Good to go once we address those.

PR #15, opened 2020-08-26
Added another template parameter to host_device/device_to_host functions.

jeff-cohere:
This allows us to use signed integer types to express View dimensions. The size data (elements of type `SizeT`), are cast to `size_t` within these functions in order to play nicely with Kokkos. 

## Motivation
This eliminates compiler warnings encountered in [SCREAM issue #524](https://github.com/E3SM-Project/scream/issues/524).

## Testing

The affected function templates, `host_to_device` and `device_to_host`, are tested in various tests within `tests/pack/pack_kokkos_tests.cpp`.

### DIFF
--- a/src/ekat/ekat_pack_kokkos.hpp
+++ b/src/ekat/ekat_pack_kokkos.hpp
@@ -249,15 +249,15 @@ repack (const Kokkos::View<Pack<T, old_pack_size>*, Parms...>& vp) {
 //
 
 // 1d
-template <size_t N, typename ViewT>
+template <typename SizeT, size_t N, typename ViewT>
 void host_to_device(const Kokkos::Array<typename ViewT::value_type::scalar const*, N>& data,
-                    const Kokkos::Array<size_t, N>& sizes,
+                    const Kokkos::Array<SizeT, N>& sizes,
                     Kokkos::Array<ViewT, N>& views)
 {
   using PackT = typename ViewT::value_type;
 
   for (size_t i = 0; i < N; ++i) {
-    const size_t size = sizes[i];
+    const size_t size = static_cast<size_t>(sizes[i]);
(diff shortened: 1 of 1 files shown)


src/ekat/ekat_pack_kokkos.hpp
```diff
@@ -249,15 +249,15 @@ repack (const Kokkos::View<Pack<T, old_pack_size>*, Parms...>& vp) {
...
 void host_to_device(const Kokkos::Array<typename ViewT::value_type::scalar const*, N>& data,
-                    const Kokkos::Array<size_t, N>& sizes,
+                    const Kokkos::Array<SizeT, N>& sizes,
                     Kokkos::Array<ViewT, N>& views)
 {
   using PackT = typename ViewT::value_type;
 
   for (size_t i = 0; i < N; ++i) {
-    const size_t size = sizes[i];
+    const size_t size = static_cast<size_t>(sizes[i]);
```
bartgol, line 260: Is there really a need to cast to size_t? Why not simply using `sizes[i]`?

jeff-cohere: I don't think there's a need, but implicit casts can be a source of error, so I'm not in the habit of using them outside of C (and even then I usually cast explicitly). Maybe it's superstitious, but C++ has enough bizarre rules that I'm in favor of making things like this more explicit anyway.

bartgol: Ah, true, there would otherwise be an implicit cast when you compute npacks. Then yeah, keep static casts.

jeff-cohere:
Anything left to do on this one? Should I set it to auto-merge?

jgfouca:
I retroactively approve.

PR #17, opened 2020-08-27
Fix overload of operator<< for std::vector

bartgol:
The overload must be placed in the std namespace, in order for the ADL rule to kick in, and find the overload.

## Motivation

The operator was already overloaded, but the overload was placed in the `ekat::util` namespace. In order for ADL (Augmented Dependent Lookup) rule to work with something like a simple `std::cout << my_vector`, the overload for op<< must be in the std namespace.

## Additional Information
Note: I still think the way it was before should have worked, since the overload was provided in the same namespace where it was used. However, putting it in the std namespace seems like a safe solution.

### DIFF
--- a/src/ekat/std_meta/ekat_std_utils.hpp
+++ b/src/ekat/std_meta/ekat_std_utils.hpp
@@ -38,21 +38,6 @@ int count (const ContainerType& c, const T& value) {
   return std::count(c.begin(), c.end(), value);
 }
 
-template<typename T>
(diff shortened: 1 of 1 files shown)


bartgol:
> This looks fine. The only potential issue I can see is that if someone wants to overload this operator for an std::vector their own way, they might be astonished to find that the `std` namespace already contains this version (if they're in a translation unit that includes `ekat_std_utils.hpp`). But I don't think that's too likely, and this does get around C++'s ADL limitations.

True. But hopefully, one implementation of op<< will be enough for all needs.

PR #22, opened 2020-08-28
Added utility for string similarity check.

bartgol:
This PR implements a string similarity algorithm, using the Jaro (and Jaro-Winkler) algorithm.

## Motivation

We are already talking about forcing standard names for fields in SCREAM. If we move forward with it (and it seems very likely), users might get errors if their field names are not conforming to a standard. But it would be nice to "suggest" similar names, which is also handy if the user simply makes a typo (e.g., 'atm_presure' instead of 'atm_pressure').

A string similarity check can help us find the "correct" name in a database.

## E3SM Stakeholder Feedback

See E3SM-Project/scream#528

## Testing

I implemented tests of the algorithm against exact values, taken from [Winkler's paper](https://www.census.gov/srd/papers/pdf/rrs2006-02.pdf), page 12. I removed a couple of entries cause Winkler truncated to 0 all similarities under 0.7.

### DIFF
--- a/src/ekat/util/ekat_string_utils.cpp
+++ b/src/ekat/util/ekat_string_utils.cpp
@@ -1,4 +1,5 @@
 #include "ekat/util/ekat_string_utils.hpp"
+#include "ekat/ekat_assert.hpp"
 
 #include <algorithm>
 #include <sstream>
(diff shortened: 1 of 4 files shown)


src/ekat/util/ekat_string_utils.cpp
```diff
@@ -58,6 +59,73 @@ std::string upper_case (const std::string& s) {
   return s_up;
 }
 
+double jaro_similarity (const std::string& s1, const std::string& s2) {
```
jeff-cohere, line 62: Can you describe briefly some typical output from this algorithm and/or give a reference to it above the function declaration? That will help folks determine where it's needed. And if you have a typical use case in mind, might want to include that here too!

src/ekat/util/ekat_string_utils.cpp
```diff
@@ -58,6 +59,73 @@ std::string upper_case (const std::string& s) {
...
+        transp += 0.5;
+      }
+      ++k;
+    }
+  }
+
+  return ( matches/len1 + matches/len2 + (matches-transp)/matches ) / 3.0;
+}
+
+double jaro_winkler_similarity (const std::string& s1, const std::string& s2,
```
jeff-cohere, line 108: Same request here--it'd just be nice to get a brief description of this algorithm, typical output values for some simple input, and maybe contrast this algorithm with the Jaro algorithm above.

src/ekat/util/ekat_string_utils.hpp
```diff
@@ -37,6 +37,21 @@ std::string strint (const std::string& s, const int i);
 // Conver the string to all upper case
 std::string upper_case (const std::string& s);
 
+// Computing similarity index between s1 and s2 using Jaro algorithm
```
jeff-cohere, line 40: Can you describe briefly some typical output from this algorithm and/or give a reference to it above the function declaration? That will help folks determine where it's needed. And if you have a typical use case in mind, might want to include that here too!

src/ekat/util/ekat_string_utils.hpp
```diff
@@ -37,6 +37,21 @@ std::string strint (const std::string& s, const int i);
...
+double jaro_similarity (const std::string& s1, const std::string& s2);
+
+// Computing similarity index between s1 and s2 using Jaro-Winkler algorithm
+// For meaning and bounds for the optional arguments l, p, see e.g.
+//      https://en.wikipedia.org/wiki/Jaro-Winkler_distance
+// This routine computes jaro similarity (sj), and if sj>thresholds,
+// it performs the winkler adjustment, otherwise returns sj.
```
jeff-cohere, line 47: This is a great description. Thanks!

bartgol:
Re: jenkins. We tried to switch to a "private" jenkins instance, but the AT is only able to use the "public" instance at SNL. It's a pity, cause the private one has some benefit. I wrote to one of the AT devs to see if we can hack the AT to use a personal jenkins server.

PR #24, opened 2020-08-29
Add none() method to Mask

### LINKED ISSUE #12: Add 'none()' method to Mask structure
**Feature description**
Add a `none()` method to the `Mask` structure, that returns `true` if all the entries in the mask are `false`, and return `false` otherwise.

Same implementation of `any()`, with `true` and `false` swapped.

**Describe alternatives you've considered**
Strictly speaking, it does not increase the set of capabilities, since `my_mask.none()` would be equivalent to `!my_mask.any()`. However, I would argue the code would be faster to parse.

bartgol:
Add `none()` method to Mask structure.

## Motivation

It does not add new capabilities, given that `m.none()` is the same as `!m.any()`. However, the former is easier/faster to read/understand than the second.

## Related Issues

* Closes #12  

## E3SM Stakeholder Feedback

Requested by developers of SHOC inside SCREAM

## Testing

Testing is performed to make sure `none()` is the opposite of `any()`.

### DIFF
--- a/src/ekat/ekat_pack.hpp
+++ b/src/ekat/ekat_pack.hpp
@@ -66,6 +66,11 @@ struct Mask {
     return b;
   }
 
+  // Are all slots false?
+  KOKKOS_FORCEINLINE_FUNCTION bool none () const {
+    return !any();
+  }
+
 private:
   type d[n];
 };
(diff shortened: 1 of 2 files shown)


bartgol:
> We will need to grep through the ekat and scream code and covert all the !any to none.

That'd be a nice modification, although not strictly necessary (read: not urgent, do it only if you're waiting on other code to compile). However, we should "encourage" to use `m.none()` instead of `!m.any()` in all future PRs, for readability.

PR #27, opened 2020-09-01
Added kokkos and yaml-cpp as submodules (plus minor fixes)

### LINKED ISSUE #19: New Kokkos versions don't work with EKAT
Kokkos release 3.2 (and the Kokkos master branch) fail to build within EKAT.

Newer Kokkos versions enforce case-sensitive CMake variables, split libkokkos.a into 2 (libkokkoscore.a and libkokkoscontainers.a), and change some memory traits (RandomAccess) that ekat currently uses.

Alternative: Force EKAT to use the same Kokkos SHA as scream.

jeff-cohere:
Here we've added a verified version of the Kokkos library and a forked version of yaml-cpp as submodules to EKAT.

Notes:

* By default, the submodule versions of Kokkos and yaml-cpp are selected. Either or both may be overridden using the appropriate CMake options.
* References to deprecated fields and members in Kokkos have been replaced with their new equivalents.

## Motivation

With this change, we're moving toward a model in which EKAT supplies these libraries to E3SM performance-related projects. We're also delivering a much more recent version of Kokkos to EKAT-based projects.

Closes #19 

## Testing

All tests pass. No new features were introduced.

## Bugfixes

* A call to `fflush` was added to the `IMPL_KERNEL_THROW` macro to ensure that the diagnostic message is printed before `abort` is called.


jeff-cohere:
Here's a funny thing: on my Linux machine(s), I encounter compile errors in the same line as on the AT when I use `DefaultDevice::execution_space::impl_is_initialized()` instead of `DefaultDevice::execution_space::is_initialized()`. This seems to me to be consistent with the fact that we're #defining `Kokkos_ENABLE_DEPRECATED_CODE` (see, for example, `extern/kokkos/core/src/Kokkos_Serial.hpp`, lines 145 and 165). So I'm not sure why the AT systems aren't building properly.

jeff-cohere:
@bartgol , how is this being built on the AT? If we're feeding EKAT directories for Kokkos and yaml-cpp, maybe we need to delete those and let EKAT use its own versions?

bartgol:
> @bartgol , how is this being built on the AT? If we're feeding EKAT directories for Kokkos and yaml-cpp, maybe we need to delete those and let EKAT use its own versions?

Ah, good point. I will edit the AT jobs to _not_ use a pre-existing kokkos/yaml src dir, and see what happens. As long as _no other EKAT prs_ are submitted before this goes in, we're good.

cmake/Kokkos.cmake
```diff
@@ -11,18 +11,26 @@ get_property(IS_EKAT_KOKKOS_BUILT GLOBAL PROPERTY EKAT_KOKKOS_BUILT SET)
 if (NOT IS_EKAT_KOKKOS_BUILT)
 
   if (NOT Kokkos_SOURCE_DIR)
-    message (FATAL_ERROR "Error! Please, specify path to Kokkos in Kokkos_SOURCE_DIR.\n")
+    message (STATUS "Kokkos_SOURCE_DIR not specified: using submodule version.")
+    if(DEFINED Kokkos_ENABLE_DEPRECATED_CODE AND Kokkos_ENABLE_DEPRECATED_CODE)
```
bartgol, line 15: I don't think you need the first part inside the if statement. `if (MY_VAR)` is automatically false if `MY_VAR` is not defined.

jeff-cohere: Ah, right. I had the opposite logic before, so that's left over.

bartgol (dismissed):

jeff-cohere:
Thanks for the heroic effort, Luca!

jeff-cohere:
It builds and runs successfully on my Mac, too (minus the FPE stuff), so I'm merging this.

PR #29, opened 2020-09-02
Flattening all namespaces within ekat except impl/detail

### LINKED ISSUE #10: Should we remove all namespaces in ekat, except ekat itself?
The ekat namespace should already be enough to encapsulate ekat's stuff in any downstream application. Is there really a point in having, say, `ekat::util`? After all, isn't everything in ekat a utility function/class anyways? Also  the `ekat::pack` namespace seems to be an overkill.

The reason why I would purge nested namespaces like `util` is that it would somewhat help keeping downstream codes shorter. Sure, one can always issue a `using ekat::util::blah`/`using blah = ekat::util::blah`, or even `using namespace ekat::util', but it all seems overly complicated.

Note: any sort of `impl` or `details` namespace should be kept, since they have a completely different purpose (isolating stuff that the user should not try to use/access).

jeff-cohere:
These commits get rid of the various namespaces within `ekat`, in an attempt to simplify the usage of the library (and shorten qualified calls to functions and references to types). Specifically, these removed namespaces include `util`, `pack`, and `tridiag`.

## Notes

`util::min`, `util::max`, and related functions now live in `impl::min` and `impl::max`, and are only provided as drop-in replacements for missing functions on GPUs.

## Related Issues

* Closes #10 

## Testing

This isn't working yet--I've created this PR to make it easier to see what's changed.

### DIFF
--- a/src/ekat/ekat_pack.hpp
+++ b/src/ekat/ekat_pack.hpp
@@ -12,7 +12,6 @@
 #include <Kokkos_Core.hpp>
 
 namespace ekat {
-namespace pack {
 
 /* API for using "packed" data in ekat. Packs are just bundles of N
    scalars within a single object. Using packed data makes it much easier
@@ -304,14 +303,14 @@ ekat_pack_gen_unary_stdfn(tanh)
 template <typename PackType> KOKKOS_INLINE_FUNCTION
 OnlyPackReturn<PackType, typename PackType::scalar> min (const PackType& p) {
   typename PackType::scalar v(p[0]);
-  vector_disabled for (int i = 0; i < PackType::n; ++i) v = util::min(v, p[i]);
+  vector_disabled for (int i = 0; i < PackType::n; ++i) v = min(v, p[i]);
   return v;
 }
 
 template <typename PackType> KOKKOS_INLINE_FUNCTION
 OnlyPackReturn<PackType, typename PackType::scalar> max (const PackType& p) {
   typename PackType::scalar v(p[0]);
-  vector_simd for (int i = 0; i < PackType::n; ++i) v = util::max(v, p[i]);
+  vector_simd for (int i = 0; i < PackType::n; ++i) v = max(v, p[i]);
   return v;
 }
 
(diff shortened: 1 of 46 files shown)


bartgol:
I think the `units` namespace has to be kept. Perhaps it can only be put around the declaration of the standard units (m, s, kg, ...) at the end of `ekat_units.hpp`, kind of like it is done at the end of `ekat_scaling_factor.hpp` (with the `prefixes` namespace).

If those file-level variables (which have (at least some) very common names) pollute the ekat namespace, we might get some errors down the road.

jeff-cohere:
Thanks.

Another issue I'm seeing is that the `kernel_assert_ut` test, which verifies that a false predicate causes a process to abort, is failing on my local machine. It seems to me that the behavior of this particular test depends on the system, and that it may be a tad pedantic to do checking for proper aborts. CMake doesn't offer an easy solution for this, and our `sh`/`grep` hack seems not to work in my environment. Nor does the `WILL_FAIL` test property make it okay for the process to abort.

I get the need to make sure a process discontinues if an assertion fails, but the amount of trouble this is causing (considering the simplicity of the mechanism) is kind of overshadowing this whole task. It's possible that something has been triggered by namespace flattening(?), but I'm kind of at a loss for how to fix this test failure.

bartgol:
I added that test out of a overly-zealous sentiment. I'm 100% fine not testing it. All it is doing is making sure that our abort routines print a non-literal string. It's quite minimal, and if that's causing any issue whatsoever, I would simply get rid of it. I'm puzzled on why it doesn't work for you though, considering it worked on my machine (gcc), and all our testing machines. But not that puzzled to slow you down.

jeff-cohere:
Okay, cool. I've commented out the test for now, and I've reintroduced the units/prefixes namespaces.

bartgol (approved):

bartgol:
The AT failed twice, so this is likely a legitimate error. It's hard to tell what the problem is (perhaps a double free?). You might have to fire up cuda-gdb.

bartgol (approved):

PR #30, opened 2020-09-02
Added install targets for ekat and yaml-cpp.

### LINKED ISSUE #28: Add an installation (make install) target
It would be nice if someone who uses ekat can place its libraries/headers into the proper location, given a `CMAKE_INSTALL_PREFIX`. This is pretty standard practice for libraries these days.

I'd like to add a `make install` target that uses CMake's `install` command to put the goods where a user wants them.

jeff-cohere:
This is a quick and hopefully not-too-dirty set of installation targets for ekat and yaml-cpp. Kokkos already seems to install itself. No changes to any ekat functionality in this PR.

## Related Issues

* Closes #28


jeff-cohere:
Drat. I guess I'm used to doing things more the library way and not the submodule way. I'll poke at this.

bartgol:
You don't need to manually install yaml, you can rely on its own install logic (like you do for kokkos). However, they made the `YAML_CPP_INSTALL` option conditional to yaml src folder being the main cmake source folder. That's annoying, cause there's no way for us to override that. But...since we are using yaml from our SNL fork, we can hack that part of their CMakeLists, and allow installing yaml even if CMAKE_SOURCE_DIR != YAML_SOURCE_DIR... I manually tried, and that works (our install tree is not very tidy, but that's another story).

jeff-cohere:
That sounds better than this manual installation. How would that work, then? Would we submit a PR to the SNL fork of yaml-cpp?

bartgol:
> That sounds better than this manual installation. How would that work, then? Would we submit a PR to the SNL fork of yaml-cpp?

Yeah, that'd be my approach. We already did change something in our fork (I think we removed the hard-coded warnings-as-errors flag). I can take care of that.

bartgol:
Ah, I just noticed you are using the master branch inside ekat/extern. In scream we're using the `e3sm-branch` branch (I just checked, the removed flag were `-pedantic -pedantic-errors`, which caused build failures with some compilers).

I'm going to push the mod to the e3sm-branch, then you can update the yaml submodule version once.

bartgol (approved):

Issue #32, opened 2020-09-04
Add parallel scan/reduce dispatch routines, that run serially under certain configurations.


bartgol:
**Describe the solution you'd like**
We implemented something like this in [homme](https://github.com/E3SM-Project/E3SM/blob/7d36ff4b568044b80d1dc5aba4194c9563e88396/components/homme/src/share/cxx/ExecSpaceDefs.hpp#L260). The idea is to force the for loop into a Kokkos::single in case some macro is defined (or we could use template specialization, so that EKAT doesn't need yet another macro). Somthing like this:

```
template<typename DeviceType, bool Serialized = false>
struct Dispatch {
  template<typename TeamMember, typename Lambda, typename ValueType>
  void team_reduce (const TeamMember& team,
                                 const int& begin, const int& end,
                                 const Lambda& lambda,
                                           ValueType& result) {
     Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, begin, end), lambda, result);
  }
  // Similar code for thread vector range and scan
}; 

template<typename DeviceType>
struct Dispatch<DeviceType,true> {
  template<typename TeamMember, typename Lambda, typename ValueType>
  void team_reduce (const TeamMember& team,
                                 const int& begin, const int& end,
                                 const Lambda& lambda,
                                           ValueType& result) {
     Kokkos::single(Kokkos::PerTeam(team),[&] {
        result = ValueType();
        for (int k=start; k<end; ++k) {
           lambda(k, result);
        }
#ifdef KOKKOS_ENABLE_CUDA
       // Broadcast result to all threads by doing sum of one thread's
       // non-0 value and the rest of the 0s.
       Kokkos::Impl::CudaTeamMember::vector_reduce(Kokkos::Sum<ValueType>(result));
#endif
     });
     // Similarly code for thread vector range and scan
  }
};
..
```

**Describe alternatives you've considered**
Now, one needs to manually put serial/parallel logic in the code.

**Additional context**
See Homme's implementation for a possible impl example.

jeff-cohere:
I talked to @tcclevenger and he's going to take this on en route to #26.

bartgol:
This was completed when #37 got merged.

Issue #26, opened 2020-08-31
Add a `sum` function to `ekat_pack.hpp`.


jeff-cohere:
**Is your feature request related to a problem? Please describe.**
See [Andrew Bradley's comment in SCREAM PR 536](https://github.com/E3SM-Project/scream/pull/536/files#r480314177). It would be helpful to have a `sum` function that returns the sum of all elements in a pack.

**Describe the solution you'd like**
The proposed solution is to define the following inline function in `ekat_pack.hpp`:

```
template <typename PackType> KOKKOS_INLINE_FUNCTION
OnlyPackReturn<PackType, typename PackType::scalar> reduce_sum (const PackType& p) {
  typename PackType::scalar v(p[0]);
  vector_simd for (int i = 1; i < PackType::n; ++i) v += p[i];
  return v;
}
```

**Additional context**

bartgol:
For similarity with Intel's AVX512 instruction, I would call it `reduce_add`, or maybe `reduce_sum`; having `reduce` in the name makes it clear it is summing _within_ the vector (and it's also easier to grep). But that's just a personal preference.

bartgol:
Also, is there really any benefit in decorating the loop with simd? I know someone might say "worst case scenario, nothing happened, so you get the same behavior as without the `vector_simd` macro", but I'm always wary of adding things "just in case", when they could be misleading...

jeff-cohere:
> For similarity with Intel's AVX512 instruction, I would call it `reduce_add`, or maybe `reduce_sum`; having `reduce` in the name makes it clear it is summing _within_ the vector (and it's also easier to grep). But that's just a personal preference.

I think that's fine. Wouldn't this also be the case for `min` and `max`, too? Can you point me to a link that describes the AVX-512 instruction you're referring to?

Regarding `vector_simd`: I had no specific reason for including it. I just copy-pasted the `min` (or `max`) implementation. I'm not versed in all of this machinery just yet.

bartgol:
[This](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#) is the intel intrinsics guide, which includes plenty of vector extension versions. Probably you're most interested in AVX2 and AVX512 (the latter comes in n+1 flavors).

I don't think using similar names to avx is stricly necessary, but it might help convey the message that packs are to enhance vectorization (in case it wasn't already clear).

[...]

PR #34, opened 2020-09-09
add parallel_reduce

tcclevenger:
## Motivation
`ExeSpaceUtils::parallel_reduce()` discussed in Issue https://github.com/E3SM-Project/EKAT/issues/32. Allows for a serial implementation for `parallel_reduce()` for the purpose of matching bfb to fortran.

`ExeSpaceUtils::reduce_add()` discussed in Issue https://github.com/E3SM-Project/EKAT/issues/26. Provides a function which gives the sum over an `ekat::Pack`. Also allows for a serial implementation for the purpose of matching bfb to fortran.

## Testing
I created 2 tests in **tests/kokkos/kokkos_utils_tests.cpp**:  `parallel_reduce` and `reduce_add`, which compare some simple serial computation done in for loops (like in fortran) with the serialized version of the implemented functions. Each test, when tested on the GPU, passes *only* when `serialize=true`. On the CPU, `parallel_reduce` passes with  `serialize=false`, but that may be expected.

## Additional Information
A few comments/questions:

- Ideally `bool serialize` would be a template parameter in both functions, however for `parallel_reduce()` I was unsure how I would call the function given that I do not know the explicit Lambda type and am relying on the compiler to detect it for me. Also, the src directory does not seem to know the correct value of `EKAT_TEST_STRICT_FP`, so I couldn't use that as in [this example (line 321)](https://github.com/E3SM-Project/E3SM/blob/7d36ff4b568044b80d1dc5aba4194c9563e88396/components/homme/src/share/cxx/ExecSpaceDefs.hpp#L260). Open to suggestions.
- For both functions, I basically have a completely identical copy for Cuda and CPU implementation. Is there a way I can avoid this? 
- In `reduce_add()`, I'm guessing that `PackType::n`return the potential size of the pack, and does not account for garbage if it is not completely full? Which could be an issue if I'm summing up all entries.. From [here](https://github.com/E3SM-Project/E3SM/blob/b7b269b3cfaca869e861fa01e6d65e201dc588a3/components/homme/src/share/cxx/ColumnOps.hpp#L474), they were able to access `ColInfo<LENGTH>::LastPackLen` to test if the final pack was full. Do I have something similar available here?
- I'm open to all suggestions, criticisms, etc. Let me know too if the formatting/indenting is acceptable.

### DIFF
--- a/src/ekat/kokkos/ekat_kokkos_utils.hpp
+++ b/src/ekat/kokkos/ekat_kokkos_utils.hpp
@@ -28,7 +28,7 @@ namespace ekat {
 
 namespace impl {
 
-  template <typename TeamMember, class Lambda, typename ValueType, bool Serialize>
+template <bool Serialize, typename TeamMember, typename Lambda, typename ValueType>
 static KOKKOS_INLINE_FUNCTION
 void parallel_reduce (const TeamMember& team,
                       const int& begin,
@@ -125,13 +125,7 @@ struct ExeSpaceUtils {
     return TeamPolicy(ni, team_size);
   }
 
-  template <typename TeamMember, class Lambda, typename ValueType, bool Serialize =
-#ifdef EKAT_TEST_STRICT_FP
-      true
-#else
-      true //This is a hack for now since EKAT_TEST_STICT_FP is not recognized in src/
-#endif
-      >
+  template <bool Serialize, typename TeamMember, typename Lambda, typename ValueType>
   static KOKKOS_INLINE_FUNCTION
   void parallel_reduce (const TeamMember& team,
 			const int& begin,
@@ -139,7 +133,7 @@ struct ExeSpaceUtils {
                         const Lambda& lambda,
                         ValueType& result)
   {
-    impl::parallel_reduce<TeamMember, Lambda, ValueType, Serialize>(team, begin, end, lambda, result);
+    impl::parallel_reduce<Serialize, TeamMember, Lambda, ValueType>(team, begin, end, lambda, result);
   }
 
   template<typename PackType, typename ValueType, bool Serialize =
@@ -186,13 +180,7 @@ struct ExeSpaceUtils<Kokkos::Cuda> {
 
 
 
-  template <typename TeamMember, class Lambda, typename ValueType, bool Serialize =
- #ifdef EKAT_TEST_STRICT_FP
-        true
- #else
-        true //This is a hack for now since EKAT_TEST_STICT_FP is not recognized in src/
- #endif
-        >
+  template <bool Serialize, typename TeamMember, typename Lambda, typename ValueType>
   static KOKKOS_INLINE_FUNCTION
   void parallel_reduce (const TeamMember& team,
 			const int& begin,
@@ -200,7 +188,7 @@ struct ExeSpaceUtils<Kokkos::Cuda> {
                         const Lambda& lambda,
                         ValueType& result)
   {
-    impl::parallel_reduce<TeamMember, Lambda, ValueType, Serialize>(team, begin, end, lambda, result);
+    impl::parallel_reduce<Serialize, TeamMember, Lambda, ValueType>(team, begin, end, lambda, result);
   }
 
   template<typename PackType, typename ValueType, bool Serialize =
--- a/src/ekat/kokkos/ekat_kokkos_utils.hpp
+++ b/src/ekat/kokkos/ekat_kokkos_utils.hpp
@@ -26,6 +26,52 @@
 
 namespace ekat {
 
+namespace impl {
+
+  template <typename TeamMember, class Lambda, typename ValueType, bool Serialize>
+static KOKKOS_INLINE_FUNCTION
+void parallel_reduce (const TeamMember& team,
+                      const int& begin,
+                      const int& end,
+                      const Lambda& lambda,
+                      ValueType& result)
+{
+  if (Serialize) {
+    // We want to get C++ on GPU to match F90 on CPU. Thus, need to
+    // serialize parallel reductions.
+
+    // All threads init result.
+    // NOTE: we *need* an automatic temporary, since we do not know
+    //       where 'result' comes from. If result itself is an automatic
+    //       variable, using it would be fine (only one vector lane would
+    //       actually have a nonzero value after the single). But if
+    //       result is taken from a view, then all vector lanes would
+    //       see the updated value before the vector_reduce call,
+    //       which will cause the final answer to be multiplied by the
+    //       size of the warp.
+    auto local_tmp = Kokkos::reduction_identity<ValueType>::sum();
+
+    Kokkos::single(Kokkos::PerThread(team),[&] {
+        for (int k=begin; k<end; ++k) {
+          lambda(k, local_tmp);
+        }
+      });
+
+#ifdef KOKKOS_ENABLE_CUDA
+    // Broadcast result to all threads by doing sum of one thread's
+    // non-0 value and the rest of the 0s.
+    Kokkos::Impl::CudaTeamMember::vector_reduce(Kokkos::Sum<ValueType>(local_tmp));
+#endif
+
+   result = local_tmp;
+  } else {
+    Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, begin, end), lambda, result);
+  }
+}
+
+} //namespace impl
+
+
 template<typename DataTypeOut, typename DataTypeIn, typename... Props>
 typename std::enable_if<GetRanks<DataTypeOut>::rank_dynamic==0,
                         typename ekat::Unmanaged<Kokkos::View<DataTypeOut,Props...>>>::type
@@ -83,7 +129,7 @@ struct ExeSpaceUtils {
 #ifdef EKAT_TEST_STRICT_FP
       true
 #else
-      false //This is a hack for now since EKAT_TEST_STICT_FP is not recognized in src/
+      true //This is a hack for now since EKAT_TEST_STICT_FP is not recognized in src/
 #endif
       >
   static KOKKOS_INLINE_FUNCTION
@@ -93,38 +139,7 @@ struct ExeSpaceUtils {
                         const Lambda& lambda,
                         ValueType& result)
   {
-    if (Serialize)
-    {
-      // We want to get C++ on GPU to match F90 on CPU. Thus, need to
-      // serialize parallel reductions.
-
-      // All threads init result.
-      // NOTE: we *need* an automatic temporary, since we do not know
-      //       where 'result' comes from. If result itself is an automatic
-      //       variable, using it would be fine (only one vector lane would
-      //       actually have a nonzero value after the single). But if
-      //       result is taken from a view, then all vector lanes would
-      //       see the updated value before the vector_reduce call,
-      //       which will cause the final answer to be multiplied by the
-      //       size of the warp.
-      auto local_tmp = Kokkos::reduction_identity<ValueType>::sum();
-
-      Kokkos::single(Kokkos::PerThread(team),[&] {
-      for (int k=begin; k<end; ++k)
-        lambda(k, local_tmp);
-        });
-
-#ifdef KOKKOS_ENABLE_CUDA
-      // Broadcast result to all threads by doing sum of one thread's
-      // non-0 value and the rest of the 0s.
-      Kokkos::Impl::CudaTeamMember::vector_reduce(Kokkos::Sum<ValueType>(local_tmp));
-#endif
-
-      result = local_tmp;
-    }
-    else {
-      Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, begin, end), lambda, result);
-    }
+    impl::parallel_reduce<TeamMember, Lambda, ValueType, Serialize>(team, begin, end, lambda, Serialize, result);
   }
 
   template<typename PackType, typename ValueType, bool Serialize =
@@ -138,13 +153,11 @@ struct ExeSpaceUtils {
   void reduce_add (const PackType& p,
                    ValueType& result)
   {
-    if (Serialize)
-    {
-      for (int i = 0; i < PackType::n; ++i)
+    if (Serialize) {
+      for (int i = 0; i < PackType::n; ++i) {
         result += p[i];
-    }
-    else
-    {
+      }
+    } else {
       typename PackType::scalar sum = p[0];
       vector_simd for (int i = 1; i < PackType::n; ++i) sum += p[i];
       result += sum;
@@ -187,7 +200,7 @@ struct ExeSpaceUtils<Kokkos::Cuda> {
                         const Lambda& lambda,
                         ValueType& result)
   {
-    ExeSpaceUtils<Kokkos::DefaultExecutionSpace>::parallel_reduce(team, begin, end, lambda, result);
+    impl::parallel_reduce<TeamMember, Lambda, ValueType, Serialize>(team, begin, end, lambda, result);
   }
 
   template<typename PackType, typename ValueType, bool Serialize =
--- a/src/ekat/kokkos/ekat_kokkos_utils.hpp
+++ b/src/ekat/kokkos/ekat_kokkos_utils.hpp
@@ -95,7 +95,6 @@ struct ExeSpaceUtils {
   {
     if (Serialize)
     {
-      std::cout << "Serialize" << std::endl;
       // We want to get C++ on GPU to match F90 on CPU. Thus, need to
       // serialize parallel reductions.
 
@@ -124,7 +123,6 @@ struct ExeSpaceUtils {
       result = local_tmp;
     }
     else {
-      std::cout << "Not Serialize" << std::endl;
       Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, begin, end), lambda, result);
     }
   }
--- a/src/ekat/kokkos/ekat_kokkos_utils.hpp
+++ b/src/ekat/kokkos/ekat_kokkos_utils.hpp
@@ -106,7 +106,7 @@ reshape (Kokkos::View<DataTypeIn,Props...> view_in,
  * for a parallel kernel. On non-GPU archictures, we will generally have
  * thread teams of 1.
  */
-  template <bool Serialize, typename ExeSpace = Kokkos::DefaultExecutionSpace>
+template <bool Serialize, typename ExeSpace = Kokkos::DefaultExecutionSpace>
 struct ExeSpaceUtils {
   using TeamPolicy = Kokkos::TeamPolicy<ExeSpace>;
 
@@ -144,7 +144,7 @@ struct ExeSpaceUtils {
  */
 #ifdef KOKKOS_ENABLE_CUDA
 template <bool Serialize>
-struct ExeSpaceUtils<Kokkos::Cuda> {
+struct ExeSpaceUtils<Serialize,Kokkos::Cuda> {
   using TeamPolicy = Kokkos::TeamPolicy<Kokkos::Cuda>;
 
   static TeamPolicy get_default_team_policy (Int ni, Int nk) {
--- a/src/ekat/kokkos/ekat_kokkos_utils.hpp
+++ b/src/ekat/kokkos/ekat_kokkos_utils.hpp
@@ -139,7 +139,7 @@ struct ExeSpaceUtils {
                         const Lambda& lambda,
                         ValueType& result)
   {
-    impl::parallel_reduce<TeamMember, Lambda, ValueType, Serialize>(team, begin, end, lambda, Serialize, result);
+    impl::parallel_reduce<TeamMember, Lambda, ValueType, Serialize>(team, begin, end, lambda, result);
   }
 
   template<typename PackType, typename ValueType, bool Serialize =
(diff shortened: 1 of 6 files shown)


src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -77,6 +78,60 @@ struct ExeSpaceUtils {
   static TeamPolicy get_team_policy_force_team_size (Int ni, Int team_size) {
     return TeamPolicy(ni, team_size);
   }
+
+  template <typename TeamMember, class Lambda, typename ValueType>
+  static KOKKOS_INLINE_FUNCTION
+  void parallel_reduce (const TeamMember& team,
+			const int& begin,
```
ambrad, line 85: Always use spaces: never tabs.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -77,6 +78,60 @@ struct ExeSpaceUtils {
...
+			const int& end,
+                        const Lambda& lambda,
+                        const bool serialize,
```
ambrad, line 88: We definitely do not want this as a run-time argument. One suggestion is to make it a template parameter with default value based on EKAT's BFBness configuration.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -97,6 +152,66 @@ struct ExeSpaceUtils<Kokkos::Cuda> {
   static TeamPolicy get_team_policy_force_team_size (Int ni, Int team_size) {
     return TeamPolicy(ni, team_size);
   }
+
+
+
+  template <typename TeamMember, class Lambda, typename ValueType/*, bool Serialize=true*/>
+  static KOKKOS_INLINE_FUNCTION
+  void parallel_reduce (const TeamMember& team,
```
ambrad, line 160: This specialization's impl is identical to the general one's impl. So factor this code out into an impl function nested in an `impl` namespace that each can call.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -97,6 +152,66 @@ struct ExeSpaceUtils<Kokkos::Cuda> {
...
+      // non-0 value and the rest of the 0s.
+      Kokkos::Impl::CudaTeamMember::vector_reduce(Kokkos::Sum<ValueType>(local_tmp));
+
+      result = local_tmp;
+    }
+    else
+      Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, begin, end), lambda, result);
+  }
+
+  template<typename PackType, typename ValueType, bool Serialize=false>
```
ambrad, line 198: This doesn't address the primary need. Luca suggested looking at https://github.com/E3SM-Project/E3SM/blob/b7b269b3cfaca869e861fa01e6d65e201dc588a3/components/homme/src/share/cxx/ColumnOps.hpp#L474, and I agree that's the best approach. The minor issue of providing a reduction over a single pack can wait for a future PR, and when it's done, it will live in ekat_pack.hpp, since it has nothing to do with Kokkos.

tcclevenger: Ok, yes I wasn't sure since the suggestion in the issue listed only the single pack version, but I will implement this tomorrow.

ambrad: Please see https://github.com/E3SM-Project/EKAT/issues/26#issuecomment-688989928 and subsequent comments.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -77,6 +78,60 @@ struct ExeSpaceUtils {
...
+  static KOKKOS_INLINE_FUNCTION
+  void reduce_add (const PackType& p,
```
bartgol, line 120: This should probably be in the ekat_pack.hpp file. I would also add it as a method of the pack class, actually, but free function is probably ok.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -77,6 +78,60 @@ struct ExeSpaceUtils {
...
+  {
+    if (Serialize)
+    {
+      for (int i = 0; i < PackType::n; ++i)
+        result += p[i];
+    }
+    else
+    {
+      typename PackType::scalar sum = p[0];
+      vector_simd for (int i = 1; i < PackType::n; ++i) sum += p[i];
```
bartgol, line 131: It might be interesting to write a micro reproducer for a simd reduction (without ekat, just plain cpp), and check if vector_simd vectorizes. If not, try with something like

```
#define vector_simd_reduce(op,result) _Pragma("omp simd reduction(op:result)")
```
(if openmp is enabled, and expand to nothing otherwise). I suspect `vector_simd` might not always vectorize, while the omp simd reduction will. Or, at least, the latter will vectorize _more_ times.

tests/kokkos/kokkos_utils_tests.cpp
```diff
@@ -194,4 +196,106 @@ TEST_CASE("team_utils_large_ni", "[kokkos_utils]")
...
+  Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const MemberType& team) {
+    const int t = team.league_rank();
+    Scalar team_result = Scalar();
+
+    const int begin = t*work_per_team;
+    const int end = begin + work_per_team;
+    ekat::ExeSpaceUtils<ExeSpace>::parallel_reduce(team, begin, end,
+        [&] (const int k, Scalar& reduction_value) {
+                 reduction_value += data[k];
+               }, true/*serialize*/, team_result);
```
bartgol, line 246: You might consider adding a short "fake" test for serialize=false. Not to check the answer (will be non deterministic), but to check stuff compiles and runs without issues.

tests/kokkos/kokkos_utils_tests.cpp
```diff
@@ -194,4 +196,106 @@ TEST_CASE("team_utils_large_ni", "[kokkos_utils]")
...
+  PackType pack;
+  for (int i=0; i<PackType::n; ++i)
+  {
+    Scalar pack_val = 1.0/(i+1);
+    serial_result += pack_val;
+    pack[i] = pack_val;		 
+  }
+
+  ekat::ExeSpaceUtils<ExeSpace>::reduce_add<PackType,Scalar,true/*serialize*/>(pack,reduce_add_result);
+  REQUIRE(reduce_add_result == serial_result);
```
bartgol, line 293: We should test both serialize and not serialize. However, is the vector reduction guaranteed to give the same result if serialize=false? I suspect not. However, if your input contains powers of two only (of not too different exponents), then the floating point ops are commutative. You could for instance check that a pack containing `[1, 0.5, 0.25, ..., 0.5^N-1]` sums up to `(1-0.5^N)/(1-0.5)` (I hope I didn't mess up the math formula, but you get the gist). This sum should always be exact (at least in single/double precision, if we ever add half precision, we may have to limit pack size N to something like 10).

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -77,6 +78,60 @@ struct ExeSpaceUtils {
...
+  template<typename PackType, typename ValueType, bool Serialize=false>
+  static KOKKOS_INLINE_FUNCTION
+  void reduce_add (const PackType& p,
+                   ValueType& result)
+  {
+    if (Serialize)
+    {
+      for (int i = 0; i < PackType::n; ++i)
+        result += p[i];
+    }
```
bartgol, line 127: This is me being pedantic (read: feel free to ignore this), but in the rest of scream, we put opening braces on the same level of the if/for statements, and else statements appear as `} else {` (one line).

Not a big deal, feel free to disregard.

bartgol (changes requested):
The implementations look good, but I have a few requests. Feel free to push back on what you disagree.

src/ekat/kokkos/ekat_kokkos_utils.hpp
tcclevenger: I've added it back as a template parameter and set based on `#ifdef EKAT_TEST_STRICT_FP`. But `EKAT_TEST_STRICT_FP` is only defined in the testing directory. Should I make it so that the definition is available everywhere?

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -25,6 +26,52 @@
 
 namespace ekat {
 
+namespace impl {
+
+  template <typename TeamMember, class Lambda, typename ValueType, bool Serialize>
```
bartgol, line 31: The standard specifies that there is no semantic difference between `class` and `typename` for (typep) template arguments (except one special case in pre C++17, but that doesn't apply here).

However, I think it _might_ be misleading for a reader (that doesn't know what I said above) to see a mix of typename and class, and he/she/they might be led to think that there are differences between the two. I would prefer to use the same keyword (either class or typename) for all the (type) template argument. At least, let's be consistent within the same `template<..>` line (consistency across different uses of `template` is less of an issue).

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -9,6 +9,7 @@
 #include "ekat/ekat_assert.hpp"
 #include "ekat/ekat_type_traits.hpp"
 #include "ekat/ekat.hpp"
+#include "ekat/ekat_macros.hpp"
```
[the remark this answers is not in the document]
tcclevenger: We need this for `vector_simd` in `reduce_add()`. But depending on what happens with that function (may move to `ekat_pack.hpp`) I'll remove it if necessary.

src/ekat/kokkos/ekat_kokkos_utils.hpp
tcclevenger: If you don't mind, I'll implement there and then add the `column_reduce()` function here as discussed [here](https://github.com/E3SM-Project/EKAT/issues/26#issuecomment-688989928)

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -97,6 +183,39 @@ struct ExeSpaceUtils<Kokkos::Cuda> {
...
     return TeamPolicy(ni, team_size);
   }
+
+
+
+  template <typename TeamMember, class Lambda, typename ValueType, bool Serialize =
+ #ifdef EKAT_TEST_STRICT_FP
+        true
+ #else
+        true //This is a hack for now since EKAT_TEST_STICT_FP is not recognized in src/
```
bartgol, line 193: I would remove the default for `Serialize`. Ekat is a library, and defaults in libraries can bite you later on. Let's have the downstream customers explicitly pass the serialize argument. If they are lazy, they can always do

```
template<...>
KOKKOS_INLINE_FUNCTION
void parallel_reduce(...) {
#ifdef SOME_BFB_MACRO
   ekat::parallel_reduce<...,true>(...);
#else
   ekat::parallel_reduce<...,false>(...);
#endif
}
```

tests/kokkos/kokkos_utils_tests.cpp
tcclevenger: On this test, I added a template parameter for whether or not the `REQUIRE(results_h(0) == serial_result);` is run based on `EKAT_TEST_STRICT_FP`. So on a non-bfb run the test runs just without the failing assertion. Is this acceptable? I'm can definitely change this.

src/ekat/kokkos/ekat_kokkos_utils.hpp
bartgol: @tcclevenger I added a comment inside the code, but since it's discussed here: let's remove the default altogether, and have downstream app explicitly pass the last one. Apps can always do

```
#ifdef SOME_BFB_MACRO
constexpr bool SerialReduction = true;
#else
constexpr bool SerialReduction = false;
#endif
```

src/ekat/kokkos/ekat_kokkos_utils.hpp
bartgol: Ah, if you remove the default (which I would support), then you should move the bool flag to the beginning of the template args list, so that one can do

```
ekat::ExecSpaceUtils<ExecSpace>::parallel_reduce<false>(...);
```
and have all other template args deduced.

src/ekat/kokkos/ekat_kokkos_utils.hpp
bartgol: It can go both ways (single PR vs two PRs). Multiple PRs make code review faster, but I don't feel strongly either way.

tcclevenger: I'll add it here

tests/kokkos/kokkos_utils_tests.cpp
bartgol: I am not a fan of `EKAT_TEST_STRICT_FP`. Ekat offers some features, that should work (perhaps following different code paths) regardless. We should test both a "strict" code flow, as well as a "release" code flow. Obviously, the checks that need to occur are different (like you suggested, not running the REQUIRE clause if not a strict run).

That said, I don't see where the extra tepmlate parameter is. Maybe you haven't pushed yet? Either way, let's make sure that the function is executed with both serialize true and false.

src/ekat/kokkos/ekat_kokkos_utils.hpp
tcclevenger: This is being moved to  a new PR

src/ekat/kokkos/ekat_kokkos_utils.hpp
tcclevenger: Moved to a new PR

tcclevenger:
@bartgol @ambrad: Talking with Luca, I decided to move the column reduction to a new PR (I will make later today). This is now only for `parallel_reduce()`. I've taken your comments into account. The test now passes on GPU and CPU, and only when `serialize=true` (the test also runs `serialize=false`, but has no `REQUIRE()` call). Let me know if you have any other comments.

tests/kokkos/kokkos_utils_tests.cpp
```diff
@@ -194,4 +194,61 @@ TEST_CASE("team_utils_large_ni", "[kokkos_utils]")
...
+template<typename Scalar, int length, bool Serialize>
+void test_parallel_reduce()
+{
+  using Device = ekat::DefaultDevice;
+  using MemberType = typename ekat::KokkosTypes<Device>::MemberType;
+  using ExeSpace = typename ekat::KokkosTypes<Device>::ExeSpace;
+
+#ifdef KOKKOS_ENABLE_OPENMP
+  const int n = omp_get_max_threads();
+  // test will not work with more than 16 threads
```
ambrad, line 206: Why?

tests/kokkos/kokkos_utils_tests.cpp
```diff
@@ -194,4 +194,61 @@ TEST_CASE("team_utils_large_ni", "[kokkos_utils]")
...
+              reduction_value += data[k];
+        }, team_result);
+
+      results(0) = team_result;
+    });
+
+  Kokkos::deep_copy(results_h, results);
+  if (Serialize) {
+    REQUIRE(results_h(0) == serial_result);
+  }
```
ambrad, line 245: Would be good, if not Serialize, to have a tol-based requirement, something like `<= 10*std::numeric_limits<Scalar>::epsilon()`.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -25,6 +25,52 @@
 
 namespace ekat {
 
+namespace impl {
+
+template <bool Serialize, typename TeamMember, typename Lambda, typename ValueType>
```
ambrad, line 30: @bartgol I'm concerned about cluttering code with the Serialize parameter. Would EKAT support putting this last with a config-time default, something like `template <..., bool Serialize=EKAT_BFB_OPS>`?

Another option is SCREAM can have a wrapper to this wrapper that handles the Serialize parameter so the actual app doesn't have to provide any serialize argument to `parallel_reduce`.

tests/kokkos/kokkos_utils_tests.cpp
tcclevenger: I copy and pasted from an existing test above and this was part of it. I can remove it.

ambrad: Please do. Also, when copy-pasting code, always study the code closely, as otherwise poor assumptions and bugs will propagate.

src/ekat/kokkos/ekat_kokkos_utils.hpp
bartgol: I wrote it as a comment somewhere, but it got lost with the push. Bottom line is that I'd like to not put bfb macros in ekat. TBC, it's just my preference. I just think of need for bfb as a "customer need" (to be addressed in the app), not a "provider service" (to appear in the library). Especially if the macro had a very limited number of places where used (like, in this case, one).

One limitation of adding the macro in EKAT is that it forces _all_ use cases to do the same thing. Although this might not be a big deal in practice, it is a limitation of sort, that can easily be avoided.

That said, as I wrote in my (lost) original comment, scream can easily define a wrapper function, with CPP logic to pick the instantiation.

bartgol: Another solution could be to add the `Serialize` template argument directly to the class, so that customers can do

```
template<typename ExecSpace>
#ifdef BFB
using ExecSpaceUtils = ekat::ExecSpaceUtils<ExecSpace,false>;
#else
using ExecSpaceUtils = ekat::ExecSpaceUtils<ExecSpace,false>;
#endif
```
so that the CPP logic is done once for all exec utils.

ambrad: Luca, that's fine. Tbc, I strongly disagree. EKAT should provide solutions to exactly what the common needs are, including handling BFB-testing needs. But I defer to yours on matters of opinion. Adding some type of wrapper in the app is fine.

tests/kokkos/kokkos_utils_tests.cpp
bartgol: What about using a sum that can be done exactly in FP arithmetic? Like `1+0.5+0.25+...+0.5^N`?

ambrad: Or just do it over int instead of FP.

ambrad (approved):

src/ekat/kokkos/ekat_kokkos_utils.hpp
jeff-cohere: Personally, I'm on the fence about including BFB stuff in EKAT, because I loathe the methodology and prefer libraries with a light touch. But the need for BFB reflects the reality of our work. So I could go either way at the moment (stick it in now, or wait for folks to ask for EKAT support for it).

tests/kokkos/kokkos_utils_tests.cpp
jeff-cohere: I like this tolerance approach, as it exercises the behavior of a reduction under "normal" non-BFB circumstances.

jeff-cohere (dismissed):
Sorry I'm late to the party on this one. I didn't have much to add to what Andrew and Luca already said. Nice work.

tests/kokkos/kokkos_utils_tests.cpp
tcclevenger: The reason I went with a sum that is not bfb-exact was to just verify that the serial algorithm did the correct thing. Then using tolerance for parallel is just more convenient so that  I don't need to implement 2 version of the test.

bartgol: I guess in this case (with fixed inputs) a tol approach is just fine, since you can find a tol that is guaranteed to work. With random inputs, we got bitten in the past for running sometimes into the unlucky case where 1ulp propagates up the mantissa (in case of triggering several carries). So I'm fine with this.

src/ekat/kokkos/ekat_kokkos_utils.hpp
bartgol: I opened an issue, let's discuss.

bartgol (dismissed):

tcclevenger:
@bartgol @ambrad @jeff-cohere I decided to make `Serialize` an template parameter of `ExeSpaceUtils` since I think this will make using these functions in scream the simplest (Luca's #ifdef example above). A lot more files are now changed since everyone who calls the struct needs a new template. 

The only issue is that, because the ExeSpace template is defaulted, `Serialize` had to come first. Is this acceptable, or should I put `Serialize` 2nd with a default value?

ambrad:
@tcclevenger I like this.
1. @bartgol, what do you think if we default `Serialize=false`? That would let us provide the user-expected default w/o yet making a decision on EKAT's support for BFB flags. It would also mean, for example, that all these additional file changes wouldn't be needed.
2. I think putting Serialize first relative to ExecSpace is fine, regardless of 1.

bartgol:
I don't have a strong opinion on defaulting Serialize to false. It is definitely the safest default, I guess, but it would not solve the problem of downstream apps having to specify the argument everywhere, or, at least, in every use case that will be part of a BFB test, no? I mean, if you are using `ExecSpaceUtils` in a function that needs to undergo BFB testing, you cannot rely on the default, but need to explicitly select its value. It would help for those use cases that we don't foresee ever needing BFB tests, but I'm not sure they are many.

Besides, given the current feedback on having EKAT_BFB_DEFAULTS (or similar), I think we will end up with a default that is set at config-time.

Imho, let's leave no default now, and probably bring configurable default in soon in another PR.

tcclevenger:
Sounds good. It looks like the it needs to be approved again before being able to merge. 

Also, do you like to squash your commits, or it good as is?

ambrad (approved):

PR #36, opened 2020-09-12
fixes CMake bug for make install

### LINKED ISSUE #28: Add an installation (make install) target

pbosler:
## Motivation

Fixes CMake bug that prevented `ekat/util` headers from being installed, which in turn prevented client codes from being able to find them at `${CMAKE_INSTALL_PREFIX}/include`

Closes #28 

## Testing

Tested on Mac OS X.


jeff-cohere (approved):
Nice catch.

PR #37, opened 2020-09-14
add view_reduction

tcclevenger:
## Motivation
From the discussion in https://github.com/E3SM-Project/EKAT/issues/26, this adds a `view_reduction()` function which computes a complete reduction over a view of packs, with both a serial and parallel implementation. 

For this pull request, a `reduce_sum()` function was also added which computes `a+sum(pack)`. 

Tests for each of these functions were also added.

## Related Issues
This is to address a need in scream, e.g., this pull request: https://github.com/E3SM-Project/scream/pull/536

## Testing
Tests were written for both functions which confirm bfb-ness.

### DIFF
--- a/src/ekat/ekat_pack.hpp
+++ b/src/ekat/ekat_pack.hpp
@@ -327,7 +327,7 @@ void reduce_sum (const PackType& p, typename PackType::scalar& sum) {
 // return sum = p[0]+p[1]+...
 template <bool Serialize, typename PackType> KOKKOS_INLINE_FUNCTION
 OnlyPackReturn<PackType, typename PackType::scalar> reduce_sum (const PackType& p) {
-  typename PackType::scalar sum = PackType::scalar();
+  typename PackType::scalar sum = typename PackType::scalar(0);
   reduce_sum<Serialize>(p,sum);
   return sum;
 }
--- a/src/ekat/ekat_pack.hpp
+++ b/src/ekat/ekat_pack.hpp
@@ -38,7 +38,7 @@ struct Mask {
   enum { n = PackSize };
 
   KOKKOS_FORCEINLINE_FUNCTION
-  explicit Mask () {}
+  Mask () {}
 
   // Init all slots of the Mask to 'init'.
   KOKKOS_FORCEINLINE_FUNCTION explicit Mask (const bool& init) {
@@ -156,7 +156,7 @@ struct Pack {
   typedef typename std::remove_const<ScalarType>::type scalar;
 
   KOKKOS_FORCEINLINE_FUNCTION
-  explicit Pack () {
+  Pack () {
     vector_simd for (int i = 0; i < n; ++i) {
       d[i] = ScalarTraits<scalar>::invalid();
     }
--- a/src/ekat/ekat_pack.hpp
+++ b/src/ekat/ekat_pack.hpp
@@ -314,6 +314,22 @@ OnlyPackReturn<PackType, typename PackType::scalar> max (const PackType& p) {
   return v;
 }
 
+// return a+p[0]+p[1]+...
+template <bool Serialize, typename PackType, typename Scalar> KOKKOS_INLINE_FUNCTION
+OnlyPackReturn<PackType, typename PackType::scalar> reduce_sum (const PackType& p, const Scalar a=Scalar()) {
+  Scalar return_val = Scalar(a);
+  if (Serialize) {
+    for (int i = 0; i < PackType::n; ++i) {
+      return_val += p[i];
+    }
+  } else {
+    typename PackType::scalar sum = p[0];
+    vector_simd for (int i = 1; i < PackType::n; ++i) sum += p[i];
+    return_val += sum;
+  }
+  return return_val;
+}
+
 // min(init, min(p(mask)))
 template <typename PackType> KOKKOS_INLINE_FUNCTION
 OnlyPackReturn<PackType, typename PackType::scalar>
--- a/src/ekat/ekat_pack.hpp
+++ b/src/ekat/ekat_pack.hpp
@@ -314,20 +314,22 @@ OnlyPackReturn<PackType, typename PackType::scalar> max (const PackType& p) {
   return v;
 }
 
-// return a+p[0]+p[1]+...
-template <bool Serialize, typename PackType, typename Scalar> KOKKOS_INLINE_FUNCTION
-OnlyPackReturn<PackType, typename PackType::scalar> reduce_sum (const PackType& p, const Scalar a=Scalar()) {
-  Scalar return_val = Scalar(a);
+// sum = sum+p[0]+p[1]+...
+template <bool Serialize, typename PackType> KOKKOS_INLINE_FUNCTION
+void reduce_sum (const PackType& p, typename PackType::scalar& sum) {
   if (Serialize) {
-    for (int i = 0; i < PackType::n; ++i) {
-      return_val += p[i];
-    }
+    for (int i = 0; i < PackType::n; ++i) sum += p[i];
   } else {
-    typename PackType::scalar sum = p[0];
-    vector_simd for (int i = 1; i < PackType::n; ++i) sum += p[i];
-    return_val += sum;
+    vector_simd for (int i = 0; i < PackType::n; ++i) sum += p[i];
   }
-  return return_val;
+}
+
+// return sum = p[0]+p[1]+...
+template <bool Serialize, typename PackType> KOKKOS_INLINE_FUNCTION
+OnlyPackReturn<PackType, typename PackType::scalar> reduce_sum (const PackType& p) {
+  typename PackType::scalar sum = PackType::scalar();
+  reduce_sum<Serialize>(p,sum);
+  return sum;
 }
 
 // min(init, min(p(mask)))
--- a/src/ekat/kokkos/ekat_kokkos_types.hpp
+++ b/src/ekat/kokkos/ekat_kokkos_types.hpp
@@ -71,4 +71,13 @@ struct KokkosTypes
 
 } // namespace ekat
 
+// Kokkos-compatible reduction identity for arbitrary packs
+namespace Kokkos {
+template<typename PackType>
+struct reduction_identity {
+  KOKKOS_FORCEINLINE_FUNCTION
+  static PackType sum()  {return PackType(typename PackType::scalar(0));}
+};
+}
+
 #endif // EKAT_KOKKOS_TYPES_HPP
--- a/src/ekat/kokkos/ekat_kokkos_types.hpp
+++ b/src/ekat/kokkos/ekat_kokkos_types.hpp
@@ -71,13 +71,4 @@ struct KokkosTypes
 
 } // namespace ekat
 
-// Kokkos-compatible reduction identity for arbitrary packs
-namespace Kokkos {
-template<typename PackType>
-struct reduction_identity {
-  KOKKOS_FORCEINLINE_FUNCTION
-  static PackType sum()  {return PackType(0);}
-};
-}
-
 #endif // EKAT_KOKKOS_TYPES_HPP
--- a/src/ekat/kokkos/ekat_kokkos_types.hpp
+++ b/src/ekat/kokkos/ekat_kokkos_types.hpp
@@ -76,7 +76,7 @@ namespace Kokkos {
 template<typename PackType>
 struct reduction_identity {
   KOKKOS_FORCEINLINE_FUNCTION
-  static PackType sum()  {return PackType(typename PackType::scalar(0));}
+  static PackType sum()  {return PackType(0);}
 };
 }
 
--- a/src/ekat/kokkos/ekat_kokkos_utils.hpp
+++ b/src/ekat/kokkos/ekat_kokkos_utils.hpp
@@ -28,10 +28,15 @@ namespace ekat {
 
 namespace impl {
 
-template <bool Serialize, typename TeamMember, typename LoopBoundaryType, typename Lambda, typename ValueType>
+// Computes a general parallel reduction. If Serialize=true, this reduction is computed
+// one element in order, useful for BFB testing with serial routines.
+// If Serialize=false, this function simply calls Kokkos::parallel_reduce().
+// If result contains a value, the output is result+reduction
+template <bool Serialize, typename TeamMember, typename Lambda, typename ValueType>
 static KOKKOS_INLINE_FUNCTION
 void parallel_reduce (const TeamMember& team,
-                      const LoopBoundaryType& loop_boundaries,
+                      const int& begin,
+                      const int& end,
                       const Lambda& lambda,
                       ValueType& result)
 {
@@ -48,10 +53,10 @@ void parallel_reduce (const TeamMember& team,
     //       see the updated value before the vector_reduce call,
     //       which will cause the final answer to be multiplied by the
     //       size of the warp.
-    auto local_tmp = Kokkos::reduction_identity<ValueType>::sum();
+    auto local_tmp = result;
 
     Kokkos::single(Kokkos::PerThread(team),[&] {
-        for (int k=loop_boundaries.start; k<loop_boundaries.end; ++k) {
+        for (int k=begin; k<end; ++k) {
           lambda(k, local_tmp);
         }
       });
@@ -64,27 +69,39 @@ void parallel_reduce (const TeamMember& team,
 
    result = local_tmp;
   } else {
-    Kokkos::parallel_reduce(loop_boundaries, lambda, result);
+    const ValueType initial(result);
+    Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, begin, end), lambda, result);
+    result += initial;
   }
 }
 
-template <bool Serialize, int TotalSize, int VectorSize, typename TeamMember, typename InputProvider, typename ValueType>
+/*
+ * Computes a reduction over the view described by 'input' for entries 'input([begin,end))'
+ * The variable 'input' should be an operator s.t. 'input(k)' returns a Pack of size 'vector_size'.
+ * The variable 'total_size' represents the total number of 'ValueType' entries in the view. The
+ * computed reduction is added to 'result'.
+ */
+template <bool Serialize, typename TeamMember, typename InputProvider, typename ValueType>
 static KOKKOS_INLINE_FUNCTION
 void view_reduction (const TeamMember& team,
-                     const int& begin,
-                     const int& end,
+                     const int total_size,
+                     const int vector_size,
+                     const int begin,
+                     const int end,
                      const InputProvider& input,
                      ValueType& result)
 {
-  if (Serialize) {
-    const int  n_packs        = end - begin;
-    const bool has_garbage    = (TotalSize % VectorSize == 0 ? false : true);
-    const int  last_pack_size = (has_garbage ? TotalSize % VectorSize : VectorSize);
-    const int  loop_length    = (has_garbage ? n_packs-1 : n_packs);
+  const int  n_packs        = (total_size + vector_size - 1)/vector_size;
+  const bool has_garbage    = (end==n_packs && total_size % vector_size != 0 ? true : false);
+  const int  loop_end       = (has_garbage ? end-1 : end);
+  const int  last_pack_size = (has_garbage ? total_size % vector_size : vector_size);
 
-    impl::parallel_reduce<Serialize>(team,Kokkos::ThreadVectorRange(team, loop_length),
+  // For parallel computation, perform a packed reduction
+  if (!Serialize) {
+    impl::parallel_reduce<Serialize>(team, begin, loop_end,
                                      [&](const int k, ValueType& local_sum) {
-        local_sum = ekat::reduce_sum<Serialize>(input(k),local_sum);
+        // Sum over pack entries and add to local_sum
+        ekat::reduce_sum<Serialize>(input(k),local_sum);
     }, result);
       
     // If last pack has garbage, we did not include it in the previous reduction,
@@ -92,16 +109,21 @@ void view_reduction (const TeamMember& team,
     if (has_garbage) {
       Kokkos::single(Kokkos::PerTeam(team),[&] {
         for (int j=0; j<last_pack_size; ++j) {
-          result += input(begin+n_packs-1)[j];
+          result += input(end-1)[j];
         }
       });
     }
   } else {
-    impl::parallel_reduce<Serialize>(team,Kokkos::ThreadVectorRange(team, TotalSize),
+    // For the serial computation, just loop over the total number of pack elements
+    // and determine their place in input
+    const int total_start = begin*vector_size;
+    const int total_end = (has_garbage ? loop_end*vector_size + last_pack_size : loop_end*vector_size);
+
+    impl::parallel_reduce<Serialize>(team,total_start,total_end,
                                      [&](const int k, ValueType& local_sum) {
-      if (VectorSize>1) {
-        const int ilev = k / VectorSize;
-        const int ivec = k % VectorSize;
+      if (vector_size > 1) {
+        const int ilev = k / vector_size;
+        const int ivec = k % vector_size;
         local_sum += input(ilev)[ivec];
       } else {
         local_sum += input(k)[0];
@@ -165,25 +187,28 @@ struct ExeSpaceUtils {
     return TeamPolicy(ni, team_size);
   }
 
-  template <typename TeamMember, typename LoopBoundaryType, typename Lambda, typename ValueType>
+  template <typename TeamMember, typename Lambda, typename ValueType>
   static KOKKOS_INLINE_FUNCTION
   void parallel_reduce (const TeamMember& team,
-                        const LoopBoundaryType& loop_boundaries,
+                        const int& begin,
+                        const int& end,
                         const Lambda& lambda,
                         ValueType& result)
   {
-    impl::parallel_reduce<Serialize, TeamMember, LoopBoundaryType, Lambda, ValueType>(team, loop_boundaries, lambda, result);
+    impl::parallel_reduce<Serialize, TeamMember, Lambda, ValueType>(team, begin, end, lambda, result);
   }
 
-  template <int TotalSize, int VectorSize, typename TeamMember, typename InputProvider, typename ValueType>
+  template <typename TeamMember, typename InputProvider, typename ValueType>
   static KOKKOS_INLINE_FUNCTION
   void view_reduction (const TeamMember& team,
+                       const int& total_size,
+                       const int& vector_size,
                        const int& begin,
                        const int& end,
                        const InputProvider& input,
                        ValueType& result)
   {
-    impl::view_reduction<Serialize,TotalSize,VectorSize,TeamMember,InputProvider,ValueType>(team,begin,end,input,result);
+    impl::view_reduction<Serialize,TeamMember,InputProvider,ValueType>(team,total_size,vector_size,begin,end,input,result);
   }
 };
 
@@ -206,25 +231,28 @@ struct ExeSpaceUtils<Serialize,Kokkos::Cuda> {
     return TeamPolicy(ni, team_size);
   }
 
-  template <typename TeamMember, typename LoopBoundaryType, typename Lambda, typename ValueType>
+  template <typename TeamMember, typename Lambda, typename ValueType>
   static KOKKOS_INLINE_FUNCTION
   void parallel_reduce (const TeamMember& team,
-                        const LoopBoundaryType& loop_boundaries,
+                        const int& begin,
+                        const int& end,
                         const Lambda& lambda,
                         ValueType& result)
   {
-    impl::parallel_reduce<Serialize, TeamMember, LoopBoundaryType, Lambda, ValueType>(team, loop_boundaries, lambda, result);
+    impl::parallel_reduce<Serialize, TeamMember, Lambda, ValueType>(team, begin, end, lambda, result);
   }
 
-  template <int TotalSize, int VectorSize, typename TeamMember, typename InputProvider, typename ValueType>
+  template <typename TeamMember, typename InputProvider, typename ValueType>
   static KOKKOS_INLINE_FUNCTION
   void view_reduction (const TeamMember& team,
+                       const int& total_size,
+                       const int& vector_size,
                        const int& begin,
                        const int& end,
                        const InputProvider& input,
                        ValueType& result)
   {
-    impl::view_reduction<Serialize,TotalSize,VectorSize,TeamMember,InputProvider,ValueType>(team,begin,end,input,result);
+    impl::view_reduction<Serialize,TeamMember,InputProvider,ValueType>(team,total_size,vector_size,begin,end,input,result);
   }
 };
 #endif
--- a/src/ekat/kokkos/ekat_kokkos_utils.hpp
+++ b/src/ekat/kokkos/ekat_kokkos_utils.hpp
@@ -24,6 +24,20 @@
 // that are probably not generic enough to appear in kokkos any time soon
 // (or ever), and are more app-specific.
 
+// Kokkos-compatible reduction identity for arbitrary packs
+namespace Kokkos {
+template<typename S, int N>
+struct reduction_identity<ekat::Pack<S,N>> {
+  using PackType = ekat::Pack<S,N>;
+
+  // Provide only sum, since that's our only use case, for now
+  KOKKOS_FORCEINLINE_FUNCTION
+  constexpr static PackType sum() {
+    return PackType (reduction_identity<S>::sum());
+  }
+};
+} // namespace Kokkos
+
 namespace ekat {
 
 namespace impl {
--- a/src/ekat/kokkos/ekat_kokkos_utils.hpp
+++ b/src/ekat/kokkos/ekat_kokkos_utils.hpp
@@ -76,50 +76,62 @@ void parallel_reduce (const TeamMember& team,
 }
 
 /*
- * Computes a reduction over the view described by 'input' for entries 'input([begin,end))'
- * The variable 'input' should be an operator s.t. 'input(k)' returns a Pack of size 'vector_size'.
- * The variable 'total_size' represents the total number of 'ValueType' entries in the view. The
- * computed reduction is added to 'result'.
+ * Computes a reduction over a routine described by 'input' for entries 'scalarize(input)([begin,end))'
+ * The variable 'input' should be an operator s.t. 'input(k)' returns a Pack. The computed reduction is
+ * added to the value 'result'.
  */
 template <bool Serialize, typename TeamMember, typename InputProvider, typename ValueType>
 static KOKKOS_INLINE_FUNCTION
 void view_reduction (const TeamMember& team,
-                     const int total_size,
-                     const int vector_size,
                      const int begin,
                      const int end,
                      const InputProvider& input,
                      ValueType& result)
 {
-  const int  n_packs        = (total_size + vector_size - 1)/vector_size;
-  const bool has_garbage    = (end==n_packs && total_size % vector_size != 0 ? true : false);
-  const int  loop_end       = (has_garbage ? end-1 : end);
-  const int  last_pack_size = (has_garbage ? total_size % vector_size : vector_size);
+  using PackType = typename std::remove_reference<decltype(input(0))>::type;
+  constexpr int vector_size = PackType::n;
 
-  // For parallel computation, perform a packed reduction
-  if (!Serialize) {
-    impl::parallel_reduce<Serialize>(team, begin, loop_end,
-                                     [&](const int k, ValueType& local_sum) {
-        // Sum over pack entries and add to local_sum
-        ekat::reduce_sum<Serialize>(input(k),local_sum);
-    }, result);
-      
-    // If last pack has garbage, we did not include it in the previous reduction,
-    // so manually add the last pack (only the non-garbage part)
-    if (has_garbage) {
+  // Perform a packed reduction over scalar indices
+  if (true) { // !Serialize) {
+    const bool has_garbage_begin = (begin%vector_size != 0 ? true : false);
+    const bool has_garbage_end   = (end%vector_size != 0 ? true : false);
+    const int pack_loop_begin    = (has_garbage_begin ? begin/vector_size + 1 : begin/vector_size);
+    const int pack_loop_end      = end/vector_size;
+
+    // If first pack has garbage, we will not include it in the below reduction,
+    // so manually add the first pack (only the non-garbage part)
+    if (has_garbage_begin) {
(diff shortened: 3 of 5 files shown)


jeff-cohere (dismissed):
This looks good to me. I know we haven't been doing much in this regard, but I think it might be helpful to newcomers to include some brief code comments (in tests as well as implementations) that give newcomers and C++ non-experts a chance to understand what's going on.

I don't mean to turn the tables on you, @tcclevenger . :-) At some point (I hope) we're probably going to start properly annotating functions with Doxygen-like comments anyway (as we should in Doxygen), and it's always easier to do that while we write the functions. Given the fact that a lot of this stuff only exists for optimization purposes, maybe we should start doing that sooner rather than later in ekat.

But this is more of a topic of discussion and less of a comment about this specific PR, so I'll approve, and maybe this will spawn an interesting conversation.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -64,10 +64,51 @@ void parallel_reduce (const TeamMember& team,
...
   } else {
-    Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, begin, end), lambda, result);
+    Kokkos::parallel_reduce(loop_boundaries, lambda, result);
   }
 }
 
+template <bool Serialize, int TotalSize, int VectorSize, typename TeamMember, typename InputProvider, typename ValueType>
+static KOKKOS_INLINE_FUNCTION
+void view_reduction (const TeamMember& team,
+                     const int& begin,
```
bartgol, line 74: If you pass TotalSize as template parameter, do you really need begin/end? My guess is 'no'.

However, I'm wondering if this impl is too restrictive. In particular, one needs to know TotalSize at compile time. For SCREAM/HOMME this is probably the case, but considering that you are not using TotalSize inside constexpr's or template parameters, we could probably make it a runtime parameter, to make the utility more flexible.

I would change the signature to `view_reduction(team, totalSize, input, result)`, with no total size parameter.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -64,10 +64,51 @@ void parallel_reduce (const TeamMember& team,
 
    result = local_tmp;
   } else {
-    Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, begin, end), lambda, result);
+    Kokkos::parallel_reduce(loop_boundaries, lambda, result);
   }
 }
 
+template <bool Serialize, int TotalSize, int VectorSize, typename TeamMember, typename InputProvider, typename ValueType>
```
bartgol, line 71: I think you could deduce VectorSize from the return type of `input(i)`. Inside the body, you could just do

```
using PackType = decltype(input(0));
constexpr int VectorSize = PackType::n; // only works if PackType is a pack
```
or, if you want to be able to handle also views of not pack types,
```
using ValueType = decltype(input(0));
constexpr int VectorSize = sizeof(PackType) / sizeof(ScalarTraits<PackType>::scalar_type);
```
Note: it might be a bit confusing if ValueType is not a Pack (talking of VectorSize for a double seems confusing). So if you DON'T want this function to be used unless the input provider provides Pack's (e.g., a View<Pack<Real,N>*>), you could make the signature like this:
```
template< [...] >
static KOKKOS_INLINE_FUNCTION
auto view_reduction( [...] ) -> OnlyPackReturn<decltype(input(0)),void>
```
so that the function is not considered by the compiler unless input(k) returns a Pack.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -64,10 +64,51 @@ void parallel_reduce (const TeamMember& team,
...
+                     const InputProvider& input,
+                     ValueType& result)
+{
+  if (Serialize) {
+    const int  n_packs        = end - begin;
+    const bool has_garbage    = (TotalSize % VectorSize == 0 ? false : true);
+    const int  last_pack_size = (has_garbage ? TotalSize % VectorSize : VectorSize);
+    const int  loop_length    = (has_garbage ? n_packs-1 : n_packs);
+
+    impl::parallel_reduce<Serialize>(team,Kokkos::ThreadVectorRange(team, loop_length),
```
bartgol, line 85: If begin!=0, then this would be buggy, since the || reduce would be in `[0,loop_length)`. I think you need to explicilty construct the TVR passing `[begin, begin+loop_length)`.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -64,10 +64,51 @@ void parallel_reduce (const TeamMember& team,
...
+    if (has_garbage) {
+      Kokkos::single(Kokkos::PerTeam(team),[&] {
+        for (int j=0; j<last_pack_size; ++j) {
+          result += input(begin+n_packs-1)[j];
+        }
+      });
+    }
+  } else {
+    impl::parallel_reduce<Serialize>(team,Kokkos::ThreadVectorRange(team, TotalSize),
+                                     [&](const int k, ValueType& local_sum) {
```
bartgol, line 101: It could be helpful to implement the non-serialized case with packed logic. In hommexx, the if was on whether to do packed reduction or not. Maybe you wanted `if (!Serialize)` instead of `if (Serialize)'?

tests/kokkos/kokkos_utils_tests.cpp
```diff
@@ -244,4 +245,81 @@ TEST_CASE("parallel_reduce", "[kokkos_utils]")
...
+  using Device = ekat::DefaultDevice;
+  using MemberType = typename ekat::KokkosTypes<Device>::MemberType;
+  using ExeSpace = typename ekat::KokkosTypes<Device>::ExeSpace;
+
+  ekat::ExeSpaceUtils<Serialize,ExeSpace> ExeSpaceUtils; 
+  
+  using PackType = ekat::Pack<Scalar, VectorSize>;
+  using ViewType = Kokkos::View<PackType*,ExeSpace>;
+
+  const int view_length = (int)std::ceil(1.0*TotalSize/VectorSize);
```
bartgol, line 261: If you want to compute the number of packs, you can use the `npack` utility from `ekat_pack.hpp`:
```
const int view_length = npack<PackType>(TotalSize);
```

tests/kokkos/kokkos_utils_tests.cpp
```diff
@@ -244,4 +245,81 @@ TEST_CASE("parallel_reduce", "[kokkos_utils]")
...
+
+  Scalar serial_result = Scalar();
```
bartgol, line 263: Since your impl supports it, it would be good to test the case where the result already contains an initial value, to make sure it is not overwritten. Could be a random value, or some FP number not representable exactly (to avoid surprises), like 1.0/3.0.

tests/kokkos/kokkos_utils_tests.cpp
```diff
@@ -244,4 +245,81 @@ TEST_CASE("parallel_reduce", "[kokkos_utils]")
...
+  ViewType data("data", view_length);
+  const auto data_h = Kokkos::create_mirror_view(data);
+  auto raw = data_h.data();
+  for (int k = 0; k < view_length; ++k) {
+    for (int p=0; p<VectorSize; ++p) {
+      if (k*VectorSize+p >= TotalSize) {
+        // represents pack garbage
+        raw[k][p] = 1e100;
```
bartgol, line 271: You could use `ScalarTraits<Scalar>::invalid()`, which sets garbage to quiet nans.

bartgol (changes requested):
Thanks @tcclevenger ! I have a few comments, let me know what you think.

bartgol:
> This looks good to me. I know we haven't been doing much in this regard, but I think it might be helpful to newcomers to include some brief code comments (in tests as well as implementations) that give newcomers and C++ non-experts a chance to understand what's going on.
> 
> I don't mean to turn the tables on you, @tcclevenger . :-) At some point (I hope) we're probably going to start properly annotating functions with Doxygen-like comments anyway (as we should in Doxygen), and it's always easier to do that while we write the functions. Given the fact that a lot of this stuff only exists for optimization purposes, maybe we should start doing that sooner rather than later in ekat.
> 
> But this is more of a topic of discussion and less of a comment about this specific PR, so I'll approve, and maybe this will spawn an interesting conversation.

I feel like Doxygen has a high surface/volume ratio, with the result that lots of people write a ton of "syntax" comments, and think "ok, i'm done". So I don't really care about Doxygen per se. However, some comments _are_ needed. What I'm trying to say is "let's add some comments explaining what a routine does, or what its (template) arguments are, and possible caveats on them, but let's not worry about doxygen-like syntax".

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -27,11 +28,10 @@ namespace ekat {
 
 namespace impl {
 
-template <typename TeamMember, typename Lambda, typename ValueType, bool Serialize>
+template <bool Serialize, typename TeamMember, typename LoopBoundaryType, typename Lambda, typename ValueType>
 static KOKKOS_INLINE_FUNCTION
 void parallel_reduce (const TeamMember& team,
-                      const int& begin,
-                      const int& end,
+                      const LoopBoundaryType& loop_boundaries,
```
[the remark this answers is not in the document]
tcclevenger: I'm not sure where this function will be used in the future, so it seemed restrictive to require `TeamThreadRange` in the non-serial loop since the lambda is user defined and I guess could be anything. Also, in my implementation (since it was my final `parallel_` call), I decided to use `ThreadVectorRange` instead of `TeamThreadRange`, so I would have needed to change the `parallel_reduce()` method anyways.

I could be misunderstanding best Kokkos practices though so I am completely ok deferring.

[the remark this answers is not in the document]
bartgol: So you want to support both a team and thread level || reduce? That would be nice. Have you tested both?

I'm ok with supporting both, it seems nice. What do you think about adding two "wrapper functions", called `thread_reduce` and `team_reduce`, which take only start/end, and call this function? Worth it? Not worth it?

tests/kokkos/kokkos_utils_tests.cpp
tcclevenger: The implementation of the pack `reduce_sum()` supports it, but the implementation of `view_reduction()` does not. I could implement this by adding a `ValueType a` input like in the pack reduce function.

src/ekat/kokkos/ekat_kokkos_utils.hpp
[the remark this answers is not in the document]
jeff-cohere: I like the flexibility of the mechanism and the wrapper functions idea. The wrappers would reduce the level of abstraction without reducing the ability to control things.

src/ekat/kokkos/ekat_kokkos_utils.hpp
tcclevenger: Yes, I do have this backwards. The tests must have been passing/failing properly due to the `reduce_sum` function.

src/ekat/ekat_pack.hpp
```diff
@@ -314,6 +314,22 @@ OnlyPackReturn<PackType, typename PackType::scalar> max (const PackType& p) {
...
+// return a+p[0]+p[1]+...
+template <bool Serialize, typename PackType, typename Scalar> KOKKOS_INLINE_FUNCTION
+OnlyPackReturn<PackType, typename PackType::scalar> reduce_sum (const PackType& p, const Scalar a=Scalar()) {
+  Scalar return_val = Scalar(a);
+  if (Serialize) {
+    for (int i = 0; i < PackType::n; ++i) {
+      return_val += p[i];
+    }
+  } else {
+    typename PackType::scalar sum = p[0];
```
ambrad, line 326: Probably when we look at the best way to do this eventually, we'll find that sum should be init'ed to 0 and then the loop should be over [0,n).

src/ekat/kokkos/ekat_kokkos_utils.hpp
ambrad: @bartgol , the motivation for [begin,end) is that it's common in P3 and therefore likely in other parameterizations to reduce over just part of a column, with range including neither the top nor bottom. It would be useful for this function to handle the masks against [0,TotalSize).

I agree TotalSize should be a run-time, not compile-time, parameter.

src/ekat/kokkos/ekat_kokkos_utils.hpp
ambrad: Also, why is this on a TVR and not a TTR?

src/ekat/kokkos/ekat_kokkos_utils.hpp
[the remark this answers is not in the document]
ambrad: Flexibility is good, but just to be clear, all TVRs have range size 1 in SCREAM params so far. The immediate need is for TTRs.

src/ekat/kokkos/ekat_kokkos_utils.hpp
tcclevenger: This could be a misunderstanding of Kokkos, but this was the last level of parallelism, so I chose TVR.

src/ekat/ekat_pack.hpp
bartgol: Since `a` is passed by value, we could even recycle `a` as the sum, so we avoid creating an extra tmp.

Maybe, the best impl would be to have two routines:

```
void reduce_sum(const Pack& p, Scalar& sum);

void reduce_sum(const Pack& p);
```
The latter would create a temp for the sum, and then call the former, while the former would sum straight into the sum input.

src/ekat/kokkos/ekat_kokkos_utils.hpp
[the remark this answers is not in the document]
bartgol: @ambrad are you saying params are sticking with policies with vector length 1 right now?

src/ekat/kokkos/ekat_kokkos_utils.hpp
ambrad: Perhaps it will help if I point out that this is operating at exactly the ||ism level that the reductions (and all the k loops) need to be done in SHOC.

src/ekat/kokkos/ekat_kokkos_utils.hpp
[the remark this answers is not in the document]
ambrad: SCREAM and SHOC use nlev Kokkos threads on GPU and Kokkos vector size 1. They use up to (nlev + Pack::n - 1)/Pack::n Kokkos threads on CPU (but in practice just 1) and again Kokkos vector size 1. There's no advantage to Kokkos vector size > 1, and >1 would mean we'd have TVR loops inside TTR loops everywhere, which would add a lot of syntax noise.

For more, see the discussion starting "Re: Kokkos. We need to map out the || loops." on [this page](https://acme-climate.atlassian.net/wiki/spaces/NGDNA/pages/762282136/Micro-app+focused+on+P3+sedimentation).

Other parameterizations are of course free to use non-trivial TVRs. The restriction is only on routines called within the top-level SHOC and P3 kernel dispatches.

tests/kokkos/kokkos_utils_tests.cpp
tcclevenger: I tested this and, in the implementation for both Serial and Parallel, `result` gets overwritten. I could implement  such that it does not get overwritten if you like?

bartgol: Ah, I got confused with `reduce_sum`. Well, since you impl-ed it for `reduce_sum`, it would make sense to impl it also here, no? Same logic should work  here.

src/ekat/ekat_pack.hpp
bartgol: Also, I think you are treating the result differently in the two cases: in the first, the type of  `return_val` is the input template parameter, while in the second is the scalar type of the pack. This could cause different behaviors in the two cases, when `Scalar!=PackType::scalar`.

Perhaps it's best to remove the Scalar template, and deducing it from the PackType?

src/ekat/kokkos/ekat_kokkos_utils.hpp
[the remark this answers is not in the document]
tcclevenger: For now, I changed back to all TTR loops. I agree that the generalization is "better" and I can implement this (tomorrow as part of this PR, or on my to-do-list for future, I'm fine with either).

src/ekat/kokkos/ekat_kokkos_utils.hpp
tcclevenger: I had issues with implementing all these suggestions. For the time being, I have just added `vector_size` as another input parameter.

From the compilation errors, `using PackType = decltype(input(0));` seems to be recognizing the Pack type (`Pack<double, 3>` for example), but then the compiler was saying 
`n` is not a member of `Pack<double, 3>&` 
when calling `PackType::n`. 
Similarly, any call `npack<PackType>(TotalSize);` or `ekat::npack<PackType>(TotalSize);` said it could not deduce the template parameters. Not really sure why the ekat_pack objects aren't being recognized, especially since `reduce_sum()`, defined in the same file, is recognized. Other than it does not have any explicitly defined template params, which must be the difference.

Any suggestions? Or is `vector_size` ok as an input parameter alongside `total_size`.? Let me know if I need to give a better explanation..

src/ekat/kokkos/ekat_kokkos_utils.hpp
tcclevenger: There is now new begin/end logic and I create a test case for each scenario.

tests/kokkos/kokkos_utils_tests.cpp
tcclevenger: I changed the implementation to add the reduction to whatever value is stored in result.

src/ekat/ekat_pack.hpp
tcclevenger: I implemented the two versions of `reduce_sum` you suggested and removed the Scalar template.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -64,10 +69,68 @@ void parallel_reduce (const TeamMember& team,
...
+                     ValueType& result)
+{
+  const int  n_packs        = (total_size + vector_size - 1)/vector_size;
+  const bool has_garbage    = (end==n_packs && total_size % vector_size != 0 ? true : false);
+  const int  loop_end       = (has_garbage ? end-1 : end);
+  const int  last_pack_size = (has_garbage ? total_size % vector_size : vector_size);
+
+  // For parallel computation, perform a packed reduction
+  if (!Serialize) {
+    impl::parallel_reduce<Serialize>(team, begin, loop_end,
```
ambrad, line 101: I don't think this can be right. `begin` is the scalar begin level, just like `end` is the scalar end level. Pack arithmetic must be done on it just like for `end`.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -64,10 +69,68 @@ void parallel_reduce (const TeamMember& team,
...
+ * Computes a reduction over the view described by 'input' for entries 'input([begin,end))'
+ * The variable 'input' should be an operator s.t. 'input(k)' returns a Pack of size 'vector_size'.
+ * The variable 'total_size' represents the total number of 'ValueType' entries in the view. The
+ * computed reduction is added to 'result'.
+ */
+template <bool Serialize, typename TeamMember, typename InputProvider, typename ValueType>
+static KOKKOS_INLINE_FUNCTION
+void view_reduction (const TeamMember& team,
+                     const int total_size,
+                     const int vector_size,
```
ambrad, line 88: I think we can omit `vector_size` by doing something like `static consexpr int vector_size = typename decltype(InputProvider)::n` in the impl.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -64,10 +69,68 @@ void parallel_reduce (const TeamMember& team,
...
+/*
+ * Computes a reduction over the view described by 'input' for entries 'input([begin,end))'
+ * The variable 'input' should be an operator s.t. 'input(k)' returns a Pack of size 'vector_size'.
+ * The variable 'total_size' represents the total number of 'ValueType' entries in the view. The
+ * computed reduction is added to 'result'.
+ */
+template <bool Serialize, typename TeamMember, typename InputProvider, typename ValueType>
+static KOKKOS_INLINE_FUNCTION
+void view_reduction (const TeamMember& team,
+                     const int total_size,
```
ambrad, line 87: I think we can omit `total_size` because `typename decltype(InputProvider)::n`, `begin`, and `end` fully specify how to deal with the packs.

src/ekat/kokkos/ekat_kokkos_utils.hpp
ambrad: For clarity: `begin` and `end` should be _scalar_ indices, not pack indices. The physics param writer thinks in terms of model levels, not packs.

src/ekat/kokkos/ekat_kokkos_utils.hpp
bartgol: Uhm, InputProvider is not a Pack though. In most cases it'll be a view of packs, but it is also possible (and in Homme it happens) that it is a lambda (this allows to reduce the result of a formula without storing the temporary in a view).

src/ekat/kokkos/ekat_kokkos_utils.hpp
bartgol: Input provider is not a pack, but `input(0)` is, so `decltype(input(0))` will do.

src/ekat/kokkos/ekat_kokkos_utils.hpp
ambrad: Sorry, I should have written `typename decltype(input(0))::n`. That's probably still not quite right, as I'm just typing this in github and not actually test-compiling it. But that will work regardless of `input`'s type, as long as it returns a pack as required.

src/ekat/kokkos/ekat_kokkos_utils.hpp
ambrad: I commented on this below.

src/ekat/kokkos/ekat_kokkos_utils.hpp
ambrad: @bartgol my hasty C++-goof aside, don't you agree that neither total_size nor vector_size are needed?

bartgol: For sure vector size. On total_size: what if begin is not at the start of a pack? If begin is always at an aligned position, then yes, I think total_size is not needed. But I have to think about the general case.

bartgol: Although, if begin is not necessarily at the start of a pack, then we need special handling also for the 1st pack...

bartgol: A mid ground solution is to assume that the user "pads" his/her views, so that the use cases of view_reduction will always have begin at an aligned position (kind of like Hommexx did for ppm remap).

ambrad: That's the whole point of `begin` and `end`: to do special handling here instead of the app. We can assume the app will have aligned columns, i.e., `a(0)[0]` is aligned. Then we can conclude that `scalaraize(a)[[begin,end)]` is valid, from which we can work out the padding here. No `total_size` needed.

ambrad: Re: "some sort of check": I suggest we do that in some other PR. A bunch of stuff, like `scalarize` and `smallize`, make quite appropriate assumptions about alignment. Don't forget that a `Pack` _always_ aligns properly, so anything composed of packs will align too. That makes most or all of these assumptions safe. In principle someone could make a `Pack<7>` on a 512-bit-vector-length machine, but I think we shouldn't worry too much about that.

bartgol: You're right, we can work out the padding: if `begin%Pack::n!=0`, then we need padding. Simple check. If we want to be fancy, we could even check in debug mode that `scalarize(a).data()-begin` is at an aligned position (to make sure the user didn't mess up).

@tcclevenger This will require a bit more code to handle the 1st pack, but it should pay off in simplicity on the customer side.

tcclevenger: Ahh I see now. As I understood the implementation [here](https://github.com/E3SM-Project/E3SM/blob/b7b269b3cfaca869e861fa01e6d65e201dc588a3/components/homme/src/share/cxx/ColumnOps.hpp#L474), they run `parallel_reduce` over `LOOP_LENGTH=NUM_PACKS`, _but_ they do not allow for a custom range, and  instead always reduce the entire view.  If we do allow for a custom range, `begin` and `end` are scalar indices, not pack indices as I have it.  If that is the case, then I agree `begin`, `end`, and `typename decltype(input(0))::n` would be sufficient (assuming `input(0)[0]` is always aligned), but we'll see if I run into any trouble implementing. If `begin` and `end` are pack indices, then I'm not so sure.

bartgol: Yeah, in Hommexx I was free to impl only what hommexx needs. Ekat, however, needs to be as general as possible (within reasonable limits) to accommodate potential needs of E3SM apps.

I think Andrew is right, and we can always assume the view to be aligned. The only scenario where it would not be aligned is if someone allocates a view of plain Reals, then subviews it at a non-aligned index, and reinterpret it as a packed view. I believe we can assume that won't happen (just like C++ in general: the language will always let you free to shoot yourself in the foot).

src/ekat/kokkos/ekat_kokkos_utils.hpp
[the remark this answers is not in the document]
bartgol: If you hard code TTR, I wonder if we could remove LoopBoundaries from the template args, and make the type of loop_boundaries to be simply `Kokkos::Impl::TeamThreadRangeBoundariesStruct<int,TeamMember>`. That would catch an error if someone tries to use the routine with a TVR, as well as it will remove a non-necessary template argument.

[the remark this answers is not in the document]
bartgol: Note: if you don't want to use the `Impl` namespace from Kokkos (it's always best to avoid using impl/details namespaces), we could stuff this into `ekat_kokkos_types.hpp`:

```
template<typename TeamMember>
using TeamLoopBoundaries = decltype(Kokkos::TeamThreadRange(std::declval<TeamMember>(),0,1);
```
so that you can then do
```
template<template TeamMember,...>
foo(const TeamMember& team, const TeamLoopBoundaries<TeamMember>& loop_boundaries,...);
```

src/ekat/kokkos/ekat_kokkos_utils.hpp
bartgol: Ah, it looks like you need to remove the reference:

```
using PackType = typename std::remove_reference<decltype(input(0))>::type;
```

tcclevenger: Yep, just found that

src/ekat/kokkos/ekat_kokkos_utils.hpp
tcclevenger: Is the reduction always going to be over contiguous indices? Or do you foresee a situation where the user needs a reduction over, for example, the first entry in each pack? Or some arbitrary set of indices?

bartgol: I think it's very safe to assume reductions will always be over a contiguous set of entries.

ambrad: Let's keep it as [begin,end), contiguous.

ambrad: Just to confirm: begin, end should be scalar indices.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -64,10 +69,80 @@ void parallel_reduce (const TeamMember& team,
...
+                     const int begin,
+                     const int end,
+                     const InputProvider& input,
+                     ValueType& result)
+{
+  using PackType = typename std::remove_reference<decltype(input(0))>::type;
+  constexpr int vector_size = PackType::n;
+
+  // Perform a packed reduction over scalar indices
+  if (true) { // !Serialize) {
```
tcclevenger, line 95: Since I took the time to implement a serialized version of all functions called here (`parallel_reduce` and `reduce_sum`), if there are no objections I vote we do the packed reduction for both serial and parallel computations, instead of only for parallel (and a single loop over indices for serial, see commented out else-statement below). It makes the entire function simpler.

If this is ok, I'll remove the `if(true)` here and commented out else-statement below. If you don't agree, I'm happy to keep both routines with the `if(!Serialized)` check.

tests/kokkos/kokkos_utils_tests.cpp
```diff
@@ -244,4 +249,96 @@ TEST_CASE("parallel_reduce", "[kokkos_utils]")
...
+  if (Serialize) {
+    REQUIRE(results_h(0)== serial_result);
+  } else {
+    REQUIRE(std::abs(results_h(0) - serial_result) <= 10*std::numeric_limits<Scalar>::epsilon());
+  }
+}
+
+TEST_CASE("view_reduction", "[kokkos_utils]")
+{
+  // VectorSize = 1
```
tcclevenger, line 313: I think I have tested all the corner cases, but let me know if you feel I've missed one. If I always have `REQUIRE(results_h(0)== serial_result);`, no matter serial or parallel, many of the `Serialize=false` fail, so I am confident my algorithms are truly bfb to serial.

tcclevenger (commented):
@bartgol @ambrad @jeff-cohere 

I believe my last commit address all the discussion. Now, `begin` and `end` represent scalar indices. There is a routine to check for both incomplete begin and end packs. Overall, we calculate the equivalent of a reduction over `scalarize(input)([begin,end))`, but using packed sums.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -163,7 +249,18 @@ struct ExeSpaceUtils<Serialize,Kokkos::Cuda> {
...
                         ValueType& result)
   {
-    impl::parallel_reduce<TeamMember, Lambda, ValueType, Serialize>(team, begin, end, lambda, result);
+    impl::parallel_reduce<Serialize, TeamMember, Lambda, ValueType>(team, begin, end, lambda, result);
+  }
+
+  template <typename TeamMember, typename InputProvider, typename ValueType>
+  static KOKKOS_INLINE_FUNCTION
+  void view_reduction (const TeamMember& team,
+                       const int& begin,
```
ambrad, line 258: Since `begin` and `end` refer variously to pack and scalar, I recommend an inline comment in each place like this: `const int& begin, // scalar index` or `const int& begin, // pack index`.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -64,10 +69,80 @@ void parallel_reduce (const TeamMember& team,
...
+                     const int end,
+                     const InputProvider& input,
+                     ValueType& result)
+{
+  using PackType = typename std::remove_reference<decltype(input(0))>::type;
+  constexpr int vector_size = PackType::n;
+
+  // Perform a packed reduction over scalar indices
+  if (true) { // !Serialize) {
+    const bool has_garbage_begin = (begin%vector_size != 0 ? true : false);
```
ambrad, line 96: Style: `begin%vector_size != 0` is a boolean, so ternary with true/false as outputs is redundant. Just write `const bool has_garbage_begin = begin%vector_size != 0`.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -64,10 +69,80 @@ void parallel_reduce (const TeamMember& team,
...
+    const int pack_loop_begin    = (has_garbage_begin ? begin/vector_size + 1 : begin/vector_size);
+    const int pack_loop_end      = end/vector_size;
+
+    // If first pack has garbage, we will not include it in the below reduction,
+    // so manually add the first pack (only the non-garbage part)
+    if (has_garbage_begin) {
+      const int first_indx = begin%vector_size;
+      Kokkos::single(Kokkos::PerTeam(team),[&] {
+        for (int j=first_indx; j<vector_size; ++j) {
+          result += input(pack_loop_begin-1)[j];
```
ambrad, line 107: `input` is potentially an expensive lambda. You should evaluate it exactly once per pack. So eval outside the `j` loop, then index a temporary inside the `j` loop.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -64,10 +69,80 @@ void parallel_reduce (const TeamMember& team,
...
+      }, result);
+    }
+      
+    // If last pack has garbage, we did not include it in the main reduction,
+    // so manually add the last pack (only the non-garbage part)
+    if (has_garbage_end) {
+      const int last_indx = end%vector_size;
+      Kokkos::single(Kokkos::PerTeam(team),[&] {
+        for (int j=0; j<last_indx; ++j) {
+          result += input(pack_loop_end)[j];
```
ambrad, line 127: Same re: calling `input` just once per pack.

src/ekat/kokkos/ekat_kokkos_utils.hpp
ambrad: I think it would be best to move the `Serialized` check to just the middle of the three blocks. If not `Serialized` then reduce at the pack level followed by reduction of the pack, like Luca does in Hommexx; this is more efficient than reducing every pack. If `Serialized`, use what you have. This way, we get maximum efficiency but reduce the code repetition to just the easy middle packs.

tests/kokkos/kokkos_utils_tests.cpp
```diff
@@ -244,4 +249,96 @@ TEST_CASE("parallel_reduce", "[kokkos_utils]")
...
+  Kokkos::View<Scalar*> results ("results", 1);
+  const auto results_h = Kokkos::create_mirror_view(results);
+
+  // parallel_for over 1 team, i.e. call view_reduction once
+  const auto policy =
+    ekat::ExeSpaceUtils<Serialize,ExeSpace>::get_default_team_policy(1, view_length);
+  Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const MemberType& team) {
+    Scalar team_result = Scalar(a);
+
+    ekat::ExeSpaceUtils<Serialize,ExeSpace>::view_reduction(team, begin, end, data, team_result);
```
ambrad, line 297: You should test the lambda InputProvider case, as well. Can just wrap data in a lambda. In trial set 0, use the raw data view; in trial set 1, use the lambda wrapper. In general, in unit tests for ekat, _every_ possible input type must be tested.

src/ekat/kokkos/ekat_kokkos_utils.hpp
bartgol: I agree with Andrew. No need to reduce each pack if we are going for performance (i.e., Serialized=false). If not serialized, perform a packed reduction, then reduce the result pack at the end.

tcclevenger: For this I needed to add a definition for `reduction_identity<PackType>::sum()` to `ekat_kokkos_types.hpp`. Let me know if my implementation is acceptable.

src/ekat/kokkos/ekat_kokkos_types.hpp
```diff
@@ -71,4 +71,13 @@ struct KokkosTypes
 
 } // namespace ekat
 
+// Kokkos-compatible reduction identity for arbitrary packs
+namespace Kokkos {
+template<typename PackType>
+struct reduction_identity {
+  KOKKOS_FORCEINLINE_FUNCTION
+  static PackType sum()  {return PackType(typename PackType::scalar(0));}
```
ambrad, line 79: I think the simpler `Packtype(0)` should work on the RHS.

src/ekat/kokkos/ekat_kokkos_utils.hpp
ambrad: In the `impl` version, too, please, where in particular it's `pack index` rather than `scalar index`.

ambrad (approved):

src/ekat/kokkos/ekat_kokkos_utils.hpp
bartgol: I don't think you pushed that mod. I don't see it in the list of touched files...

tcclevenger:
The GPU tests are failing because of a compile error: 
`error: no default constructor exists for class "ekat::Pack<Real, 1>"`
When I then change the packs empty constructor to 
`KOKKOS_FORCEINLINE_FUNCTION`
`Pack () = default;`
it compiles and all tests (except one pack test requiring uninitialized packs have nan entries, since I removed the empty constructor). I found this solution reading from [this Kokkos issue](https://github.com/kokkos/kokkos/issues/1696)

src/ekat/kokkos/ekat_kokkos_types.hpp
```diff
@@ -71,4 +71,13 @@ struct KokkosTypes
 
 } // namespace ekat
 
+// Kokkos-compatible reduction identity for arbitrary packs
+namespace Kokkos {
+template<typename PackType>
+struct reduction_identity {
```
bartgol, line 77: I believe this should be a specialization, while this one is a declaration of a generic template. I'm not sure why it does not conflict with the one in `Kokkos_NumericTraits.hpp`.

bartgol:
I'm puzzled. The PR had not yet received an approving review, and yet the autotester merged it anyways. I'll ping the AT devs to see if they have an idea of what happened.

allevin:
@bartgol I need to see the autotester run logs from the runs around 9:51 9/15/20 and 8:32 9/16/20.  I don't have access to the server since your jobs moved.

bartgol:
So, for those interested, PR #42 was built on top of this, and since it got merged, this one was considered merged by github.

I added the [DEP](https://github.com/apps/dep) github app to this repo, and marked 42 as dependent on 37 inside that pr, but for some reason it did not work (maybe the DEP app is buggy, maybe I haven't configured it correctly, not sure).

This is somewhat underwhelming, since reviewing 42, Conrad probably meant to approve only the commits that came _after_ 37, but the AT doesn't know that, and consider the whole PR as approved.

I think the best way forward is to no longer have outstanding PRs that already include commits from another pending PR. That was poor choice on my end.

PR #41, opened 2020-09-16
Add PackInfo struct for pack index arithmetics.

### LINKED ISSUE #39: Add indexing utilities for packed arrays
When going back-forth from packed to unpacked quantities, one needs to continuously do index arithmetics. Besides being tedious, it is error prone.

Create a set of utilities, that perform this kind of index arithmetic for you. E.g.:
```
template<int PackSize>
PackInfo {
   // Which pack does array_idx map to?
   KOKKOS_INLINE_FUNCTION
   static int pack_idx (const int array_length, const int array_idx);
   // What entry *within the pack* does array_idx map to?
   KOKKOS_INLINE_FUNCTION
   static int vec_idx (const int array_length, const int array_idx);
   // How many packs are needed for the array
   KOKKOS_INLINE_FUNCTION
   static int num_packs (const int array_length);
   // Does the packed array contain padding?
   KOKKOS_INLINE_FUNCTION
   static bool has_padding (const int array_length);
   ..
};
```
and possibly others. One would then do
```
using Info = PackInfo<MY_PACK_SIZE>;
...
const int pack_id = Info::pack_idx(nlev,k);
const int vec_id   = Info::vec_idx(nlev,k);
if (Info::has_padding(nlev)) {
  // handle last pack carefully (namely, unpack it)
} else {
  // keep on rolling with packed operations
}
```

**Describe alternatives you've considered**
We could template on the actual pack type (which might be easier to use), but it would be somewhat wrong (and re-compile the functions for packs of same length but different scalar type). More interesting might be to offer, in addition to the above, a compile-time versions, where the array length is a compile time constant.

bartgol:
## Motivation

Computing pack/vector indices by hand is error prone, and can make the code heavier and/or harder to read. Utility functions can make the code more self-documenting, and most importantly safe.

## Related Issues

* Closes #39

## Testing

Tests added for all the features.

### DIFF
--- a/src/ekat/CMakeLists.txt
+++ b/src/ekat/CMakeLists.txt
@@ -1,13 +1,13 @@
 set(EKAT_SOURCES
-  mpi/ekat_comm.cpp
   ekat_assert.cpp
-  ekat_session.cpp
   ekat_parameter_list.cpp
   ekat_parse_yaml_file.cpp
+  ekat_session.cpp
+  mpi/ekat_comm.cpp
+  util/ekat_arch.cpp
   util/ekat_array_io.cpp
   util/ekat_array_io_mod.f90
   util/ekat_string_utils.cpp
-  util/ekat_arch.cpp
   util/ekat_test_utils.cpp
 )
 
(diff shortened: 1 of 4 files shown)


pbosler (approved):
Looks great, thanks @bartgol .  Tested & passed on Mac.

PR #42, opened 2020-09-16
Bartgol/default bfb

### LINKED ISSUE #35: Should EKAT have macros for defaults floating point behaviors?
This discussion came up during the review of #34 . The issue is the following. Some operations need to be done serially to achieve BFB. Obviously, we don't want a runtime switch, so we stick a template argument (like `bool Serialize`) in the template arguments. The question is now: should Ekat provide a default for `Serialize`, which depends on some config options? The code would look like this:

```
// With config-dependent defaults
#ifdef EKAT_BFB_DEFAULTS
template<typename T, bool Serialize = true>
#else
template<typename T, bool Serialize = false>
#endif
void foo (T& t) {
...
}

// Without config-dependent defaults (keep deducible template args last)
template<bool Serialize, typename T>
void bar (T& t) {
...
}           
```

Pros:
 - defaults make downstream code smaller; provided you configured EKAT with the correct flags, you can call `foo();` rather than `foo<true>();`;
 - people who read the code don't have to find out what the true/false tempate argument is for.

Cons:
 - if the library is installed, switching implementation (e.g., quickly checking lack of serialization yields a bug) requires reconfiguring rather than just recompiling (experienced cmake/make folks might simply hack a bunch of flags.make and link.txt files in the CMake directories, but it could be risky too, and/or be needed in a lot of cmake files).
 - if you have several template args, `Serialize` must be last (args with defaults must come after args with no defaults), so if you want/need to override the default at a specific location (perhaps for debugging) it might get complicated.

My original opinion was to not add defaults, and force customers to be explicit on what they want. If `foo` is used a lot, the app can define a wrapper that sets the `Serialize` argument depending on some _app-specific_ configuration option. However, I can see that there _is_ a value in providing the user with simpler/shorter code.

Thoughts?

bartgol:
## Motivation

It is cumbersome to always specify template args that switch between performant and bfb implementation. Having a default that can be set at config time is quite handy, and reduces code complexity.

## Related Issues

* Closes #35
* Depends on #37 

## Testing

Not sure how to add a test for this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,11 +69,12 @@ set (EKAT_TPL_INCLUDE_DIRS ${EKAT_TPL_INCLUDE_DIRS_INTERNAL} CACHE INTERNAL "Lis
 ############################
 
 option (EKAT_MPI_ERRORS_ARE_FATAL " Whether EKAT should crash when MPI errors happen." ON)
-
+option (EKAT_DEFAULT_BFB "Whether EKAT should default to BFB behavior whenever possible/appropriate." OFF)
 
 ####################################
 ###  MAC-SPECIFIC CONFIG OPTIONS ###
 ####################################
+
 include(CheckSymbolExists)
 # Most of the FPE stuff was defined in c++99, but gnu has some additional non-std
 # functions that are helpful (such as feeenableexcept) --- Apple stopped including these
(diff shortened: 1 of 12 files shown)


tcclevenger (approved):
Looks good to me.

PR #43, opened 2020-09-16
Some cmake upgrades/fixes

bartgol:
## Motivation

Some things were not working when I tried to use the current master inside scream. In particular on GPU. For instance, you need to set the nvcc wrapper correctly before calling `project`, but you don't know the kokkos src dir until you call project (if you are using the ekat submodule).

## Testing
SCREAM will test everything works when we update the ekat submodule in it.

### DIFF
--- a/cmake/Ekat.cmake
+++ b/cmake/Ekat.cmake
@@ -1,3 +1,6 @@
+# Where ekat's tpls live
+set (EKAT_CMAKE_PATH ${CMAKE_CURRENT_LIST_DIR} CACHE INTERNAL "")
+
 # This cmake utility allows you to build ekat just by adding
 #
 #   Include(Ekat)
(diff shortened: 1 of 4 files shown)


cmake/Ekat.cmake
```diff
@@ -1,3 +1,6 @@
+# Where ekat's tpls live
```
jeff-cohere, line 1: Is this true? It seems like it's where EKAT's CMake files live.

bartgol: Yeah, that's what I meant: ekat's cmake scripts for tpls. I'll fix it.

bartgol:
Argh, the AT automerged before I could push the fix. I'm going to push a second micro pr, with a bit more cmake clean up.

PR #46, opened 2020-09-17
Some more cmake upgrades

bartgol:
Besides reorganizing the cmake subfolder, this PR packages the ekat library into a cmake package, which can then be used with `find_package(Ekat)`.

I tested it locally on my laptop, and it seems to work, meaning that

```
project(foo CXX Fortran)
find_package(Ekat REQUIRED PATHS ${my_ekat_path})

add_executable(my.exe main.cpp)
target_link_libraries(my.exe ekat)
```
correctly configures, adds compiler/link flags, builds, and run. I tested a cpp utility (Pack) and a f90 one (array_io_file_exists). I also tested a kokkos view, to make sure it links correctly.

Notice that I do not have to specify where to find kokkos and yaml, since Ekat pkg configuration already contains the location of those subproject configuration (see EkatConfig.cmake.in). The foo project does not reference directly (in cmake or cpp code) kokkos or yaml.

Additionally, we now also install our cmake utilities, so that downstream projects can use them.


CMakeLists.txt
```diff
@@ -107,11 +101,44 @@ if (EKAT_ENABLE_TESTS)
...
+
+###########################################
+###   Install ekat cmake util scripts   ###
+###########################################
+
+install (FILES
+  ${EKAT_CMAKE_PATH}/EkatCreateUnitTest.cmake
+  ${EKAT_CMAKE_PATH}/EkatSetCompilerFlags.cmake
+  ${EKAT_CMAKE_PATH}/EkatUtils.cmake
+  DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/ekat/cmake)
```
bartgol, line 138: An alternative could be `<prefix>/share/cmake`, but it already contains yamlcpp cmake config stuff (albeit in a yaml-cpp subfolder). We could modify yamlcpp (we are using our own fork, afterall), to install its cmake config sutff in lib64/cmake, just like ekat and kokkos (would be simmetrycal).

jeff-cohere: The `share` path is the approach I've seen taken most often. I don't have strong feelings, though.

Regarding `yaml-cpp`--it's a much more general library than `ekat` or `kokkos`, so I don't mind it having a different set of conventions. If we ever want to merge our changes back into `yaml-cpp`, it might be in our interest to leave things like this alone.

jeff-cohere (dismissed):
These changes look good. I just responded to your comment about cmake paths. I'm approving, pending a fix to whatever's going on with Weaver.

CMakeLists.txt
bartgol: I also would like to use `share/cmake`. Maybe I'll stuff our cmake scripts in `share/cmake/Modules`, which mimics cmake installation folder.

bartgol:
The weaver error is me forgetting to change relative path in a cmake script. Fix coming up.

jeff-cohere (approved):
Looks great!

Issue #52, opened 2020-09-21
Odd behavior for reciprocal units.


pbosler:
**Describe the bug**
Odd behavior for reciprocal units, but `pow` is working correctly.

**To Reproduce**
Steps to reproduce the behavior:  in `ekat/tests/units/units.cpp`
1. Add `REQUIRE(one/mol == pow(mol,-1));`
2. Or (same result) `REQUIRE(1/mol == pow(mol,-1));`
3. run the test

> /Users/pabosle/ekat/tests/units/units.cpp:69: FAILED:
>   REQUIRE( one/mol == pow(mol,-1) )
> with expansion:
>   m s kg K A cd == mol^-1

bartgol:
Ugh, let me check what's going on here.

bartgol:
Uh, wow, look at what I wrote!

```
inline Units operator/(const ScalingFactor& lhs, const Units& rhs) {
  return Units(1-rhs.m_units[0],
               1-rhs.m_units[1],
               1-rhs.m_units[2],
               1-rhs.m_units[3],
               1-rhs.m_units[4],
               1-rhs.m_units[5],
               1-rhs.m_units[6],
               lhs/rhs.m_scaling);
```

So `1/(a^n) = a^(1-n)`. Math PhD fail.

bartgol:
Also, this means that we didn't have a test for `op/`, and I'm guessing for many other operators. I will fix this.

Issue #48, opened 2020-09-17
Physical constants


pbosler:
## Question

What would be the best (or at least acceptable to a majority) way to handle physical constants in a consistent manner throughout scream?   

This issue has come up in several recent discussions (e.g., @PeterCaldwell, @AaronDonahue ) and at least 2 issues:
[haero issue #6](https://github.com/E3SM-Project/haero/issues/6) ( @jeff-cohere and @pbosler )
[scream pr #579](https://github.com/E3SM-Project/scream/pull/579) (@tcclevenger, @ambrad)

### For discussion:
1. Currently, different parameterizations often define the same constants, but with different numbers of significant figures and/or units.  The ideal solution is to use NIST-defined values for each constant, defined in a single location (file or struct) that dynamics and parameterization can use/alias with local variables as needed.  
2. Suppose people agree with that statement;  where should such code live?
Physical constants have nothing to do with Kokkos, but EKAT already contains other supporting features like `ekat::units` that are also unrelated to performance portability, but generally useful to everything in E3SM, so I'm inclined to put physical constants in EKAT (@bartgol).
3. We've found that some constants require double precision (e.g., they may be very small or very large, or they may be required by possibly ill-conditioned algorithms) whereas others can be single precision.  How should these requirements be handled in scream & ekat?
4. What should the timeline be for this work?
5. Should this be a forward looking c++ only change (to avoid BFB issues), or should it include fortran as well?

ambrad:
I'll take the easy one, 3. Just define it as hard-coded double rather than Real. Do that consistently in F90 and C++ when needing BFB tests. If implicit type conversion in the presence of multiple types is not the same in C++ and F90 by default, I believe you can add an F90 flag or flags to get it to be the same.

pbosler:
Oof.  BFB.  Added item 5 to the discussion.

PeterCaldwell:
I agree this stuff should live in EKAT. I agree with Andrew about making params double by default. I don't think we can be "forward thinking" by just doing this in C++ because of our BFB tests. I'd suggest just copying whatever E3SM uses for our initial version of this constants database (rather than matching NIST) in order to circumvent the BFB issue while still targeting C++ only. Then when we jettison F90, we can update our constants file with NIST values. 

One problem with what I just said is that E3SM may use different values for the same constant in different places in the code. And collecting E3SM values may be a hassle.

pbosler:
> E3SM may use different values for the same constant in different places in the code. 
[...]

PR #53, opened 2020-09-21
Prototype for PhysicalConstant class [WIP]

pbosler:
## Motivation
To further the discussion about Physical Constants, here is a possible impl, based in large part on the `ekat::units` design.  The purpose of this WIP PR is to collect comments/criticism on the class design.

The basic data structure has three member variables, a `RealType` value (template parameter default is `double`), an `ekat::units::Units` object,  and a `std::string` name.  An additional field for extra info, e.g., assumptions related to the constant such as "at constant temperature" can be recorded.

There is an additional member variable to record two possibly different values (one used by e3sm, one defined by NIST).  The default `operator()` returns the e3sm value.  

Constructors allow for new constants to be defined explicitly or as the product of 2 previously defined constants.

The class is written for all-scream constants and allows for inheritance so that, for example, a subcomponent e.g., Haero or Shoc, could define module-specific constants as `class ShocConstant : public PhysicalConstant`, for example.   A few examples of some specific constants are defined as namespace variables in `ekat_physical_constants.hpp`, based on the convention used for the common SI units in `ekat_units.hpp`.  

A unit test is included, which demonstrates some basis behaviors.

## Related Issues
First step toward  #48.
Includes a test for #52 (currently fails at runtime).

### DIFF
--- a/src/ekat/ekat_physical_constants.hpp
+++ b/src/ekat/ekat_physical_constants.hpp
@@ -1,28 +1,29 @@
 #ifndef EKAT_PHYSICAL_CONSTANTS
+
 #define EKAT_PHYSICAL_CONSTANTS
 
 #include "ekat/util/ekat_units.hpp"
 
 namespace ekat {
 
-template <typename RealType=double>
 class PhysicalConstant {
   public:
+    using RealType = double;
     //enum Op {Times, Divide};
 
     // no default
     PhysicalConstant() = delete;
 
-    PhysicalConstant(const PhysicalConstant<RealType>&) = default;
+    PhysicalConstant(const PhysicalConstant&) = default;
 
     // Constructor for single value (i.e., NIST std)
     PhysicalConstant(const RealType& nvalue, const ekat::units::Units& units, const std::string& name) :
-      m_nist_val(nvalue), m_e3sm_val(nvalue), m_units(units), m_name(name), m_info("") {}
+      m_e3sm_val(nvalue), m_nist_val(nvalue), m_units(units), m_name(name), m_info("") {}
 
     // Constructor for single value (i.e., NIST std)
     PhysicalConstant(const RealType& nvalue, const ekat::units::Units& units, const std::string& name,
       const std::string& info) :
-      m_nist_val(nvalue), m_e3sm_val(nvalue), m_units(units), m_name(name), m_info(info) {}
+      m_e3sm_val(nvalue), m_nist_val(nvalue), m_units(units), m_name(name), m_info(info) {}
 
 
     // Constructor for two different values (i.e., when E3SM value and NIST value are different)
@@ -35,12 +36,6 @@ class PhysicalConstant {
       const std::string& name, const std::string& info) : m_e3sm_val(evalue), m_nist_val(nvalue), m_units(units),
       m_name(name), m_info(info) {}
 
-    // Constructor from the product of two other PhysicalConstants
-    PhysicalConstant(const PhysicalConstant& a, const PhysicalConstant& b, const std::string& name,
-      const std::string& info="") : m_nist_val(a.m_nist_val*b.m_nist_val),
-        m_e3sm_val(a.m_e3sm_val*b.m_e3sm_val), m_units(a.m_units*b.m_units),
-        m_name(name), m_info(info) {}
-
     std::string name() const {return m_name;}
 
     std::string get_string() const {return m_name + " [" + to_string(m_units) +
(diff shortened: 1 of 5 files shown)


src/ekat/ekat_physical_constants.hpp
```diff
@@ -0,0 +1,92 @@
...
+    std::string name() const {return m_name;}
+
+    std::string get_string() const {return m_name + " [" + to_string(m_units) +
+      (m_info.empty() ? "]" : "] (" + m_info + ")");}
+
+    std::string unit_string() const {return to_string(m_units);}
+
+    RealType value() const {return m_e3sm_val;}
+
+    RealType operator() () const {return value();}
```
jeff-cohere, line 53: I'm curious what others think about using an implicit casting operator for `RealType` here, which would eliminate a set of parenthesis in our math expressions. Given that constants need to be converted to packs (using `sp(constant)`, for example, in SCREAM), maybe that would look a little cleaner. (I'm not usually an advocate for implicit casting, mind you.)

And maybe it would be a good idea to take some example expressions from P3 and SHOC in SCREAM and see how this class might be used to simplify them, or at least see how it looks when used in that context.

src/ekat/ekat_physical_constants.hpp
```diff
@@ -0,0 +1,105 @@
...
+      m_name(name), m_info("") {}
+
+    // Constructor for two different values (i.e., when E3SM value and NIST value are different)
+    PhysicalConstant(const RealType& evalue, const RealType& nvalue, const ekat::units::Units& units,
+      const std::string& name, const std::string& info) : m_e3sm_val(evalue), m_nist_val(nvalue), m_units(units),
+      m_name(name), m_info(info) {}
+
+
+    // Constructor for a constant that is defined by the product or quotient of 2 other constants.
+    PhysicalConstant(const PhysicalConstant<RealType>& lhs, const Op& op, const PhysicalConstant<RealType>& rhs,
```
bartgol: I think it would be easier if you overloaded op* and op/. Then one would simply do

```
auto gamma (cp/cv, "specific heat ratio");
```
which mean you need a copy ctor that takes another PhysicalConstant, with a different name. The overload of op* and op/ could simply use "name1 * name2" and "name1 / name2" for the name, and "info1; info2" for the info section.

src/ekat/ekat_physical_constants.hpp
```diff
@@ -0,0 +1,105 @@
...
+          }
+        }
+    }
+
+    std::string name() const {return m_name;}
+
+    std::string get_string() const {return m_name + " " + to_string(m_units) +
+      (m_info.empty() ? "" : " (" + m_info + ")");}
+
+    std::string unit_string() const {return to_string(m_units);}
```
bartgol, line 49: Why not simply add `get_units()`, and let the user decide what to do with it? If they want the units stirng, they can always call `to_string(m_units)` (or other ways of displaying units). But returning units also allows to create other quantities with compatible units.

pbosler: I meant to do that 👍

src/ekat/ekat_physical_constants.hpp
```diff
@@ -0,0 +1,105 @@
...
+         lhs.m_nist_val == rhs.m_nist_val &&
+         lhs.m_units == rhs.m_units;
+}
+
+template <typename RealType>
+inline bool operator != (const PhysicalConstant<RealType>& lhs, const PhysicalConstant<RealType>& rhs) {
+  return !(lhs == rhs);
+}
+
+// ================== COMMON PHYSICAL CONSTANTS =============== //
```
bartgol: I would put all these constants in a separate namespace, something like `physical_constants`. The way it is now, they pollute the ekat namespace.

Also, I would put them in a separate file, with "double" in the filename, so people know these constants have a pre-defined precision.

src/ekat/ekat_physical_constants.hpp
```diff
@@ -0,0 +1,105 @@
...
+      m_nist_val(nvalue), m_e3sm_val(nvalue), m_units(units), m_name(name), m_info("") {}
+
+    // Constructor for single value (i.e., NIST std)
+    PhysicalConstant(const RealType& nvalue, const ekat::units::Units& units, const std::string& name,
+      const std::string& info) :
+      m_nist_val(nvalue), m_e3sm_val(nvalue), m_units(units), m_name(name), m_info(info) {}
+
+
+    // Constructor for two different values (i.e., when E3SM value and NIST value are different)
+    PhysicalConstant(const RealType& evalue, const RealType& nvalue, const ekat::units::Units& units,
```
bartgol, line 29: Note (for here and the rest of the file): you don't need to prepend `ekat::`, since you are already in the ekat namespace. That might help keeping lines shorter, which is helpful for the reader.

src/ekat/ekat_physical_constants.hpp
pbosler: Agree completely.  I put them here just for the sake of this PR, to see if people are ok with them being namespace variables (in their own namespace) or if people want them wrapped up in a struct.

bartgol: I see +/- of either approach:

- namespace: you can "add" to a namespace in a separate file, but then the namespace (and file) need to make clear what precision is used.
- struct: whether sp or dp is clear from the struct template arg, but you cannot extend the struct.

Which one to pick, is a matter of taste, I think. Personally, I'd probably go with a struct, and have apps define their own namespace. But I don't feel really strong about it.

One thing I'd do, though, is to explicitly type `<double>` in the lines below, just in case someone changes the class default (impossible for downstream app to detect) or remove the default altogether (more likely, although would immediately generate compiler errors).

src/ekat/ekat_physical_constants.hpp
pbosler: Thanks.  The only reason I did it this way was because of the `name` member variable.   I agree it's ugly.  

I'm happy to make this change, but I wonder for things like the universal gas constant, R = Avogadro*Boltzmann, is that name ok?  And if not, how do you change the name while also respecting `const`?

src/ekat/ekat_physical_constants.hpp
pbosler: Agree.  Once I get a basic design ok'd, finding a few examples in p3 and shoc is where I'll go next.  I won't put a final PR in until we are ok with both the design and how it's used.

src/ekat/ekat_physical_constants.hpp
pbosler: Added a separate file and namespace for common constants; it's what `ekat_units.hpp` does, and it's closer to the fortran domain scientists are familiar with already.  No strong feelings though --- happy to encapsulate in a templated struct if others prefer it.   On that note...

Changed all constants to double precision.  Is there harm in this choice? (params can still compute in single)

src/ekat/ekat_physical_constants.hpp
pbosler: Resolved this with a few thoughts:
1. units are all boiled-down to their fundamentals (i.e., Joules becomes kg m^2 / s^2) so it's consistent if constants record their fundamental origins too.
2. the in-source name can be different e.g., R_gas_universal, than the member variable name (again, similar to units like Newtons).
3. The overloaded operators are more intuitive and therefore less error prone (@jeff-cohere POLA).

src/ekat/ekat_physical_constants.hpp
bartgol: > Changed all constants to double precision. Is there harm in this choice? (params can still compute in single)

Not sure what you mean. You mean having 14 digits when hard-coding stuff like this?
```
const auto Pi = PhysicalConstant<>(3.14159265358979323846264, ekat::units::Units::nondimensional(), "Pi");
```
That's fine.

pbosler: No I meant, is there harm in `using RealType = double;` (with regard to single-precision builds)

src/ekat/ekat_physical_constants.hpp
bartgol: I think implicit cast is ok here. One is already specifying the underlying precision via the RealType, so there's not intrinsic risk of unexpected precision loss/gain.

src/ekat/ekat_physical_constants.hpp
```diff
@@ -0,0 +1,114 @@
...
+    std::string m_name;
+    std::string m_info;
+
+    friend bool operator == (const PhysicalConstant&, const PhysicalConstant&);
+    friend PhysicalConstant operator*(const PhysicalConstant&, const PhysicalConstant&);
+    friend PhysicalConstant operator/(const PhysicalConstant&, const PhysicalConstant&);
+    friend std::string combine_info(const PhysicalConstant&, const PhysicalConstant&);
+};
+
+inline std::string combine_info(const PhysicalConstant& a, const PhysicalConstant& b) {
```
bartgol, line 66: Would having a default info like `[EMPTY]` (or something along that line) be acceptable? If so, you could do away with this function, and simply use `lhs.info + "; " + rhs.info` in your op* and op/ impl.

bartgol: I might vote against my own suggestion here.

src/ekat/ekat_physical_constants.hpp
bartgol: I don't think so, at least not necessarily. But the user need to be aware of that, or else hard-to-track non bfb issues might emerge. Take this:

```
float my_mass = 1.0/3.0; // or some other non-exact sp floating point
auto force = PhysicalConstants<>::g_accel * my_mass;
```
The variable `force` will be a double, due to C++ promotion rules. And even forcing the result to be a float might not help:
```
float my_mass = 1.0/3.0; // or some other non-exact sp floating point
float my_length = 1.0/3.0; // or some other non-exact sp floating point
float work1 = (PhysicalConstants<>::g_accel * my_mass) * my_length;
float work2 = (PhysicalConstants<float>::g_accel * my_mass) * my_length;
```
The intermediate result in parentheses for work has double precision, so my_length will be temporarily promoted to double for the second product. The result will likely not be bfb with work2.

Again, that's fine, as long as users are making responsible use of the default. E.g., in scream, if they wrap the constant inside a `sp(...)` call, then all is good.

bartgol: Actually, in my 1st example, even with `float` instead of `auto` you might get quite a different output. Without carries, you might expect 0-1 ulp error. With carries, anything is on the table (say your constant is 1+epsilon<float>/2, and you add epsilon<float>/2. If the constant is first converted to float, you get 1, otherwise you get 1+epsilon<float>.

pbosler:
Closing for now.  Will reopen when I have more time to work on this.

PR #55, opened 2020-09-23
Using NEW behavior for cmp0074

bartgol:
## Motivation

CMake 3.12 and above defaults to OLD, but issues a warning. OLD behavior ignores cmake/env `<PackageName>_ROOT` vars already defined when calling `find_package(<PackageName>)`, while the NEW behavior uses those variables.

Since we are ok with the NEW behavior, and OLD behaviors are deprecated by default (and may disappear in the future), better switch explicitly to NEW.


jgfouca (approved):
Thx!

PR #56, opened 2020-09-29
fix view_reduction

tcclevenger:
## Motivation
Solves issue from https://github.com/E3SM-Project/scream/pull/615, namely, when running with OPENMP threads, the `has_garbage_begin/end` if-blocks of `impl::view_reduction()` were not summing over all view values. 

Easy fix: needs to run over `Kokkos:single(PerThread)` instead of `Kokkos::single(PerTeam)`.

## Testing
The original view_reduction test in EKAT never used threads, and so no error. I updated the test to run using a team policy with threads when OPENMP is used. I verified that the new test fails with the old PerTeam implementation, and passes with the PerThread impl.

### DIFF
--- a/tests/kokkos/kokkos_utils_tests.cpp
+++ b/tests/kokkos/kokkos_utils_tests.cpp
@@ -250,7 +250,7 @@ TEST_CASE("parallel_reduce", "[kokkos_utils]")
 }
 
 
-template<typename Scalar, bool Serialize, bool UseLambda, int TotalSize, int VectorSize, bool UseThreads>
+template<typename Scalar, bool Serialize, bool UseLambda, int TotalSize, int VectorSize>
 void test_view_reduction(const Scalar a=Scalar(0.0), const int begin=0, const int end=TotalSize)
 {
   using Device = ekat::DefaultDevice;
@@ -288,14 +288,15 @@ void test_view_reduction(const Scalar a=Scalar(0.0), const int begin=0, const in
   Kokkos::View<Scalar*> results ("results", 1);
   const auto results_h = Kokkos::create_mirror_view(results);
 
-#ifdef KOKKOS_ENABLE_OPENMP
-  const int n_threads = omp_get_max_threads();
-#else
-  const int n_threads = 1;
-#endif
+  // Set team size
+  int team_size = ExeSpace::concurrency();
+  if (ekat::OnGpu<ExeSpace>::value) {
+    team_size /= (ekat::is_single_precision<Real>::value ? 64 : 32);
+  }
+
   // parallel_for over 1 team, i.e. call view_reduction once
   const auto policy =
-    ekat::ExeSpaceUtils<ExeSpace>::get_team_policy_force_team_size(1, (UseThreads ? n_threads : 1));
+    ekat::ExeSpaceUtils<ExeSpace>::get_team_policy_force_team_size(1, team_size);
   Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const MemberType& team) {
     Scalar team_result = Scalar(a);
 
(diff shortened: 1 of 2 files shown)


ambrad (dismissed):

tests/kokkos/kokkos_utils_tests.cpp
```diff
@@ -288,9 +288,14 @@ void test_view_reduction(const Scalar a=Scalar(0.0), const int begin=0, const in
...
 
+#ifdef KOKKOS_ENABLE_OPENMP
+  const int n_threads = omp_get_max_threads();
+#else
+  const int n_threads = 1;
+#endif
   // parallel_for over 1 team, i.e. call view_reduction once
   const auto policy =
-    ekat::ExeSpaceUtils<ExeSpace>::get_default_team_policy(1, view_length);
+    ekat::ExeSpaceUtils<ExeSpace>::get_team_policy_force_team_size(1, (UseThreads ? n_threads : 1));
```
bartgol, line 298: I don't understand this. On GPU, this only tests a trivial team/vector size (team_size=1, vector_length=1). In that case, the serialized and non-serialized reductions should do the same job, so the test might not see bugs.

I think the policy should have team size larger than 1, and, on Cuda, a vector length larger than 1.

tests/kokkos/kokkos_utils_tests.cpp
```diff
@@ -288,9 +288,14 @@ void test_view_reduction(const Scalar a=Scalar(0.0), const int begin=0, const in
   Kokkos::View<Scalar*> results ("results", 1);
   const auto results_h = Kokkos::create_mirror_view(results);
 
+#ifdef KOKKOS_ENABLE_OPENMP
+  const int n_threads = omp_get_max_threads();
+#else
+  const int n_threads = 1;
```
bartgol, line 294: This always forces a team size of 1 on Cuda. You could ask for

```
team_size = ExeSpaceUtils<ExecSpace>::concurrency();
if (OnGpu<ExecSpace>::value) {
  team_size /= (is_single_precision<Real>::value ? 64 : 32);
}
```
That would give you all the available threads in OpenMP builds, and the number of warps in the block on Cuda.

tcclevenger: Can I simply just call `get_team_policy_force_team_size(1, VectorSize);`? Or do I need example with `packsize!=teamsize`?

The issue with the above is that `ExeSpace::concurrency() =  163,840` and I would get an error that team size should be `< 1024`. Another option is I could just set `team_size /= 161` on GPU.

tests/kokkos/kokkos_utils_tests.cpp
```diff
@@ -288,9 +288,15 @@ void test_view_reduction(const Scalar a=Scalar(0.0), const int begin=0, const in
...
+  // Set team size
+  int team_size = ExeSpace::concurrency();
+  if (ekat::OnGpu<ExeSpace>::value) {
+    team_size /= (ekat::is_single_precision<Real>::value ? 64 : 32);
+  }
+
   // parallel_for over 1 team, i.e. call view_reduction once
   const auto policy =
-    ekat::ExeSpaceUtils<ExeSpace>::get_default_team_policy(1, view_length);
+    ekat::ExeSpaceUtils<ExeSpace>::get_team_policy_force_team_size(1, team_size);
```
bartgol, line 299: It might not matter here whether the vector length is 1 or larger. But I _think_ we should give the possibility in `get_team_policy` to specify the vector length (ignored on non-GPU builds).

bartgol (dismissed):
I had a comment on vector length, but it was mostly me thinking out loud. I _think_ this PR would exert all testing that is needed.

bartgol:
@tcclevenger my bad. `Cuda::concurrency()` counts the numberr of threads across the whole GPU.

bartgol:
It looks like you can get the SM count with

```
auto num_sm = Kokkos::Cuda::impl_internal_space_instance()->m_multiProcCount;
```
So you should be fine by dividing team_size by `num_sm*64` or `num_sm*32` (depending on precision).

tcclevenger:
@bartgol Let me know what you think of my changes.

bartgol (approved):

Issue #58, opened 2020-10-04
SCREAM compilation fails with CMAKE version 3.11


singhbalwinder:
**Describe the bug**
cmake_policy CMP0074 is only available with CMAKE 3.12.0 version or higher. It might fail for machines such as compy which has a cmake version of 3.11.

**To Reproduce**
Compile Scream on Compy machine (or any any machine with older CMAKE version) 

**Expected behavior**
Model should compile with versions older than CMAKE 3.12.0
The following code if added to `CMakeLists.txt` fixes this issue:
```
IF(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.12.0")
    cmake_policy(SET CMP0074 NEW)
ENDIF(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.12.0")
```

bartgol:
@singhbalwinder could you please put your fix in a quick PR?

singhbalwinder:
Sure. I will issue a PR soon.

PR #59, opened 2020-10-05
Adds a fix so that CMAKE works for CMAKE versions older than 3.12.0

singhbalwinder:
cmake_policy CMP0074 is only available in versions 3.12.0 or newer. Some machines (such as Compy) still has older version of CMAKE where the code fails to compile. This fix wraps application of this policy in an "if" condition so that it is applied only for CMAKE versions newer than 3.12.0.  

I have tested it on Compy and the code compiles on Compy with this change.
Fixes: #58


bartgol (approved):
Thanks for the quick fix!

PR #60, opened 2020-10-07
Set default value of EKAT_DEFAULT_BFB to ON/OFF depending on CMAKE_BUILD_TYPE

bartgol:
## Motivation
Users may forget to enable the EKAT_DEFAULT_BFB option. To make life easy, we set the default of this option to ON in debug builds, and OFF in non debug builds.


singhbalwinder (approved):
Looks good! Thanks for this quick PR!

PR #61, opened 2020-10-08
Removed FindAVX logic, let kokkos handle avx flags

bartgol:
## Motivation

FindAVX was a way to automatically detect avx flags from the host. However, that would not work in case of cross compiling. The most reliable way to do this is simply to set the correct Kokkos arch, and let kokkos add the proper vectorization flags. This way we also reduce the number of files we need to maintain.


jeff-cohere (approved):
So we're just getting rid of logic and explicitly relying on Kokkos to treat the specified architecture properly? Sounds good to me.

bartgol:
Yes. `Kokkos_ARCH_XYZ` already handles the vectorization flags, and it does it in a better way that we could (more archs, more compilers, better flags).

PR #63, opened 2020-10-12
More granular handling of unit tests compiler defs/flags

bartgol:
## Motivation

Currently, `EkatCreateUnitTest` only allowed to specify flags/definitions for all languages. While compilers may only issue a warning if, say, a F90 flag is used with the C++ compiler, it is still better to give the user the ability to specify separate flags/defs for each language (on top of common defs/flags). This may be even more important if different compilers vendors are used for different languages (e.g., when using nvcc for cxx).

## Testing

I manually tested a few combinations of the individual/common flags/defs, and manually verified in the `flags.make` file that they were correctly assigned.


jeff-cohere (approved):
This looks a lot cleaner and a lot more flexible. Nice.

bartgol:
The blake failure looked like a fluke. Rather than waiting for the AT to go through all scream PRs and come back to this, I manually fired the blake job again, and it passed. I'll merge manually.

PR #64, opened 2020-10-12
Two hot fixes to EkatCreateUnitTest

bartgol:
## Motivation

1) there was a typo in CXX inside a var name; it was spelled CxX.
2) when linking libs, link ekat_test_main *before* any libs passed to the
   macro, since the user might as for EXCLUDE_TEST_SESSION, and those extra
   libs might contain the implementation of the test session functions.
   By linking user provided libs *after* ekat_test_main, dependencies are
   correctly resolved.

## Testing

I configured/compiled SCREAM without/with this mod, and this PR fixes config/linking errors I was getting.


jeff-cohere (approved):
Darned case sensitivity!

Looks like the mappy failure comes from using an ancient (2.x) version of cmake there.

bartgol:
Yeah, I realized I did a sloppy job setting up the mappy jenkins stuff. I edited the bowman job, but forgot to update a bunch of stuff, like the modules... :)

bartgol:
Manually retested on mappy, after fixing the jenkins shell script. Tests pass. Merging manually so we don't have to wait till the AT picks it up again.

PR #66, opened 2020-10-16
Add 3d version of transpose and add tests

jgfouca:
This will enable us to support 3d arrays in scream physics f90 <-> cxx bridges.


jgfouca:
I'll need en EKAT update in SCREAM as soon as this is merged.

bartgol (approved):

PR #70, opened 2020-10-21
Fix compilation warnings regarding std::fn on device

### LINKED ISSUE #8: Fix warning regarding calling host function from device in ekat_pack.hpp
The error looks like this:

```
ekat_pack.hpp(295): warning: calling a __host__ function from a __host__ __device__ function is not allowed
          detected during:
            instantiation of "ekat::pack::OnlyPack<PackType> ekat::pack::sqrt(const PackType &) [with PackType=ekat::pack::OnlyPack<ekat::pack::Pack<long, 1>>]" 
```

The warning seems to be innocuous, but it would be nice to put in place some logic (probably involving CPP) that prevents this warning from showing up.

bartgol:
## Motivation

On CUDA, the compilation of pack math functions caused a bunch of warnings, regarding std::blah functions not being a device fcn but being called inside a `__host__ __device__` function. I am not 100% sure why this warning generated, and yet was not a real problem.

This PR fixes the warning, by providing 3 implementation of all the math functions, for double, float, and integer types.

## Related Issues

* Closes #8 

## Testing

These modification affect only pack operations implementations, which continue to be tested as they were before.


bartgol:
@jgfouca @ambrad I pinged you two cause you implemented most of the pack stuff. Feel free to take your name off if you'd rather pass.

ambrad:
This moves some code and then for testing only replaces a macro with a function? Or is there more?

bartgol:
> This moves some code and then for testing only replaces a macro with a function? Or is there more?

I moved the pack math overloads to their own file, cause ekat_pack.hpp was a bit long. For testing, nothing has changed, it's the same as before. What changed is how the different math fcns overloads are generated: the macro instantiating the pack overloads now calls `MathFcn<ScalarT>::fcn`, which takes care of doing a different impl depending on a) ScalarT, and b) whether we're inside a CUDA kernel.

ambrad:
Sorry for being dense, but is this introducing one extra level of function call for things like pow?

bartgol:
@ambrad yes, although it should be inlined (I may replace `KOKKOS_INLINE_FUNCTION` with `KOKKOS_FORCEINLINE_FUNCTION` if you prefer).

jgfouca (approved):

ambrad:
Ok. Yes, please use KOKKOS_FORCEINLINE_FUNCTION. We have clear evidence from studies ~2 years ago that if one isn't very careful, `inline` isn't enough and non-inlined function calls can destroy vectorization.

Also, I don't see why an extra fn call is necessary. Why not just use macros whose def depends on __CUDA_ARCH__? This is super low-level code, so worrying about things like fn calls destroying vectorization is important.

bartgol:
To give more context: things were also (almost) working with something like this

```

#ifdef __CUDA_ARCH__
#define ekat_pack_gen_unary_stdfn(fn)               \
  template <typename ScalarT, int N>                \
  KOKKOS_INLINE_FUNCTION                            \
  Pack<ScalarT,N> fn (const Pack<ScalarT,N>& p) {   \
    Pack<ScalarT,N> s;                              \
    vector_simd                                     \
    for (int i = 0; i < N; ++i) {                   \
      s[i] = fn(p[i]);                              \
    }                                               \
    return s;                                       \
  }
#else
#define ekat_pack_gen_unary_stdfn(fn)               \
  template <typename ScalarT, int N>                \
  KOKKOS_INLINE_FUNCTION                            \
  Pack<ScalarT,N> fn (const Pack<ScalarT,N>& p) {   \
    Pack<ScalarT,N> s;                              \
    vector_simd                                     \
    for (int i = 0; i < N; ++i) {                   \
      s[i] = std::fn(p[i]);                         \
    }                                               \
    return s;                                       \
  }
#endif
```
since cuda overloads `blah` for both double and float. However, the compilation would crap out when trying to test on int/long.

jgfouca:
Just a semi-related question, if KOKKOS_FORCEINLINE_FUNCTION is more reliable than KOKKOS_INLINE_FUNCTION, should we just always use KOKKOS_FORCEINLINE? If not, when to prefer one over the other?

bartgol:
> Just a semi-related question, if KOKKOS_FORCEINLINE_FUNCTION is more reliable than KOKKOS_INLINE_FUNCTION, should we just always use KOKKOS_FORCEINLINE? If not, when to prefer one over the other?

I think the point is that you don't want to inline the whole library. For key functions, you want to be _sure_ it is inlined, e.g., for vectorization reasons, as Andrew pointed out. But in other less crucial cases, it's probably best to let the compiler decide.

ambrad:
@jgfouca, I agree strongly with Luca. FORCEINLINE must be used very carefully. It's pretty much meant only for dealing with things like this.

bartgol:
Ok, I removed testing of math functions for int/long. Math fcns calls on packs of integers will result in a compiler error on cuda (except for abs, which is overloaded for int/long). This also removed the need for MathFcn, and the discussion on FORCEINLINE.

Note: I _did_ however replace INLINE with FORCEINLINE on the max/min functions in ekat_math_utils.hpp, since it is possible they might get called inside vectorizable loops.

ambrad (approved):

jgfouca (approved):

PR #71, opened 2020-11-05
When checking for fenv.h, use the cxx compiler.

bartgol:
## Motivation

Strictly speaking, the C++ compiler is not needed. What is needed is __GNU_SOURCE to be defined. Now,
the g++ compiler *does* define that macro, and since we compile ekat as c++ sources, we might as well check if the *cxx* compiler picks up the feenableexcept symbol.

## Additional Information

Note: I have no idea why this never got picked up. Using `check_symbol_exists` should _always_ use the C compiler, which should not have feenableexcept defined. However, looking at the output of our autotester past builds, I see that mpicxx compiler was used, I'm not sure why. Still, when building scream within cime just now, I had mpicc as compiler, which caused the check to fail (with consequent compilation of `ekat_feutils.hpp`, which contains MacOS specific stuff.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,11 +77,11 @@ option (EKAT_DEFAULT_BFB "Whether EKAT should default to BFB behavior whenever p
 ###  COMPILER/OS-SPECIFIC CONFIG OPTIONS ###
 ############################################
 
-include(CheckSymbolExists)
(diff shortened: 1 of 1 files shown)


jeff-cohere (approved):
This builds and passes tests on my Mac.

PR #73, opened 2020-11-16
Add 3d device-host functions

tcclevenger:
Add `host_to_device` and `device_to_host` functions for 3d data. Needed for the SHOC conversion in SCREAM. I include a "host_device_packs_3d" test, similar to the 2d version right above it.

### DIFF
--- a/src/ekat/ekat_pack_kokkos.hpp
+++ b/src/ekat/ekat_pack_kokkos.hpp
@@ -312,6 +312,52 @@ void host_to_device(const Kokkos::Array<typename ViewT::value_type::scalar const
   }
 }
 
+// 3d - set do_transpose to true if host data is coming from fortran
+template <typename SizeT, size_t N, typename ViewT>
+void host_to_device(const Kokkos::Array<typename ViewT::value_type::scalar const*, N>& data,
+                    const Kokkos::Array<SizeT, N>& dim1_sizes,
+                    const Kokkos::Array<SizeT, N>& dim2_sizes,
+                    const Kokkos::Array<SizeT, N>& dim3_sizes,
+                    Kokkos::Array<ViewT, N>& views,
+                    bool do_transpose=false)
+{
+  using PackT = typename ViewT::value_type;
+  using ScalarT = typename PackT::scalar;
+
+  std::vector<ScalarT> tdata;
+  for (size_t n = 0; n < N; ++n) {
+    const size_t dim1_size = static_cast<size_t>(dim1_sizes[n]);
+    const size_t dim2_size = static_cast<size_t>(dim2_sizes[n]);
+    const size_t dim3_size = static_cast<size_t>(dim3_sizes[n]);
+    const size_t npack = (dim3_size + PackT::n - 1) / PackT::n;
+    views[n] = ViewT("", dim1_size, dim2_size, npack);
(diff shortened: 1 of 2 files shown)


src/ekat/ekat_pack_kokkos.hpp
```diff
@@ -312,6 +312,52 @@ void host_to_device(const Kokkos::Array<typename ViewT::value_type::scalar const
...
+      the_data = const_cast<ScalarT*>(data[n]);
+    }
+
+    for (size_t i = 0; i < dim1_size; ++i) {
+      for (size_t k = 0; k < dim2_size; ++k) {
+        for (size_t p = 0; p < npack; ++p) {
+          const size_t num_scalars_this_col = p*PackT::n;
+          const size_t scalar_offset = i*(dim2_size*dim3_size) + k*dim3_size + num_scalars_this_col;
+          for (size_t s = 0; s < PackT::n && num_scalars_this_col+s < dim3_size; ++s) {
+            host_view(i, k, p)[s] = the_data[scalar_offset + s];
```
bartgol, line 352: Wouldn't a simple

```
auto data_out = reinterpret_cast<ScalarT*>(host_view.data());
auto size = dim1_size*dim2_size*dim3_size;
std::memcpy(data_out,the_data,size);
```
work? It seems like this is a simple linear copy of memory, so relying on memcpy might be clearer/shorter (beside arguably being faster).

src/ekat/ekat_pack_kokkos.hpp
```diff
@@ -409,6 +471,54 @@ void device_to_host(const Kokkos::Array<typename ViewT::value_type::scalar*, N>&
...
+      the_data = data[n];
+    }
+
+    for (size_t i = 0; i < dim1_size; ++i) {
+      for (size_t k = 0; k < dim2_size; ++k) {
+        for (size_t p = 0; p < npack; ++p) {
+          const size_t num_scalars_this_col = p*PackT::n;
+          const size_t scalar_offset = i*(dim2_size*dim3_size) + k*dim3_size + num_scalars_this_col;
+          for (size_t s = 0; s < PackT::n && num_scalars_this_col+s < dim3_size; ++s) {
+            the_data[scalar_offset + s] = host_view(i, k, p)[s];
```
bartgol, line 510: Same commend (backwards) as before: why not just using memcpy, rather than the three for loops (which are arguably harder to follow)?

src/ekat/ekat_pack_kokkos.hpp
bartgol: Ah, nvm, it would not work if there's padding.

bartgol (approved):

ambrad (approved):

src/ekat/ekat_pack_kokkos.hpp
tcclevenger: It appears this would work, but I am having trouble getting it to pass the tests (some values are not being copied). I see you merged this branch, should we just leave it as is and maybe change in the future?

bartgol: I think memcpy was not correct, since you have to skip some entries in the view. If the host raw ptrs had the same size of the packed views (i.e., they had padding included), it would work, but as is, a raw memcpy would be wrong.

I think it's good as is.

PR #74, opened 2020-11-18
Some mods to EkatCreateUnitTest

bartgol:
Some minor changes to EkatCreateUnitTest.cmake, including comments, hard coded OMP env choices (rather than relying on defaults), and more correct specification of resource groups for threaded tests.


jgfouca (dismissed):

bartgol:
After further investication, this PR seems to not be necessary. Closing it for now.

PR #76, opened 2020-11-21
Implemented a Jaccard string similarity index function (and a test).

jeff-cohere:
This adds a Jaccard similarity index to our list of string utilities, along with a helper function that allows us to tokenize a string with a set of several delimiters.

### DIFF
--- a/src/ekat/util/ekat_string_utils.cpp
+++ b/src/ekat/util/ekat_string_utils.cpp
@@ -2,6 +2,7 @@
 #include "ekat/ekat_assert.hpp"
 
 #include <algorithm>
+#include <set>
 #include <sstream>
 
 namespace ekat {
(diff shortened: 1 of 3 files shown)


jeff-cohere:
Hrmmm. I did update the submodules recursively because of an error I was seeing, but I'm not sure I should have for this PR. @bartgol , should I re-sync the kokkos module to what it was before.

jeff-cohere:
Hey @bartgol , can you take a look at this and let me know what you think? I might need to rewind the `kokkos` submodule--not sure why it changed with the `update` I did in my workspace.

bartgol (dismissed):
Looks good and clear!

bartgol:
Yeah, it would be nice if this PR did not change the kokkos submodule. I'm guessing you were switching branches, and didn't update the submodules when you created this branch, then commited all the tree. Seems easy to fix, and probably best to do it.

bartgol (approved):

PR #79, opened 2020-12-02
Improvement to EkatCreateUnitTest

bartgol:
## Motivation

Some unit tests can manipulate resources (such as files) in a way that does not allow multiple versions of the test (corresponding to different ranks/threads combination) to run concurrently. In this case, we want the suite of test to run sequentially. We do this via the RESOURCE_LOCK property, rather than adding dependencies, since it allows for the other tests in the suite to still run even if one fails.
Additionally, while at it, I also added the possibility to add a set of properties to the whole suite (meaning the same properties attached to all the threads/ranks combinations of the test).

Also, I added a micro fix for an odd compilation error in `pack_kokkos_tests.cpp`, that I was getting on my laptop (gcc 8.2.0).

## Testing
I manually tested the features on the WSM tests, but did not commit it, since we are ok with EKAT tests running concurrently.

### DIFF
--- a/cmake/EkatCreateUnitTest.cmake
+++ b/cmake/EkatCreateUnitTest.cmake
@@ -21,6 +21,9 @@ include(EkatUtils) # To check macro args
 #    - LIBS_DIRS: a list of directories to add to the linker search path
 #    - LINKER_FLAGS: a list of additional flags for the linker
 #    - LABELS: a set of labels to attach to the test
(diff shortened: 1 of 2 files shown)


jeff-cohere (approved):
Looks good. If you've tested that these properties are treated properly by ctest, then I have a thumbs up for you. 👍

PR #80, opened 2020-12-10
host_to_device and device_to_host now use vectors

jgfouca:
host_to_device/device_to_host API now use std::vectors instead of Kokkos::Arrays for managing input arrays. Kokkos::Arrays are less convenient than std::vector for the user and should only be preferred over std::vector if the inputs will need to be used in kokkos kernels. This should also help reduce warnings in scream.

## Testing
This API is already well-tested. Added some additional tests for size_t vs. int SizeT.

### DIFF
--- a/cmake/EkatSetCompilerFlags.cmake
+++ b/cmake/EkatSetCompilerFlags.cmake
@@ -41,8 +41,8 @@ macro (SetCompilerFlags)
 
   # enable all warning but disable vectorization remarks like "remark: simd loop has only one iteration"
   # since we would get hit with 1000's of those anytime we set packsize to 1.
-  set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Werror")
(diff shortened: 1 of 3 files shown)


cmake/EkatSetCompilerFlags.cmake
```diff
@@ -41,8 +41,8 @@ macro (SetCompilerFlags)
 
   # enable all warning but disable vectorization remarks like "remark: simd loop has only one iteration"
   # since we would get hit with 1000's of those anytime we set packsize to 1.
-  set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall")
-  set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
+  set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Werror")
```
jgfouca, line 44: Whoops, didn't mean to add this.

bartgol (dismissed):

bartgol (approved):

PR #81, opened 2021-02-03
Implement subview utilities for rank 2-6 layout right views.

bartgol:
These routines are *very* similar to the ones implemented in HOMME. There are two kinds of utilities:
 - subview 1 or more of the slowest indices: the easy ones, since the subview is still contiguous.
 - subview the second slowest index: slightly less easy, since Kokkos::subview only preserve layout right under certain conditions (like output rank <=2).

## Motivation

These routines are handy shortcuts to Kokkos longer syntax (for the first kind of subview routines) and also improve Kokkos capability when it comes to subviewing the second slowest index. Such routines will be needed in SCREAM to generate subview of fields while avoiding the switch to LayoutStride.
 

## Testing

The new routines are all tested. The `subview_1` routines (which subview along the the second slowest dimension) are also tested when applied twice (to ensure that it works on an input view with funky strides).

### DIFF
--- a/src/ekat/kokkos/ekat_kokkos_types.hpp
+++ b/src/ekat/kokkos/ekat_kokkos_types.hpp
@@ -3,6 +3,7 @@
 
 #include "ekat/ekat_config.h"
 #include "ekat/kokkos/ekat_kokkos_meta.hpp"
+#include "ekat/std_meta/ekat_std_type_traits.hpp"
 
 /*
  * Header contains globally useful kokkos-related types for ekat.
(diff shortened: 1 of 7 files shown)


src/ekat/kokkos/ekat_kokkos_types.hpp
```diff
@@ -23,6 +24,9 @@ using DefaultDevice = Kokkos::Device<Kokkos::DefaultExecutionSpace, Kokkos::Defa
 // A device type to force host execution
 using HostDevice = Kokkos::Device<Kokkos::DefaultHostExecutionSpace, Kokkos::DefaultHostExecutionSpace::memory_space>;
 
+template<typename DT, typename... Props>
+using ViewLR = Kokkos::View<DT,Kokkos::LayoutRight,Props...>;
```
jeff-cohere, line 28: Do we ever use LayoutLeft views? If not, this symbol is probably descriptive enough.

jeff-cohere (dismissed):
Looks great! Very clean template magic.

src/ekat/kokkos/ekat_kokkos_types.hpp
jgfouca: @jeff-cohere , yes. I've used them in places where fortran compatibility is needed.

jeff-cohere (approved):

PR #86, opened 2021-02-17
Avoid printing lines of tildes in catch2

### LINKED ISSUE #85: Hack catch2.hpp to avoid char lines that can be parsed as markdown
Catch2 can use some char lines to separate sections. I'm referring to seeing something like

```
blah blah
~~~~~~~~~~~~~~~~~
other stuff
```

However, when the AT reports output in a github PR (due to tests failures), some lines can trigger markdown parsing. In particular, lines with 3+ tilde's (`~`) or backticks (`` ` ``) will trigger the opening of a code block. This can mess up the rendering of the message. For instance, see #E3SM-Project/scream810 : some AT messages reporting failures only show two machines (blake and weaver), but if you expand the weaver section you will notice that it also contains output from the mappy section (the missing one). The reason is that in the weaver output there was a line of `~` chars which github parses as opening a code block. Since no matching line is found, the whole msg is parsed as that code block, which therefore swallows all the msg that the AT was trying to forward to the PR.

Possible solution: hack line 15777 of catch2.hpp, to print a line of chars other than `~`. We could use `=` or `*`.

bartgol:
Github parses a line of tilde as a code block opening. If tests output is
written in a msg on a PR/issue, this causes parsing mess.

## Motivation

When the AutoTester writes a message on a PR, it adds outputs from the tests in case there are failures, but lines of `~` chars mess up the parsing.

## Related Issues

* Closes #85

### DIFF
--- a/extern/catch2/include/catch2/catch.hpp
+++ b/extern/catch2/include/catch2/catch.hpp
@@ -15774,7 +15774,7 @@ void ConsoleReporter::lazyPrintWithoutClosingBenchmarkTable() {
     }
 }
 void ConsoleReporter::lazyPrintRunInfo() {
-    stream << '\n' << getLineOfChars<'~'>() << '\n';
+    stream << '\n' << getLineOfChars<'='>() << '\n';
     Colour colour(Colour::SecondaryText);
     stream << currentTestRunInfo->name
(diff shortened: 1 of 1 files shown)


jeff-cohere (approved):
Approved, subject to successful testing.

PR #87, opened 2021-02-22
Remove const from ekat::find signature.

bartgol:
If the input container `c` is const, the type of `c.begin()`, is an iterator to const, which can later mess up fcn overload resolution when using the result to call some other function. If the container is indeed const, the 'const' will be part of `ContainerType` anyways, yielding the same behavior as before.

## Motivation

Consider this example:

```
std::list<std::string> names = {"Luca", "Jim"};
auto it = ekat::find(names,"Jim");
std::iter_swap(names.begin(),it); // error!
```
The compiler will bark an error, saying that `*it` is a const reference, so it cannot be modified. That's because `ekat::find` returns the type of `begin()` called on a const list, which is an iterator to const.

### DIFF
--- a/src/ekat/std_meta/ekat_std_utils.hpp
+++ b/src/ekat/std_meta/ekat_std_utils.hpp
@@ -11,7 +11,7 @@ namespace ekat {
 
 // This function returns an iterator which is of the same type of c.begin()
 template<typename ContainerType, typename T>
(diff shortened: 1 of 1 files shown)


jeff-cohere (approved):
Good ol' functional programming!

PR #88, opened 2021-03-01
Add support for using valgrind on all created tests

jgfouca:
CMAKE option is EKAT_ENABLE_VALGRIND. Supports automatic generation
of suppresion files which is necessary for clean runs of tests that use
MPI.

Discovered a minor memory problem in upper_bound test which I fixed.

Once this is merged and available to SCREAM, we can begin exploring options for having valgrind in our CI/nightly testing.

I tested this by hand. First, tested EKAT both with and without EKAT_ENABLE_VALGRIND. Without it, tests ran as before. With it, I checked ctest with verbose mode on and confirmed the injection of valgrind. I then ran all tests with valgrind on and noticed the valgrind problem in upper_bound.

Second, I repeated this process for SCREAM. SCREAM has a a number of valgrind issues, mostly in SHOC.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,7 @@ list(APPEND CMAKE_MODULE_PATH
 # (for openpmi; -lmpicxx for mpich) flag.
 include(EkatMpiUtils)
 DisableMpiCxxBindings()
+
 # Scripts containing macros needed to handle TPLs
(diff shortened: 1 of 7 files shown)


bartgol (approved):
Looks good. The fact that you already found an issue in ekat shows the importance of this feature.

PR #90, opened 2021-03-10
SCREAM: in debug mode, init workspace buffer with NaN

bartgol:
In debug mode, fill a slot with NaN during the 'take' calls.

## Motivation
Not initializing can cause hard-to-find bugs. This is b/c during a malloc request, if the OS provides a fresh new page (as opposed to a previous one being recycled) it is _always_ going to zero it out (for security reasons). This could hide some bugs. However, if during another execution a previous page is reused, no zeroing out happens, which can cause erratic behavior

NaN's should already cause some check to fail, should these arrays be used without being properly initialized. This is why it is enough to do this only in debug mode: once all bugs (due to uninited vals being used) are fixed, it is safe to let the WSM provide uninitialized memory, since we know that the provided arrays will be written before they are read.

## E3SM Stakeholder Feedback

This will help fixing some valgrind errors in scream.

## Testing

The WSM is already tested.

### DIFF
--- a/src/ekat/ekat_workspace.hpp
+++ b/src/ekat/ekat_workspace.hpp
@@ -253,8 +253,7 @@ class WorkspaceManager
   KOKKOS_INLINE_FUNCTION
   void init_metadata(const int ws_idx, const int slot) const;
 
-  static void init(const WorkspaceManager& wm, const view_2d<T>& data,
-                   const int max_ws_idx, const int max_used, const int total);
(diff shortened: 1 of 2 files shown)


bartgol:
@jgfouca I purged some args from a method of WSM, cause they were never used. I'm not an expert of WSM, so please let me know if I missed something.

jgfouca (dismissed):

jgfouca (approved):

PR #92, opened 2021-03-11
Bug fix to gather_tokens in ekat_string_utils.cpp

bartgol:
Fixes implementation of gather_tokens for the case where an atomic substring is passed

## Motivation

In the case where atomic!="" and the atomic string is found, we need to verify that it is preceded/followed by a delimiter
or the begin/end of s. Otherwise, we must proceed as if it wasn't found. E.g., when tokenizing `my_birthday_is_coming_soon` using `_` as delimiter, the atomic string `my_birth` should not yield a token. On the other hand, the atomic string `my_birthday` is a legit token.

## Testing

I added a test for the `gather_tokens` function.

### DIFF
--- a/src/ekat/util/ekat_string_utils.cpp
+++ b/src/ekat/util/ekat_string_utils.cpp
@@ -131,6 +131,10 @@ std::list<std::string> gather_tokens(const std::string& s,
                                      const std::vector<char>& delimiters,
                                      const std::string& atomic) {
   std::list<std::string> all_tokens;
+  std::string delim_str;
+  for (char delim: delimiters) {
+    delim_str.append(1, delim);
+  }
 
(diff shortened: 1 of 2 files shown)


src/ekat/util/ekat_string_utils.cpp
```diff
@@ -198,6 +234,9 @@ double jaccard_similarity (const std::string& s1, const std::string& s2,
     // We already took care of the case were we don't tokenize either one,
     // so we can be sure tokenize_s2 is true.
     // S2 is a single token
+      if (s2=="zenith_angle") {
```
jeff-cohere, line 237: Is this a debugging print?

jeff-cohere (approved):
Subtle!

PR #93, opened 2021-03-11
Units simplification, plus more string utils tests

bartgol:
## Motivation

The units class ended up storing more than it was actually needed. In particular, each `RationalConstant` object stored a `Format` member, specifying how to print it. Since the format is only needed when printing the class, it makes sense to simply pass it in when calling `to_string`, saving us from storing an int (which was 20% of the class size). This way is more logical, and the implementation is slightly leaner.

As for the string utils, there were so many things to test, that it felt wrong to mix them up with other utils in `util_tests.cpp`. It's more logical for all these utils to have their own test file. I also added a test for the `trim` utility, which wasn't tested.

## Testing

Everything was already tested.


src/ekat/util/ekat_rational_constant.hpp
```diff
@@ -36,8 +32,12 @@ constexpr long long gcd (const long long a, const long long b) {
...
 //       considered the scope of the usage of this class
 
-class RationalConstant {
-public:
+struct RationalConstant {
+
+  using iType = long long;
+
+  const iType num;
+  const iType den;
```
jeff-cohere, line 40: What do you think about `denom` instead of `den`? I'm just thinking about all the technical language that might be embedded in a given expression involving a rational constant, but maybe I'm overthinking it.

bartgol: I liked having the same short length for both (I would even have gone for `n` and `d`), but have no strong feeling. Do you think there might be ambiguities with the name 'den'?

jeff-cohere: Probably not. There is always "density", which appears everywhere, but likely not in this context. I guess it's more of a cadence between the syllables in **num**erator and **denom**inator, but this is very subjective. I don't have strong feelings. :-)

jeff-cohere (approved):
Looks great, Luca.

src/ekat/util/ekat_rational_constant.hpp
bartgol: Yeah, I don't think this class will be used outside the context of the Units framework impl, so ppl should not even see these when using Units in their code...

PR #95, opened 2021-03-17
Some modifications to Pack class

bartgol:
Adds a default copy ctor to Pack, and changes template args of most fcns in ekat_pack.hpp from `MaskType` or `PackType` to scalar type `T` and pack size `n`, where possible.

## Motivation
The main modification is to add a copy ctor to Pack. The current one is templated, which means its signature does not match that of a default copy ctor. This causes the compiler to still generate a copy ctor, which is selected when doing stuff like `auto a = b;`. For some reason, the default copy ctor was causing buggy behavior on blake. It might be this is a red herring, but at the very least, the default copy ctor was not vectorized, which is one more reason to spell out our copy ctor.

For the template signature changes, the implementation is actually shorter when templating on `n` (and possibly `T`). Moreover, by encoding `n` in the input pack/mask types, we avoid the need to check compatibility of sizes.

## Testing

No additional testing added.

### DIFF
--- a/src/ekat/ekat_pack.hpp
+++ b/src/ekat/ekat_pack.hpp
@@ -223,33 +223,33 @@ using OnlyPackReturn = typename std::enable_if<PackType::packtag,ReturnType>::ty
 // promote a pack's scalar type in mixed-type arithmetic.
 
 #define ekat_pack_gen_bin_op_pp(op)                                   \
-  template <typename PackType>                                          \
+  template <typename T, int n>                                          \
   KOKKOS_FORCEINLINE_FUNCTION                                           \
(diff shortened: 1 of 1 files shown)


jgfouca (approved):

jgfouca:
Nice! We'll have to remember to undo the hacks that I added to SHOC the next time we do an EKAT update.

bartgol:
> Nice! We'll have to remember to undo the hacks that I added to SHOC the next time we do an EKAT update.

I'm planning on undoing them (well, "it", since one was in the inputs initializer, which got purged) in #E3SM-Project/scream926.

PR #97, opened 2021-04-02
Aarondonahue/any cast throw exception

AaronDonahue:
This PR introduces the function `isType` to the YAML parameter list parser.

`isType` is a templated function that allows the user to check if a parameter in a parameter list is
of a specific type.  This is particularly useful in cases where a YAML node may have the option to be
of multiple types. 

## Motivation
Allows a node in a YAML file to be given different types of entries, providing flexibility for applications.

## E3SM Stakeholder Feedback
Addresses SCREAM issue: https://github.com/E3SM-Project/scream/issues/975

## Testing
This new function is tested in the `yaml_parser` test.

### DIFF
--- a/src/ekat/ekat_parameter_list.hpp
+++ b/src/ekat/ekat_parameter_list.hpp
@@ -62,6 +62,10 @@ public:
   bool isParameter (const std::string& name) const { return m_params.find(name)!=m_params.end(); }
   bool isSublist   (const std::string& name) const { return m_sublists.find(name)!=m_sublists.end(); }
 
+  // Check methods, to determine the type of a node
+  template<typename T>
+  bool isType (const std::string& name);
+
   // Display the sublist.
   // NOTE: this *requires* op<< to be overloaded for all the stored parameters.
   //       The code won't crash otherwise, but instead of the parameter, you
@@ -111,6 +115,19 @@ inline void ParameterList::set (const std::string& name, const T& value) {
   } else {
     get<T>(name) = value;
   }
+}
+
+template<typename T>
+inline bool ParameterList::isType (const std::string& name) {
+  // Try to get variable with type T, return true if successful, false if an error is thrown
+  try {
+    auto& dummy = any_cast<T>(m_params.at(name));
+    return true;
+  }
+  catch (std::exception& e) {
+    // Do nothing with catch
+  }
+  return false;
 }
 
 } // namespace ekat
--- a/src/ekat/ekat_parameter_list.hpp
+++ b/src/ekat/ekat_parameter_list.hpp
@@ -64,7 +64,7 @@ public:
 
   // Check methods, to determine the type of a node
   template<typename T>
-  bool isType (const std::string& name);
+  const bool isType (const std::string& name);
 
   // Display the sublist.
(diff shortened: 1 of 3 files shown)


src/ekat/ekat_parameter_list.hpp
```diff
@@ -62,6 +62,10 @@ class ParameterList {
   bool isParameter (const std::string& name) const { return m_params.find(name)!=m_params.end(); }
   bool isSublist   (const std::string& name) const { return m_sublists.find(name)!=m_sublists.end(); }
 
+  // Check methods, to determine the type of a node
+  template<typename T>
+  bool isType (const std::string& name);
```
bartgol, line 67: This method should be marked const.

src/ekat/ekat_parameter_list.hpp
```diff
@@ -113,6 +117,19 @@ inline void ParameterList::set (const std::string& name, const T& value) {
   }
 }
 
+template<typename T>
+inline bool ParameterList::isType (const std::string& name) {
+  // Try to get variable with type T, return true if successful, false if an error is thrown
+  try {
+    auto& dummy = any_cast<T>(m_params.at(name));
```
bartgol, line 124: Before entering the try block, we should check that the parameter exists. If it doesn't, we should throw and notify the user.

Note: it must be done outside the try block, cause otherwise the exception we throw would be caught by the catch block.

src/ekat/ekat_parameter_list.hpp
```diff
@@ -113,6 +117,19 @@ inline void ParameterList::set (const std::string& name, const T& value) {
...
+    return true;
+  }
+  catch (std::exception& e) {
```
bartgol, line 127: I believe some compilers might complain that `e` is defined but not used. Also, you can catch by _const_ reference. So in the end, something like `catch (const std::exception&)`.

bartgol (changes requested):
Looks good. There's only one important change (const method) to be made, plus a couple other minor things maybe.

AaronDonahue:
@bartgol , Thank you for the quick review.  I think I have addressed all of your comments in the most recent commit.

src/ekat/ekat_parameter_list.hpp
```diff
@@ -64,7 +64,7 @@ class ParameterList {
 
   // Check methods, to determine the type of a node
   template<typename T>
-  bool isType (const std::string& name);
+  const bool isType (const std::string& name);
```
bartgol, line 67: Sorry, this is not what I meant with marking the method const. This means that the returned bool is constant (and a decent compiler will actually issue a warning here, since you're returning by value, so const is pointless). To mark the method const, you need to put `const` _after_ the signature, right before the semicolon:

```
bool isType (...) const;
```

bartgol (changes requested):

src/ekat/ekat_parameter_list.hpp
AaronDonahue: Ahhh, my bad.  I think it is fixed now.

bartgol (approved):

Issue #100, opened 2021-04-03
Add Pack::set(mask, const Pack pack_true, const Pack pack_false) function.


overfelt:
I would like to have an if/then/else type set in order to replace
  A.set(mask, B);
  A.set(!mask, C); 
with
  A.set(mask,B,C);
From what I have have read in SIMD tutorials, this would directly translate into a SIMD call that is widely available and could default to the two assignment operators if not.

bartgol:
@overfelt I suggested precisely this when I was reviewing a PR in scream. It would be much better than doing

```
A = B;
A.set(mask,C);
```
which is much harder to read.

I can get on this today.

overfelt:
Thanks.

PR #101, opened 2021-04-05
Upgrades to Pack's masked operations

### LINKED ISSUE #99: Have Pack::set() return Pack &.
I would like to chain set() with other operators e.g. 
   p1 = log(p0.set(p0==0, -large_val)) 
   p2 = exp(p1).set(mask, x0)

The set() is like operator= which returns a reference and the return value can always be ignored if not used.

bartgol:
## Motivation

This PR does two things: 1) allows using Pack's `set` method inside chains of ops, and 2) add a false value for masked ops.

The first mod is just for ease of use, like `auto p2 = exp (p1.set(m,p3));`. The second is also for ease of use, but also has a minor performance benefit. It allows to change
```
Pack p1 = val1;
p1.set(mask1, val2);
Pack p2;
...
p2 = p1;
p2.set(mask2,p3);
```
into
```
Pack p1(mask1.val1,val2);
Pack p2;
...
p2.set(mask2,p3,p1);
```
Notice that setting both values for mask=true and mask=false allows to only do one loop.

## Related Issues

* Closes #99 #100 

## E3SM Stakeholder Feedback

This might come handy in some scream code (e.g., cut a line of code here and there).

## Testing

Added tests for masked constructors and set operations, as well as for chaining set with other operations.

### DIFF
--- a/src/ekat/ekat_pack.hpp
+++ b/src/ekat/ekat_pack.hpp
@@ -76,6 +76,15 @@ private:
   type d[n];
 };
 
+template <int n>
+inline std::ostream&
+operator << (std::ostream& os, const Mask<n>& m) {
+  for (int i=0; i<n; ++i) {
+    os << m[i] << ' ';
+  }
+  return os;
+}
+
 // Codify how a user can construct their own loops conditioned on mask slot
 // values.
 #define ekat_masked_loop(mask, s)                         \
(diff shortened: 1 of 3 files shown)


tcclevenger (approved):
I think this will clean up some shoc code quite a bit. I'll check that out once it's merged. Thanks @bartgol !

PR #102, opened 2021-04-06
Update kokkos submodule

bartgol:
Rebase onto kokkos v3.3.00 tag.

## Motivation

It turns out that our version of kokkos does not have a team-level parallel_scan on CUDA. Version 3.3.00 does, so bring in that, by rebasing our current version onto kokkos' v3.3.00 tag. Create a new kokkos tag/branch, to leave current version intact and working.

## E3SM Stakeholder Feedback

PR E3SM-Project/scream#954 depends on this.

## Testing

No additional testing.

### DIFF
--- a/src/ekat/kokkos/ekat_kokkos_utils.hpp
+++ b/src/ekat/kokkos/ekat_kokkos_utils.hpp
@@ -155,7 +155,9 @@ void view_reduction (const TeamMember& team,
   // so manually add the last pack (only the non-garbage part)
   if (has_garbage_end) {
     const PackType temp_input = input(pack_loop_end);
-    const int last_indx = end%vector_size;
+    // The following is morally a const var, but there are issues with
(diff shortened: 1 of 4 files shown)


bartgol:
Ugh, after some digging, it appears that there's an issue in gcc 7.2 with std=c++14 (which is what kokkos 3.3 uses), and apparently it got fixed in gcc 7.3. Unfortunately, there is _no_ openmpi installation on weaver that uses gcc>7.3, so I can't verify that, until I'm done installing my own openmpi (with a newer gcc).

bartgol:
After trying openmpi built against gcc 7.4 without success, I realized this is another instance of the compiler bug that brought to the declaration of the [ConstExceptGnu](https://github.com/E3SM-Project/EKAT/blob/9d097dd279cb152ea75bc12071245c7c24884088/src/ekat/kokkos/ekat_kokkos_types.hpp#L16) macro.

Using this macro in the place that seemed to cause issues, fixes the compilation. I'll push the fix soon.

jgfouca (approved):

PR #103, opened 2021-04-09
Removed warning from ParameterList method

bartgol:
Also, reworked impl details in ekat::any.

## Motivation

There was a warning in the impl of `bool ParameterList::isType<T>()` due to unused var. While removing it, I improved a bit the function, by improving `any`'s interface, so that `ParameterList` doesn't need a try-catch block for the impl of `isType`.

## Testing

Covered by same testing as before.


bartgol:
Adding @jgfouca as reviewer, since this addresses also the warning issue.

jgfouca (approved):

PR #104, opened 2021-04-10
Minor change to remove unused warning

jgfouca:
Trick compiler into thinking variable is used.

## Motivation
Fewer warnings when building SCREAM.

## Testing
None


bartgol:
Actually, #103 already solves the issue by changing a bit the impl of the any class (and its utilities). I'm gonna close this PR in favor of that one.

PR #105, opened 2021-04-20
add Workspace::reserve_n_contiguous_sub_blocks

tcclevenger:
Adds a function `Workspace::reserve_n_contiguous_sub_blocks()` which can reserve a certain number of sub-blocks defined at runtime.

## Motivation
The other contiguous function does not allow for variable size.

## Testing
Still need to add a test..

### DIFF
--- a/src/ekat/ekat_workspace.hpp
+++ b/src/ekat/ekat_workspace.hpp
@@ -142,7 +142,7 @@ class WorkspaceManager
     //                                                     n, m_sizes);
     template <typename S=T>
     KOKKOS_INLINE_FUNCTION
-    Unmanaged<view_1d<S> > take_n_size_block(const char* name, const int n) const;
+    Unmanaged<view_1d<S> > take_n_size_block(const char* name, const int n_sub_blocks) const;
 
     // Combines reset and take_many_contiguous_unsafe. This is the most-performant
     // option for a kernel to use N sub-blocks that are needed for the duration of the
--- a/src/ekat/ekat_workspace.hpp
+++ b/src/ekat/ekat_workspace.hpp
@@ -131,6 +131,23 @@ class WorkspaceManager
     void take_many_contiguous_unsafe(const Kokkos::Array<const char*, N>& names,
                                      const view_1d_ptr_array<S, N>& ptrs) const;
 
+    // Take an individual sub-block and reserve enough contiguous memory for
+    // the next n-1 sub-blocks. This is useful when the number of sub-blocks
+    // needed is a runtime variable.
+    //
+    // Example: A 2d view of size (n, wsm_size) where n is a runtime variable
+    //          and wsm_size is the size used to construct the WorkspaceManager
+    // Code:
+    //   const auto local_slot = workspace.reserve_n_contiguous_sub_blocks("local_slot",n);
+    //   const auto local_var  = Unmanaged2dViewT<ScalarT>(reinterpret_cast<ScalarT*>(local_slot.data()),
+    //                                                     n, wsm_size);
+    //
+    // Since the memory is contiguous, we can just resize the returned sub-block and
+    // be guarenteed not to write out-of-bounds.
+    template <typename S=T>
+    KOKKOS_INLINE_FUNCTION
+    Unmanaged<view_1d<S> > reserve_n_contiguous_sub_blocks(const char* name, const int n) const;
+
     // Combines reset and take_many_contiguous_unsafe. This is the most-performant
     // option for a kernel to use N sub-blocks that are needed for the duration of the
     // kernel.
@@ -149,6 +166,11 @@ class WorkspaceManager
     template <size_t N, typename S=T>
     KOKKOS_INLINE_FUNCTION
     void release_many_contiguous(const view_1d_ptr_array<S, N>& ptrs) const;
+
+    // Release input sub-block as well as the next n-1 sub-blocks.
+    template <typename S=T>
+    KOKKOS_INLINE_FUNCTION
+    void release_n_contiguous_sub_blocks(const Unmanaged<view_1d<S> >& space, const int num_slots) const;
 
 #ifndef NDEBUG
     // Get the name of a sub-block
--- a/src/ekat/ekat_workspace.hpp
+++ b/src/ekat/ekat_workspace.hpp
@@ -166,7 +166,7 @@ class WorkspaceManager
     // Release block of size n*m_size.
     template <typename S=T>
     KOKKOS_INLINE_FUNCTION
-    void release_n_size_block(const Unmanaged<view_1d<S> >& space, const int n) const;
+    void release_n_size_block(const Unmanaged<view_1d<S> >& space, const int n_sub_blocks) const;
 
 #ifndef NDEBUG
     // Get the name of a sub-block
--- a/src/ekat/ekat_workspace.hpp
+++ b/src/ekat/ekat_workspace.hpp
@@ -131,9 +131,10 @@ class WorkspaceManager
     void take_many_contiguous_unsafe(const Kokkos::Array<const char*, N>& names,
                                      const view_1d_ptr_array<S, N>& ptrs) const;
 
-    // Take a block of size n*m_size, where m_size is the size used to construct
-    // the WorkspaceManager. This is useful for creating local 2d views through
-    // the WorkspaceManager.
+    // Take an individual sub-block while telling the WorkSpaceManager to skip over the
+    // next n_sub_blocks-1 sub-blocks. This allows the user to safely access the memory
+    // of these sub-block, which is useful for creating local 2d views through the
+    // WorkspaceManager.
     //
     // Example: Local 2d view of size (n, m_size).
     // Code:
@@ -166,7 +167,7 @@ class WorkspaceManager
     // Release block of size n*m_size.
     template <typename S=T>
     KOKKOS_INLINE_FUNCTION
-    void release_n_size_block(const Unmanaged<view_1d<S> >& space, const int n_sub_blocks) const;
+    void release_macro_block(const Unmanaged<view_1d<S> >& space, const int n_sub_blocks) const;
 
 #ifndef NDEBUG
     // Get the name of a sub-block
@@ -267,10 +268,6 @@ class WorkspaceManager
   template <typename S=T>
   KOKKOS_FORCEINLINE_FUNCTION
   Unmanaged<view_1d<S> > get_space_in_slot(const int team_idx, const int slot) const;
-
-  template <typename S=T>
-  KOKKOS_FORCEINLINE_FUNCTION
-  Unmanaged<view_1d<S> > get_n_spaces_in_slot(const int team_idx, const int slot, const int n) const;
 
   KOKKOS_INLINE_FUNCTION
   void init_metadata(const int ws_idx, const int slot) const;
--- a/src/ekat/ekat_workspace.hpp
+++ b/src/ekat/ekat_workspace.hpp
@@ -142,7 +142,7 @@ class WorkspaceManager
     //                                                     n, m_sizes);
     template <typename S=T>
     KOKKOS_INLINE_FUNCTION
-    Unmanaged<view_1d<S> > take_n_size_block(const char* name, const int n_sub_blocks) const;
+    Unmanaged<view_1d<S> > take_macro_block(const char* name, const int n_sub_blocks) const;
 
     // Combines reset and take_many_contiguous_unsafe. This is the most-performant
     // option for a kernel to use N sub-blocks that are needed for the duration of the
--- a/src/ekat/ekat_workspace.hpp
+++ b/src/ekat/ekat_workspace.hpp
@@ -131,22 +131,18 @@ class WorkspaceManager
     void take_many_contiguous_unsafe(const Kokkos::Array<const char*, N>& names,
                                      const view_1d_ptr_array<S, N>& ptrs) const;
 
-    // Take an individual sub-block and reserve enough contiguous memory for
-    // the next n-1 sub-blocks. This is useful when the number of sub-blocks
-    // needed is a runtime variable.
+    // Take a block of size n*m_size, where m_size is the size used to construct
+    // the WorkspaceManager. This is useful for creating local 2d views through
+    // the WorkspaceManager.
     //
-    // Example: A 2d view of size (n, wsm_size) where n is a runtime variable
-    //          and wsm_size is the size used to construct the WorkspaceManager
+    // Example: Local 2d view of size (n, m_size).
     // Code:
-    //   const auto local_slot = workspace.reserve_n_contiguous_sub_blocks("local_slot",n);
+    //   const auto local_slot = workspace.take_n_size_block("local_slot",n);
     //   const auto local_var  = Unmanaged2dViewT<ScalarT>(reinterpret_cast<ScalarT*>(local_slot.data()),
-    //                                                     n, wsm_size);
-    //
-    // Since the memory is contiguous, we can just resize the returned sub-block and
-    // be guarenteed not to write out-of-bounds.
+    //                                                     n, m_sizes);
     template <typename S=T>
     KOKKOS_INLINE_FUNCTION
-    Unmanaged<view_1d<S> > reserve_n_contiguous_sub_blocks(const char* name, const int n) const;
+    Unmanaged<view_1d<S> > take_n_size_block(const char* name, const int n) const;
 
     // Combines reset and take_many_contiguous_unsafe. This is the most-performant
     // option for a kernel to use N sub-blocks that are needed for the duration of the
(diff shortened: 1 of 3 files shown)


tcclevenger:
@jgfouca  and @bartgol. Is this reasonable? I use this for defining a 2d view in SHOC as a workspace variable. I didn't see an obvious way to use the other contiguous function since the number of columns I need is a runtime variable.

Example of how I use it:
```
 // 2d allocations for solver RHS
  const auto tracers_slot = workspace.take_n_contiguous_slots("tracers_slot",num_tracers);

  // Reshape 2d views
  const auto local_tracers = uview_2d<Spack>(reinterpret_cast<Spack*>(tracers_slot.data()),
                                             num_tracers, npack(nlev));
```

It's label WIP since I still need to include a test and add some DEBUG checks in the implementation, but wanted to make sure I am on the right track.

jgfouca:
@tcclevenger , I think this concept could work fine but could be a bit tricky. Taking a stab at implementation:

```
WorkspaceManager<T, D>::Workspace::reserve_n_contiguous_sub_blocks(
  const char* name, const int num_slots) const
{
#ifndef NDEBUG
  change_num_used(N);
  // Verify contiguous
  for (int n = 0; n < num_slots; ++n) {
    const auto space = m_parent.get_space_in_slot<S>(m_ws_idx, m_next_slot + n);
    EKAT_KERNEL_ASSERT(m_parent.get_next<S>(space) == m_next_slot + n + 1);
  }
#endif

  const auto return_space = m_parent.get_space_in_slot<S>(m_ws_idx, m_next_slot);
  // need to set all values (num_slots * m_size) to invalid in debug?

  // We need a barrier here so get_space_in_slot above returns consistent results
  // w/in the team.
  m_team.team_barrier();
  Kokkos::single(Kokkos::PerTeam(m_team), [&] () {
    m_next_slot += num_slots;
#ifndef NDEBUG
    change_indv_meta<S>(return_space, name);
#endif
  });
  // We need a barrier here so that a subsequent call to take or release
  // starts with the metadata in the correct state.
  m_team.team_barrier();
}
```

I'm a little worried that this new 2d view will include data that's normally reserved for use as metadata within Workspace, but apparently that metadata is not used for contiguous operations. We will need to test this pretty thoroughly.

bartgol:
Isn't `take_many_contiguous_unsafe` still doing the same thing? I mean, it returns N pointers, but wouldn't those pointers refer to contiguous slots? If so, one could call `take_many_contiguous_unsafe`, and then use only the ptr from the 1st slot.

OTOH, a method that has `unsafe` in the name might raise concerns, so I'm a bit hesitant using it. @jgfouca what's 'unsafe' in that call? Is it just that there may not be N contiguous slots and the code would throw? Or is there more?

tcclevenger:
> Isn't `take_many_contiguous_unsafe` still doing the same thing? I mean, it returns N pointers, but wouldn't those pointers refer to contiguous slots? If so, one could call `take_many_contiguous_unsafe`, and then use only the ptr from the 1st slot.
The issue is that `take_many_contiguous_unsafe<N>(names,ptr)` requires the size to be known at runtime. Instead of a Kokkos array of views there, we could use a 1d view of pointers?

bartgol:
> > Isn't `take_many_contiguous_unsafe` still doing the same thing? I mean, it returns N pointers, but wouldn't those pointers refer to contiguous slots? If so, one could call `take_many_contiguous_unsafe`, and then use only the ptr from the 1st slot.
> 
> The issue is that `take_many_contiguous_unsafe<N>(names,ptr)` requires the size to be known at compile time. Instead of a Kokkos array of views there, we could use a 1d view of pointers?

Ah, good point! It looks to me as if N has to be compile time only b/c of the names/pointers arrays. The fact that the loop has compile-time length might allow loop unrolling, but I doubt we'd notice the difference in the big scheme of things. Re: view_1d of pointers instead of array: that would require a dyn allocation within a kernel (while a Kokkos::Array is an automatic variable), so I think that's out of the picture.

I think the new method is indeed needed. Imho, I would keep the method name equal to the existing one. A different name might suggest that it's doing something intrinsically different, while the only difference is that it returns a ptr to the 1st slot only, without the ptrs to the other ones.

Also, it'd be nice to implement the current method (with N ptrs) in terms of the new one, but this might not be possible b/c of change_indv_meta implementation.

src/ekat/ekat_workspace_impl.hpp
```diff
@@ -152,6 +152,25 @@ WorkspaceManager<T, D>::get_space_in_slot(const int team_idx, const int slot) co
...
+template <typename T, typename D>
+template <typename S>
+KOKKOS_FORCEINLINE_FUNCTION
+Unmanaged<typename WorkspaceManager<T, D>::template view_1d<S> >
+WorkspaceManager<T, D>::get_n_spaces_in_slot(const int team_idx, const int slot, const int n) const
+{
+  Unmanaged<view_1d<S> > space(
+    reinterpret_cast<S*>(&m_data(team_idx, slot*m_total) + m_reserve),
+    n*(sizeof(T) == sizeof(S) ?
+    m_size :
```
jgfouca, line 164: I think you want the `n*` multiplier on this line

src/ekat/ekat_workspace_impl.hpp
```diff
@@ -152,6 +152,25 @@ WorkspaceManager<T, D>::get_space_in_slot(const int team_idx, const int slot) co
...
+    (m_size*sizeof(T))/sizeof(S)));
```
jgfouca, line 165: And this line

src/ekat/ekat_workspace_impl.hpp
```diff
@@ -152,6 +152,25 @@ WorkspaceManager<T, D>::get_space_in_slot(const int team_idx, const int slot) co
...
 
+template <typename T, typename D>
+template <typename S>
+KOKKOS_FORCEINLINE_FUNCTION
+Unmanaged<typename WorkspaceManager<T, D>::template view_1d<S> >
+WorkspaceManager<T, D>::get_n_spaces_in_slot(const int team_idx, const int slot, const int n) const
+{
+  Unmanaged<view_1d<S> > space(
+    reinterpret_cast<S*>(&m_data(team_idx, slot*m_total) + m_reserve),
+    n*(sizeof(T) == sizeof(S) ?
```
jgfouca, line 163: do not want n* here

src/ekat/ekat_workspace_impl.hpp
```diff
@@ -486,6 +535,29 @@ void WorkspaceManager<T, D>::Workspace::release_many_contiguous(
...
 }
 
+template <typename T, typename D>
+template <typename S>
+KOKKOS_INLINE_FUNCTION
+void WorkspaceManager<T, D>::Workspace::release_n_size_block(
+  const Unmanaged<view_1d<S> >& space, const int n) const
+{
+#ifndef NDEBUG
+  change_num_used(-n);
```
jgfouca, line 545: I think this is correct, we can't check metadata here since it may have been overwritten by application data.

src/ekat/ekat_workspace_impl.hpp
```diff
@@ -241,6 +260,36 @@ void WorkspaceManager<T, D>::Workspace::take_many_contiguous_unsafe(
...
 
+template <typename T, typename D>
+template <typename S>
+KOKKOS_INLINE_FUNCTION
+Unmanaged<typename WorkspaceManager<T, D>::template view_1d<S> >
+WorkspaceManager<T, D>::Workspace::take_n_size_block(
+  const char* name, const int n) const
+{
+#ifndef NDEBUG
+  change_num_used(n);
```
jgfouca, line 271: I think we should keep the check that the data is contiguous here. This is similar to the check at the top of take_many_contiguous_unsafe.

jgfouca (changes requested):

src/ekat/ekat_workspace_impl.hpp
bartgol: Check the parentheses: sneaky @tcclevenger has wrapped the whole ?: ternary op in the parentheses, I think.

src/ekat/ekat_workspace_impl.hpp
bartgol: Ah. Are you saying we are returning something like nlev*sizeof(double)+room_for_metadata _per slot_, but since we're using all those slots as a single 2d array, we are writing actual data in the slots where usually we keep metadata? And are you saying this is not a pb, since the release method for the 2d array is not gonna check those?

src/ekat/ekat_workspace.hpp
```diff
@@ -131,6 +131,19 @@ class WorkspaceManager
...
+    // the WorkspaceManager.
+    //
+    // Example: Local 2d view of size (n, m_size).
+    // Code:
+    //   const auto local_slot = workspace.take_n_size_block("local_slot",n);
+    //   const auto local_var  = Unmanaged2dViewT<ScalarT>(reinterpret_cast<ScalarT*>(local_slot.data()),
+    //                                                     n, m_sizes);
+    template <typename S=T>
+    KOKKOS_INLINE_FUNCTION
+    Unmanaged<view_1d<S> > take_n_size_block(const char* name, const int n) const;
```
bartgol, line 145: The name of the method seems confusing. I guess it should be read as "take n size-sized blocks"? Or "take a NxSize block"?

How about something like `take_spliced_blocks`? Or another word (instead of splice) that conveys the msg that you are joining n blocks? `take_macro_block`? `take_super_block`?

src/ekat/ekat_workspace.hpp
```diff
@@ -250,6 +268,10 @@ class WorkspaceManager
   KOKKOS_FORCEINLINE_FUNCTION
   Unmanaged<view_1d<S> > get_space_in_slot(const int team_idx, const int slot) const;
 
+  template <typename S=T>
+  KOKKOS_FORCEINLINE_FUNCTION
+  Unmanaged<view_1d<S> > get_n_spaces_in_slot(const int team_idx, const int slot, const int n) const;
```
bartgol, line 273: Might be worth putting a comment above, specifying that the n spaces are contiguous?

bartgol (dismissed):
I think it looks good. I have a suggestion regarding names, but being that subjective, I'm not gonna block the PR.

src/ekat/ekat_workspace_impl.hpp
tcclevenger: Yeah, It looks ugly though. I changed it.

src/ekat/ekat_workspace_impl.hpp
jgfouca: @bartgol , yes that's exactly what I mean.

src/ekat/ekat_workspace_impl.hpp
jgfouca: Ah, my bad.

src/ekat/ekat_workspace_impl.hpp
bartgol: Side question: is it worth adding a check in get_index, that makes sure the input 1d view is "aligned", meaning that the user didn't accidentally pass in the wrong view? It would require checking that the `space.data()-m_data = K*m_size`, with 0<=0<m_max_used.

Or is it an overkill?

src/ekat/ekat_workspace_impl.hpp
```diff
@@ -241,6 +260,42 @@ void WorkspaceManager<T, D>::Workspace::take_many_contiguous_unsafe(
...
+WorkspaceManager<T, D>::Workspace::take_n_size_block(
+  const char* name, const int n_sub_blocks) const
+{
+#ifndef NDEBUG
+  change_num_used(n_sub_blocks);
+  // Verify contiguous
+  for (int n = 0; n < n_sub_blocks; ++n) {
+    const auto space = m_parent.get_space_in_slot<S>(m_ws_idx, m_next_slot + n);
+    EKAT_KERNEL_ASSERT(m_parent.get_next<S>(space) == m_next_slot + n + 1);
+    m_team.team_barrier();
```
tcclevenger, line 276: @bartgol @jgfouca Feel free to ignore this till next week, but without this team barrier (or if the team barrier exists outside of this for-loop), `m_parent.get_next<S>(space)` has a nonsense value for some of the spaces when run with NUM_THREADS>1. Even when there is just one call to this function (i.e., teampolicy(1) and only taking slots via this method).

Any ideas why this would be the case? It seems almost identical the check in the above function, so I don't know why a barrier is needed here.

src/ekat/ekat_workspace.hpp
tcclevenger: I like `take_macro_block`

src/ekat/ekat_workspace_impl.hpp
jgfouca: That is strange. Maybe try removing this barrier and running with valgrind?

src/ekat/ekat_workspace_impl.hpp
```diff
@@ -274,15 +254,15 @@ WorkspaceManager<T, D>::Workspace::take_macro_block(
   for (int n = 0; n < n_sub_blocks; ++n) {
     const auto space = m_parent.get_space_in_slot<S>(m_ws_idx, m_next_slot + n);
     EKAT_KERNEL_ASSERT(m_parent.get_next<S>(space) == m_next_slot + n + 1);
-    m_team.team_barrier();
   }
 #endif
 
-  const auto space = m_parent.get_n_spaces_in_slot<S>(m_ws_idx, m_next_slot, n_sub_blocks);
+  const auto space = m_parent.get_space_in_slot<S>(m_ws_idx, m_next_slot);
```
bartgol, line 260: What if there are no N consecutive blocks? I think this code would still work, and lead to hard to detect bugs... Am I overthiking this? If not, we should try to put some sort of check to make sure n consecutive blocks are indeed available...

tests/kokkos/workspace_tests.cpp
```diff
@@ -152,27 +152,33 @@ static void unittest_workspace()
...
 
     Kokkos::Array<Unmanaged<view_1d<int> >, num_ws> wssub;
+    Unmanaged<view_2d<int> wsmacro;
+    const auto space = ws.take_macro_block("space", num_ws);
 
     // Main test. Test different means of taking and release spaces.
-    for (int r = 0; r < 8; ++r) {
-      if (r % 4 == 0) {
+    for (int r = 0; r < 10; ++r) {
+      if (r % 5 == 0) {
```
bartgol, line 161: Maybe this is just me, but...this switch on `r % 5` it's' a bit hard to follow. Since the code shared by all code paths is minimal (that I can tell), wouldn't be easier to just get rid of the outer for?

src/ekat/ekat_workspace_impl.hpp
tcclevenger: Is it enough that above we call `get_space_in_slot(m_ws_idx, m_next_slot+n)` for `n=0,...,n_sub_blocks`?

tests/kokkos/workspace_tests.cpp
tcclevenger: Yeah, I'm not sure why it was written this way. Maybe it is so you have to take and release and then take again?

src/ekat/ekat_workspace_impl.hpp
bartgol: Ah, right, I missed that. Then it's fine, I think.

bartgol: But isn't this check too much at the last iter?
```
EKAT_KERNEL_ASSERT(m_parent.get_next<S>(space) == m_next_slot + n + 1);
```
We should do this check only for the first n-1 blocks, right?

bartgol (dismissed):

src/ekat/ekat_workspace_impl.hpp
tcclevenger: Good point. Should this also be changed for the check in `take_many_contiguous_unsafe` as well?

bartgol: I think so, yeah.

bartgol: But wait, if it's indeed wrong, how come it never failed so far? Are all tests/uses of the take many case always in a setup where there are always more slots after?

bartgol (approved):

src/ekat/ekat_workspace_impl.hpp
tcclevenger: I think that since `get_next()` is only run on spaces we will have initialized, the metadata for the last space should still have the correct next data?

tcclevenger: Even if the next slot cannot be reserved..

bartgol: Makes sens.

jgfouca: We just never hit the where the case where the n+1 slot was not contiguous. With individual takes/releases happening after the contiguous block, this mistake would have been uncovered. Nice catch!

jgfouca (approved):

tests/kokkos/workspace_tests.cpp
jgfouca: The idea is you cycle through 4 (now 5) different actions twice.

PR #106, opened 2021-04-28
Yet another overload of pack's op= with volatile cv

bartgol:
This one assumes the src object to be volatile.

## Motivation

In scream, we were getting some errors claiming no overload of `Pack::operator=(volatile Pack)` was found.

## Testing

No additional testing. But I verified that this PR fixes the compilation error in scream.


jeff-cohere (approved):
Sorry, this fell off my radar! Approved.

PR #108, opened 2021-05-13
Add scan_sum feature to Comm class

bartgol:
## Motivation

The Comm class is a convenience wrapper around MPI_Comm. Adding a scan_sum function allows to keep the code leaner and easier to read at the place of use. E.g.

```
comm.scan_sum(&my_vals,&my_sums,count);
```
instead of
```
MPI_Scan(&my_vals,&my_sums,count,MPI_DOUBLE,MPI_SUM,comm.mpi_comm());
```

## E3SM Stakeholder Feedback

This will be used in SCREAM

## Testing

Added simple mpi test checking Gauss formula for sum(1,...,n);


AaronDonahue:
It looks like the fails on Blake and Weaver have to do with openMPI slots, and just impact the comm test which was changed in this PR.  Is this a relatively simple issue to fix?

jeff-cohere:
It seems like the slot configuration is something that can be adjusted on the AT systems. Because this is an AUTOMERGE PR, I'll hold off on reviewing it till the tests pass.

jeff-cohere:
Thanks, @jgfouca ! I forgot about that option.

jgfouca:
@jeff-cohere , @AaronDonahue , @bartgol , I pushed a fix that should allow this PR to pass AT testing. The comm test raises a broader question of how to handle np>1 MPI tests. The issue is that our test infrastructure/scripts are mostly hard-coded to ask for a single node on batch systems. On many of these systems, if you ask for a single node but then try to run `mpiexec -np N` where N>1, the command will fail with an oversubscription error.

Potential solutions:
1) Ask batch systems for SCREAM|EKAT_TEST_MAX_RANKS nodes. This will make it harder for scream/ekat to get jobs running since we'd be asking for bigger allocations.
2) Allow for oversubscription on tests that want more that one node. That's the approach I took for the fix I made here. This should work fine for computationally cheap tests but could be a problem for tests that need a lot of memory/computation. Developers would need to remember to explicitly allow oversubscription for np>1 tests.
3) Always allow oversubscription, so just always add the oversubscribe flag to all MPI tests. We wouldn't need to think about this issue again until we add a test that oversubscribes the node too much.

PeterCaldwell:
It seems like for the time being allowing oversubscription is fine. When we eventually create tests that are too expensive for this option, it seems like the natural solution would be to create a new testing suite which consists only of tests which are expensive enough to require multiple nodes. That way most tests still get allocated quick and tests that need multiple nodes get what they need.

jgfouca:
Blake passed this time but weaver timed out. Retest should fix it.

jeff-cohere (approved):

bartgol:
This might be specific for ekat testing, since test-all-scream is a bit more structured. In particular, test-all-scream sets `SCREAM_TEST_MAX_RANKS` to whatever the python scripts determines it's the max cores/gpus available for the build. So, in that case, each test should always have enough mpi slots (*). The ekat testing is done manually with the good 'ol `cmake && make && ctest` chain from command line. We could probably adjust our jenkins scripts so that we do `ctest -jN` and `N*${EKAT_TEST_MAX_RANKS}` is less-than-or-equal to the number of cores on the testbed...

(*) We should probably think carefully about this for GPU build, since weaver has 4 gpus, and we currently have 4 builds in scream, which means we are not _really_ testing MPI-related features.

PR #109, opened 2021-05-17
add WSM constructor

tcclevenger:
Adds a new WSM constructor which takes in data pointer. This allows the user to per-reserve the memory used by the WSM. I also created a new function called `get_total_slots_to_be_used()` to calculate the total number of slots the WSM will require for a given set of constructor inputs, which will allow the creation of the WSM in a different scope than where the memory is reserved.

## Motivation
Memory for WSM needs to be reserved in SCREAM-AD in a different scope than the WSM is constructed. To see an example of this being used: https://github.com/E3SM-Project/scream/pull/1048

## Testing
Added a test for the new constructor and `get_total_slots_to_be_used()` function. Test asserts that data accessed in the WSM does not exceed the bounds of the view data it is constructed with.

### DIFF
--- a/src/ekat/ekat_workspace.hpp
+++ b/src/ekat/ekat_workspace.hpp
@@ -101,8 +101,8 @@ class WorkspaceManager
   // set of constructor inputs. Note, this does not actually create an instance of the WSM,
   // but is useful for when memory needs to be reserved in a different scope than the
   // WSM is created.
-  static size_t get_total_slots_to_be_used(int size, int max_used, TeamPolicy policy,
-                                           const double& overprov_factor=GPU_DEFAULT_OVERPROVISION_FACTOR);
+  static int get_total_bytes_needed(int size, int max_used, TeamPolicy policy,
+                                    const double& overprov_factor=GPU_DEFAULT_OVERPROVISION_FACTOR);
 
   // call from host.
   //
--- a/src/ekat/ekat_workspace.hpp
+++ b/src/ekat/ekat_workspace.hpp
@@ -266,9 +266,6 @@ class WorkspaceManager
  private:
 #endif
 
-  // Initialize int and view-type class variables.
-  void initialize_variables(const int size, const int max_used);
-
   friend struct unit_test::UnitWrap;
 
   template <typename S=T>
@@ -292,7 +289,9 @@ class WorkspaceManager
   KOKKOS_INLINE_FUNCTION
   void init_metadata(const int ws_idx, const int slot) const;
 
-  static void init(const WorkspaceManager& wm, const int max_ws_idx, const int max_used);
+  void init(const int max_ws_idx, const int max_used);
+
+  void compute_internals(const int size, const int max_used);
 
   //
   // data
--- a/src/ekat/ekat_workspace.hpp
+++ b/src/ekat/ekat_workspace.hpp
@@ -287,9 +287,9 @@ class WorkspaceManager
   Unmanaged<view_1d<S> > get_space_in_slot(const int team_idx, const int slot) const;
 
   KOKKOS_INLINE_FUNCTION
-  void init_metadata(const int ws_idx, const int slot) const;
+  void init_slot_metadata(const int ws_idx, const int slot) const;
 
-  void init(const int max_ws_idx, const int max_used);
+  void init_all_metadata(const int max_ws_idx, const int max_used);
 
   void compute_internals(const int size, const int max_used);
 
--- a/src/ekat/ekat_workspace.hpp
+++ b/src/ekat/ekat_workspace.hpp
@@ -87,6 +87,23 @@ class WorkspaceManager
   WorkspaceManager(int size, int max_used, TeamPolicy policy,
                    const double& overprov_factor=GPU_DEFAULT_OVERPROVISION_FACTOR);
 
+  // Constructor, call from host
+  //   Same as above, but here the user initializes the data.
+  //   This is useful when the user wants to pre-reserve the data
+  //   for the WorkspaceManager. The user is responsible for
+  //   ensuring that m_max_ws_idx*m_total*m_max_used contiguous
+  //   data is available, for which the get_total_slots_to_be_used()
+  //   function can be helpful.
+  WorkspaceManager(T* data, int size, int max_used, TeamPolicy policy,
+                   const double& overprov_factor=GPU_DEFAULT_OVERPROVISION_FACTOR);
+
+  // Helper functions which return the number of slots that will be reserved for a given
+  // set of constructor inputs. Note, this does not actually create an instance of the WSM,
+  // but is useful for when memory needs to be reserved in a different scope than the
+  // WSM is created.
+  static size_t get_total_slots_to_be_used(int size, int max_used, TeamPolicy policy,
+                                           const double& overprov_factor=GPU_DEFAULT_OVERPROVISION_FACTOR);
+
   // call from host.
   //
   // Will report usage statistics for your workspaces. These statistics will
--- a/src/ekat/ekat_workspace.hpp
+++ b/src/ekat/ekat_workspace.hpp
@@ -97,7 +97,7 @@ class WorkspaceManager
   WorkspaceManager(T* data, int size, int max_used, TeamPolicy policy,
                    const double& overprov_factor=GPU_DEFAULT_OVERPROVISION_FACTOR);
 
-  // Helper functions which return the number of slots that will be reserved for a given
+  // Helper functions which return the number of bytes that will be reserved for a given
   // set of constructor inputs. Note, this does not actually create an instance of the WSM,
   // but is useful for when memory needs to be reserved in a different scope than the
   // WSM is created.
@@ -265,6 +265,9 @@ class WorkspaceManager
 #ifndef KOKKOS_ENABLE_CUDA
  private:
 #endif
+
+  // Initialize int and view-type class variables.
+  void initialize_variables(const int size, const int max_used);
 
   friend struct unit_test::UnitWrap;
 
--- a/src/ekat/ekat_workspace_impl.hpp
+++ b/src/ekat/ekat_workspace_impl.hpp
@@ -38,6 +38,41 @@ WorkspaceManager<T, D>::WorkspaceManager(int size, int max_used, TeamPolicy poli
 }
 
 template <typename T, typename D>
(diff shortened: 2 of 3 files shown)


src/ekat/ekat_workspace_impl.hpp
```diff
@@ -37,6 +37,41 @@ WorkspaceManager<T, D>::WorkspaceManager(int size, int max_used, TeamPolicy poli
...
+  m_num_used("Workspace.m_num_used", m_max_ws_idx),
+  m_high_water("Workspace.m_high_water", m_max_ws_idx),
+  m_active("Workspace.m_active", m_max_ws_idx, m_max_used),
+  m_curr_names("Workspace.m_curr_names", m_max_ws_idx, m_max_used, m_max_name_len),
+  m_all_names("Workspace.m_all_names", m_max_ws_idx, m_max_names, m_max_name_len),
+  // A name's index in m_all_names is used to index into m_counts
+  m_counts("Workspace.m_counts", m_max_ws_idx, m_max_names, 2),
+#endif
+  m_next_slot("Workspace.m_next_slot", m_pad_factor*m_max_ws_idx),
+  m_data(data, m_max_ws_idx, m_total * m_max_used)
```
jgfouca, line 60: It looks like the two constructors are identical except for this line. I'm wondering if there's a way to use C++11's constructor delegation to get some code reuse here?

jgfouca (changes requested):

src/ekat/ekat_workspace_impl.hpp
bartgol: We can have a private ctor that inits everything except `m_data`, and the other two call it. Since that ctor would be private, nobody would be able to call it (avoiding creating a WSM with uninited `m_data`).

tcclevenger: I agree, but I'm running into the issue the private constructor would have the exact same inputs as the original (non-data) constructor. Not sure how to get around this..

src/ekat/ekat_workspace.hpp
```diff
@@ -87,6 +87,23 @@ class WorkspaceManager
...
+  //   data is available, for which the get_total_slots_to_be_used()
+  //   function can be helpful.
+  WorkspaceManager(T* data, int size, int max_used, TeamPolicy policy,
+                   const double& overprov_factor=GPU_DEFAULT_OVERPROVISION_FACTOR);
+
+  // Helper functions which return the number of slots that will be reserved for a given
+  // set of constructor inputs. Note, this does not actually create an instance of the WSM,
+  // but is useful for when memory needs to be reserved in a different scope than the
+  // WSM is created.
+  static size_t get_total_slots_to_be_used(int size, int max_used, TeamPolicy policy,
```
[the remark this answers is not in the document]
tcclevenger: I now return the memory in bytes where I calculate the slots and return `n_slots*sizeof(T)`.

jgfouca (approved):

jgfouca:
Removing my modification request so that this PR is not blocked while I'm on vacation.

src/ekat/ekat_workspace.hpp
```diff
@@ -87,6 +87,23 @@ class WorkspaceManager
...
+  //   Same as above, but here the user initializes the data.
+  //   This is useful when the user wants to pre-reserve the data
+  //   for the WorkspaceManager. The user is responsible for
+  //   ensuring that m_max_ws_idx*m_total*m_max_used contiguous
+  //   data is available, for which the get_total_slots_to_be_used()
+  //   function can be helpful.
+  WorkspaceManager(T* data, int size, int max_used, TeamPolicy policy,
+                   const double& overprov_factor=GPU_DEFAULT_OVERPROVISION_FACTOR);
+
+  // Helper functions which return the number of slots that will be reserved for a given
```
bartgol, line 100: I think the comment needs to be modified (slots -> bytes).

src/ekat/ekat_workspace_impl.hpp
bartgol: Maybe a setup method? You'd have to give up initialization lists, and use assignments inside the setup method, but that's just a cosmetic difference in this case.

tcclevenger:
@bartgol Let me know what you think. I'm starting to 2nd guess the name `initialize_variables()` since there is also an `init()` functions. Sadly they can't be combined since `init()` need  `m_data` to be initialized, and `m_data` initialization need to follow `initialize_variables()`. Maybe `setup_variables()`?

bartgol:
> @bartgol Let me know what you think. I'm starting to 2nd guess the name `initialize_variables()` since there is also an `init()` functions. Sadly they can't be combined since `init()` need `m_data` to be initialized, and `m_data` initialization need to follow `initialize_variables()`. Maybe `setup_variables()`?

It seems to me that the `init` method is never called externally (even in scream). So maybe we could a) make it private, b) make it not static, and remove the WSM pointer input (which is really just manually doing what a non-static method does, btw), c) rename it something like `init_metadata`, which is more appropriate probably. Then, the `initialize_variables` can be called simply `init` or, if we want to avoid confusion (since the other init method is `init_metadata`), something like `init_counters` or `compute_internal_sizes` or `compute_internals`.

tcclevenger:
> It seems to me that the `init` method is never called externally (even in scream). So maybe we could a) make it private, b) make it not static, and remove the WSM pointer input (which is really just manually doing what a non-static method does, btw), c) rename it something like `init_metadata`, which is more appropriate probably. Then, the `initialize_variables` can be called simply `init` or, if we want to avoid confusion (since the other init method is `init_metadata`), something like `init_counters` or `compute_internal_sizes` or `compute_internals`.

Turns out `init` is already private. I'm removing `static` and the WSM pointer, then renaming my function to `compute_internals` (since it computes sizes and initializes some views). There is already a function called `init_metadata` which `init` calls, but is also called independently in various places for debug runs, so that name wouldn't work. We could just leave it as `init`?

bartgol:
> > It seems to me that the `init` method is never called externally (even in scream). So maybe we could a) make it private, b) make it not static, and remove the WSM pointer input (which is really just manually doing what a non-static method does, btw), c) rename it something like `init_metadata`, which is more appropriate probably. Then, the `initialize_variables` can be called simply `init` or, if we want to avoid confusion (since the other init method is `init_metadata`), something like `init_counters` or `compute_internal_sizes` or `compute_internals`.
> 
> Turns out `init` is already private. I'm removing `static` and the WSM pointer, then renaming my function to `compute_internals` (since it computes sizes and initializes some views). There is already a function called `init_metadata` which `init` calls, but is also called independently in various places for debug runs, so that name wouldn't work. We could just leave it as `init`?

Ah, I missed the other `init_metadata`. Too bad they have the same args, so we can't even overload it. An alternative would be `init_all_metadata` vs `init_slot_metadata`, with the former calling the latter. But if you don't feel like it, we can keep it as is.

tcclevenger:
> Ah, I missed the other `init_metadata`. Too bad they have the same args, so we can't even overload it. An alternative would be `init_all_metadata` vs `init_slot_metadata`, with the former calling the latter. But if you don't feel like it, we can keep it as is.

I think I like that naming. I'll change it. Seems better than having an `init` function which doesn't initialize everything

tcclevenger:
@bartgol Let me know if you have any other comments.

tests/kokkos/workspace_tests.cpp
```diff
@@ -98,6 +98,37 @@ static void unittest_workspace()
...
+    // Test constructing the WSM using view data
+    size_t n_bytes = WorkspaceManager<double, Device>::get_total_bytes_needed(17, num_ws, policy);
+    size_t n_slots = n_bytes/sizeof(double);
+    view_1d<double> tmp_view("", n_slots);
+
+    // Calculate next available location
+    double* data_end = tmp_view.data();
+    data_end += tmp_view.size();
+
+    WorkspaceManager<double, Device> wsm(tmp_view.data(), 17, num_ws, policy);
```
bartgol, line 111: Maybe it's me, but I am getting a bit confused by the var names here. So, 17 is the size of each slot (as number of double's) in each workspace, right? Then num_ws sounds like 'num_workspaces', but it is the number of (17-doubles-long) slots needed in each workspace. And n_slots is not really the number of slots, but rather the number of doubles needed (which is (17+reserve)*num_workspaces*num_slots_per_ws), right?

bartgol: I think there are some things in the WSM stack that are not clear. E.g., `get_space_in_slot`, makes it sound like there are more than 1 space in a slot. But I _think_ the returned space _is a slot_.

tcclevenger: You are correct on all counts. I just pushed some new names.

tests/kokkos/workspace_tests.cpp
```diff
@@ -87,31 +87,33 @@ static void unittest_workspace()
...
+  static constexpr const int num_workspaces = 4;
   const int ni = 128;
   const int nk = 128;
 
   TeamPolicy policy(ExeSpaceUtils<ExeSpace>::get_default_team_policy(ni, nk));
 
   {
-    WorkspaceManager<double, Device> wsmd(17, num_ws, policy);
+    const int slot_length = 17;
+    WorkspaceManager<double, Device> wsmd(slot_length, num_workspaces, policy);
```
bartgol, line 98: Sorry for being pedantic, but is the 2nd ctor arg the number of workspaces or rather the number of slots in a workspace? I thoguht the latter, since the number of workspaces depends on the number of concurrent teams, which is deduced from the policy...

tcclevenger: I see what you are saying, yeah it should be `n_slots`, each with `slot_length=17`. Maybe the original `num_ws` stood for "Number of Workspace Slots"?

tcclevenger: Should it be something like `num_slots_per_team`?

bartgol: Yeah, I think so. Or, if you want to stick with WS-names, something like `num_slots_per_ws`, or something.

tcclevenger:
@bartgol I changed some of the names in the workspace test. I also grouped the detailed take-release testing at the end in its own scope so that naming could be consistent with the earlier test. Unfortunately it makes the diff look awful, but nothing about the actual test changed, just names.

bartgol (approved):

PR #112, opened 2021-05-21
changed Pack copy ctor to allow different scalar types for mixed-precision

### LINKED ISSUE #111: Mixed-precision packs
Some ill-conditioned algorithms require double precision, even in single precision configurations.  

We'd like to be able to copy-construct `ekat::Pack` types of the same size but possibly different scalar types.

pbosler:
This PR changes the `ekat::Pack` copy constructor to allow copying between packs of the same size, but different scalar types.  It adds a test that copies between packs of `double` and packs of `float` type.

## Motivation

Some ill-conditioned algorithms require double precision arithemetic.  This PR allows client codes that are configured for single precision to convert to double precision upon entry to such an algorithm, perform the computations in double precision, then convert pack to single precision to return.

## Related Issues

Closes #111

## Testing

Added an associated test function to the pack unit tests.

### DIFF
--- a/src/ekat/ekat_pack.hpp
+++ b/src/ekat/ekat_pack.hpp
@@ -218,7 +218,7 @@ struct Pack {
     }
   }
 
-  // Init this Pack form two scalars, according to a given mask:
+  // Init this Pack from two scalars, according to a given mask:
   // if mask is true, set first value, otherwise the other.
   template <typename T, typename S>
   KOKKOS_FORCEINLINE_FUNCTION
--- a/src/ekat/ekat_pack.hpp
+++ b/src/ekat/ekat_pack.hpp
@@ -133,7 +133,7 @@ Mask<n> operator ! (const Mask<n>& m) {
 // input is a Pack; _s means the input is a scalar.
 // NOTE: for volatile overload, you should return void. If not, if/when Kokkos
 //       atomic ops are compiled for Pack, you will see a compiler warning like
-//   warning: implicit dereference will not access object of type ‘volatile ekat::Pack<...>' in statement 
+//   warning: implicit dereference will not access object of type ‘volatile ekat::Pack<...>' in statement
 //       *dest = return_val + val;
 //       ^
 #define ekat_pack_gen_assign_op_p(op)                       \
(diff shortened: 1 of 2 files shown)


src/ekat/ekat_pack.hpp
```diff
@@ -218,7 +218,7 @@ struct Pack {
     }
   }
 
-  // Init this Pack from two scalars, according to a given mask:
+  // Init this Pack form two scalars, according to a given mask:
```
jeff-cohere, line 221: Introduced a typo here

tests/pack/pack_tests.cpp
```diff
@@ -195,11 +195,24 @@ struct TestPack {
...
 
+  static void test_mixed_prec () {
+    const ekat::Pack<float, PACKN> ones(1.0);
+    const ekat::Pack<double, PACKN> twos(2.0);
+
+    const ekat::Pack<double, PACKN> dones(ones);
+    const ekat::Pack<float, PACKN> ftwos(twos);
+
+    vector_novec for (int i=0; i<PACKN; ++i) {
+      REQUIRE( std::abs(dones[i] - 1.0) < 1e-7 );
```
jeff-cohere, line 206: Do you want a lower tolerance for the `double` version to illustrate what's going on here?

jeff-cohere (changes requested):
Nice! Suggesting some very minor changes.

jeff-cohere (approved):

src/ekat/ekat_pack.hpp
```diff
@@ -189,13 +189,6 @@ struct Pack {
     vector_simd for (int i = 0; i < n; ++i) d[i] = v;
   }
 
-  // Init this Pack from another one.
-  template <typename T>
-  KOKKOS_FORCEINLINE_FUNCTION explicit
-  Pack (const Pack<T,n>& v) {
```
bartgol, line 195: Was this copy ctor not doing what you needed? That seems weird. IIt seems the new ctor only adds an explicit call to the scalar type ctor. But the builtin type double is implicitly constructible from float, so `d[i] = v[i]` should work when the src has a scalar type `float` and the current obj has scalar type `double`.

bartgol (changes requested):
Sorry, I'm not requesting changes, but I want to block the PR until I understand if/why it's needed (i.e., if/why the current code was not working).

bartgol:
This works as expected for me:
```
  ekat::Pack<float,1> p1 (1.0/3.0);
  ekat::Pack<double,1> p2(p1);
  ekat::Pack<double,1> p3 (1.0/3.0);
  ekat::Pack<float,1> p4 (p3);

  std::cout << std::setprecision(18) << "p1[0] = " << p1[0] << "\n";
  std::cout << std::setprecision(18) << "p2[0] = " << p2[0] << "\n";
  std::cout << std::setprecision(18) << "p3[0] = " << p3[0] << "\n";
  std::cout << std::setprecision(18) << "p4[0] = " << p4[0] << "\n";
```
Output:
```
p1[0] = 0.333333343267440796
p2[0] = 0.333333343267440796
p3[0] = 0.333333333333333315
p4[0] = 0.333333343267440796
```

pbosler:
And it didn't fix the error I'm getting.   Good eye, Luca.

PR #113, opened 2021-05-21
Add MemoryTraits to template signature of view aliases

bartgol:
## Motivation

If you don't specify MemoryTraits in a view template signature, the resulting type is different (and therefore incompatible) with a type that has it declared. E.g., with
```
using V1 = Kokkos::View<int*,Kokkos::MemoryManaged>;
using V2 = Kokkos::View<int*>;
```
an object of type V2 cannot be bound to a reference V2&. This makes writing functions that take in, say, view_1d<Real>, more complicated. E.g., the two functions
```
void f1 (const view_1d<Real>& v);
void f2 (const ekat::Unmanaged<view_1d<Real>>& v);
```
will both fail to compile if passed either a V2 or V1 respectively. Adding MemoryTraits to the templated signature of view_1d makes things easier:

```
template<typename MT>
void f1 (const view_1d<Real,MT>& v);

view_2d<Real> a("",10,10
view_1d<Real> b("",10);
auto c = subivew(a,0); // this view has MemoryTraits=Kokkos::MemoryUnmanaged

f1(b); // works
f1(c); // works
```

NOTE: of course, in order to keep things simple, and not to break existing code, the MemoryTraits argument of views aliases in KokkosTypes is defaulted to Kokkos::MemoryManaged (which is what kokkos still picks under the hood if no memory traits are specified).

## E3SM Stakeholder Feedback

Needed in SCREAM.

## Testing

All current tests pass. Views with non-defaulted MemoryTraits working (tested in scream).

### DIFF
--- a/src/ekat/kokkos/ekat_kokkos_types.hpp
+++ b/src/ekat/kokkos/ekat_kokkos_types.hpp
@@ -4,6 +4,7 @@
 #include "ekat/ekat_config.h"
 #include "ekat/kokkos/ekat_kokkos_meta.hpp"
 #include "ekat/std_meta/ekat_std_type_traits.hpp"
+#include <Kokkos_MemoryTraits.hpp>
 
 /*
  * Header contains globally useful kokkos-related types for ekat.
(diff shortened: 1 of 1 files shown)


jeff-cohere (approved):
lgtm

PR #115, opened 2021-05-29
Add a `Logger` to EKAT

pbosler:
# Adds logging capability using the `spdlog` library

The goal of this PR is to supply an EKAT-unified capability for logging output from programs.  The `Logger` class it introduces uses a policy-based design, where the policies determine its behavior with respect to file output (none, or one file per rank), its console logging (from all ranks, or only rank 0), and the ability to have multiple logs (e.,g., for driver, io, and other individual components) share the same output file; their messages are tagged with their individual names.

Each `Logger` instance can have up to 2 output "sinks," one for the console, and one for the (optional) file.  Each sink has a priority level; messages with priority below the sink priority level will not be output through the sink.  Independently, a log can have its own priority level.  Messages with priority below the log priority level will not be sent to any of its sinks.

The priority levels are (increasing priority):
  - off
  - trace
  - debug
  - info
  - warn
  - err
  - critical

The main use-case I envision is Scream.  The driver could initialize a main log, then each subcomponent (Hommexx, Haero) could initialize their own logs that (optionally) share the same log files so that program flow is clear from a single file (per mpi rank).  An example of this is provided in `ekat/tests/logging/mpi_logger_tests.cpp`.

## This PR includes

1. CMake functions to auto-build the new `spdlog` submodule alongside `yaml-cpp` and `kokkos` TPLs.
2. The `spdlog` default logger (easiest to use) is exposed via namespace `Log`, and tested.
3. Customized loggers (as desribed above) are also provided, and tested.
4. Tests that serve as examples for the most common use-cases.
5. Support for logging `ekat::Pack` types as easily as POD types.

## Why `spdlog`?

For the sake of not reinventing the wheel, I checked out the top GitHub repos with tags "c++" and "logging."  I looked for permissive, non-viral licenses, thread safety and asynchronous logging capability, attention to performance, lightweight (header only, in most cases, to match the ekat precedent of Catch2) standalone (not boost) logging libraries.  

After a lot of browsing, I looked in detail at two options.   They both provide easy formatting using the python formatting syntax and they both use the MIT license.
- https://github.com/odygrd/quill 
  - Pros: 
    - Fast.  Github page compares its single- and multi-threaded performance to many other loggers. 
    - Actively developed with recent GitHub PRs, releases and updated.
  - Cons: 
    - Requires c++14.  
    - Not clear from the docs how to adapt it to most common EKAT use cases.
-  https://github.com/gabime/spdlog
  - Pros: 
    - Top GitHub result with > 11k stars, > 200 contributors and active development.
    - Excellent documentation in-source and on the GitHub wiki; easy to adapt to my impression of E3SM logging practices.
    - Can be used header-only or precompiled (to reduce build times).
    - C++ 11 ok

Ultimately, since I'm not sure when c++14 will make it to EKAT and because of the nice documentation in spdlog, I chose spdlog for a first implementation.  

## Testing

Basic examples in `ekat/tests/logging/logger_tests.cpp`.
MPI and multi-component examples in `ekat/tests/logging/mpi_logger_tests.cpp`.


src/ekat/ekat_pack.hpp
```diff
@@ -10,6 +10,7 @@
 #include "ekat/ekat_type_traits.hpp"
 
 #include <Kokkos_Core.hpp>
+#include "spdlog/fmt/ostr.h"
```
jeff-cohere, line 13: Is this added to `ekat_pack.hpp` for convenience? To make logging more seamless when using packs?

tests/logging/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+include(EkatCreateUnitTest)
+
+# Test units framework
+EkatCreateUnitTest(logger logger_tests.cpp LIBS ekat)
+
+EkatCreateUnitTest(mpilogger mpi_logger_tests.cpp
+  LIBS ekat
+  MPI_RANKS 1 ${EKAT_TEST_MAX_RANKS}
```
jeff-cohere, line 8: Do you think it's worth adding a 2-rank case to make sure logging works with > 1 process?

src/ekat/logging/ekat_log_mpi.hpp
```diff
@@ -0,0 +1,66 @@
...
+#include "ekat/mpi/ekat_comm.hpp"
+#include "spdlog/sinks/null_sink.h"
+#include <mpi.h>
+
+// This file contains policy classes for choosing the behavior of the ekat
+// logger with respect to mpi ranks.
+// Each policy must implement the
+//
+// static void update_sinks(std::vector<spdlog::sink_ptr>& sinks)
+// static std::string name_append_rank(const std::string& name)
```
jeff-cohere, line 13: Maybe consider `name_with_rank`, since the function returns the logger's name with the rank appended.

jeff-cohere: Or possible `append_rank`, if you want to be "verby" in the context of this policy.

src/ekat/logging/ekat_log_file.hpp
```diff
@@ -0,0 +1,74 @@
...
+  static std::shared_ptr<spdlog::sinks::basic_file_sink_mt> get_file_sink(
+    const std::string& logfilename) {
+    auto result = std::make_shared<spdlog::sinks::basic_file_sink_mt>(logfilename, true);
+    result->set_level(FileLogLevel);
+    return result;
+  }
+};
+
+
+struct LogSharedFile {
```
jeff-cohere, line 45: Can you add a comment? It's intuitively obvious that this logger shares a file with others, but it's not obvious how it works, so one might not be sure what to make of this at first glance.

jeff-cohere (commented):
This is really cool, Pete. Aside from minor doc-related comments, I wonder if we should try to capture the output from the loggers in their unit tests (either using regular expressions in CMake or by intercepting file output). This would check the logger's ability to send the right output to the right places, but it might be a little involved.

src/ekat/ekat_pack.hpp
pbosler: Yes.  The spdlog docs say you have to include that file wherever you define a type's `operator <<` to get native support from spdlogs functions.

tests/logging/CMakeLists.txt
pbosler: I believe EKAT_TEST_MAX_RANKS defaults to 4

pbosler:
> This is really cool, Pete. Aside from minor doc-related comments, I wonder if we should try to capture the output from the loggers in their unit tests (either using regular expressions in CMake or by intercepting file output). This would check the logger's ability to send the right output to the right places, but it might be a little involved.

I was thinking that, too.  Both that it would be good to do and that it would take a lot of CMake-ing.

tests/logging/CMakeLists.txt
jeff-cohere: Ah, my eyes sailed right past that. Thanks.

bartgol:
Thanks for putting a lot of effort into this, Pete! I especially like how you worked us through how you got to pick the spdlog library.

> Ultimately, since I'm not sure when c++14 will make it to EKAT and because of the nice documentation in spdlog, I chose spdlog for a first implementation.

EKAT _is_ using c++14. The switch came when we last updated our Kokkos version to 3.3, which defaults c++ to 14. I am not 100% sure if c++14 is _required_ by Kokkos 3.3 or if it is simply the default one. I suspect the former, but I haven't tried. Either way, EKAT (and SCREAM) are actively using C++14.

That's not to say that we need to use that std, of course. Libraries in c++11 can be just as good as newer ones.

I will review this PR in detail today.

src/ekat/logging/ekat_logger.hpp
```diff
@@ -0,0 +1,158 @@
...
+    std::string result;
+    auto fsink = this->get_file_sink();
+    spdlog::sinks::basic_file_sink_mt* bptr =
+      dynamic_cast<spdlog::sinks::basic_file_sink_mt*>(fsink.get());
+    spdlog::sinks::rotating_file_sink_mt* rptr =
+      dynamic_cast<spdlog::sinks::rotating_file_sink_mt*>(fsink.get());
+    if (bptr) {
+      result = bptr->filename();
+    }
+    else if (rptr) {
```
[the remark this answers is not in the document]
bartgol: You might do the second cast in here, and maybe throw if the second cast fails (since we expect a file sink to be either basic or rotatin, right?).

src/ekat/logging/ekat_logger.hpp
```diff
@@ -0,0 +1,158 @@
...
+    return result;
+  }
+
+  template <typename FP = LogFilePolicy>
+  typename std::enable_if<!FP::has_filename, std::string>::type
+  get_logfile_name() {return "null";}
+};
+
+template <typename LogMpiPolicy>
+class Logger<LogSharedFile, LogMpiPolicy> : public spdlog::logger {
```
bartgol, line 109: This seems like a lot of code duplication, mostly just to eliminate the creation of the file sink. My suggestion would be to add a method to the main Logger class called something like `share(name,level)` or something like that, that returns a `Logger<LogSharedFile,LogMpiPolicy` pointer sharing the sink with `*this`. This method creates a logger internally, and sets the csink/fsink in it (can they also share the csink, if the clevel is the same?).

Note: this would require the ability to set the sinks, but I think that's relatively easy and short:
- turn `Logger()=delete;` into `Logger() = default`, and make it private;
- make `template<typename FP,typename MP> Logger<FP,MP>` a friend class (so that `*this` can access the copied log inside the `share(name,level)` method to set the sinks.

Alternatively to the previous 2 items, add a ctor to the main Logger class that takes a csink and a fsink, and directly sets them internally, no queestion asked, and use that in the `share(name,level)` method.

src/ekat/logging/ekat_logger.hpp
```diff
@@ -0,0 +1,158 @@
...
+
+    std::string fname = LogMpiPolicy::name_append_rank(log_name) + "_logfile.txt";
+    auto fsink = LogFilePolicy::get_file_sink(fname);
+
+    this->sinks().push_back(csink);
+    this->sinks().push_back(fsink);
+
+    LogMpiPolicy::update_sinks(this->sinks());
+
+    this->set_level(clevel);
```
bartgol, line 74: Question (on spdlog): I'm a bit confused by the fact that we can set a log level on both the logger and the sink(s). I get that different sinks may have different log levels, so that's fine, but why also a level on the Logger? Is the idea that the Logger defines some sort of "general" level, and then each sink can have its own "additional" way of sift logging?

src/ekat/logging/ekat_logger.hpp
```diff
@@ -0,0 +1,158 @@
...
+  public:
+    Logger() = delete;
+
+    Logger(const std::string& log_name, const std::string& level_str="debug") :
+     spdlog::logger(LogMpiPolicy::name_append_rank(log_name))
+  {
+
+    const auto clevel = Log::level::from_str(level_str);
+    auto csink = std::make_shared<spdlog::sinks::stdout_color_sink_mt>();
+    csink->set_level(clevel);
```
bartgol, line 64: Why don't we set a log level on the fsink, like we do for the csink?

src/ekat/logging/ekat_logger.hpp
```diff
@@ -0,0 +1,158 @@
...
+  features of that library to the ones EKAT is most likely to use, while retaining all
+  of its functionality.
+*/
+template <typename LogFilePolicy = LogNoFile,    // see ekat_log_file.hpp
+          typename LogMpiPolicy = LogIgnoreRank> // see ekat_log_mpi.hpp
+class Logger : public spdlog::logger {
+  public:
+    Logger() = delete;
+
+    Logger(const std::string& log_name, const std::string& level_str="debug") :
```
bartgol, line 58: I think strings may be subject to typos that fly under the radar. I would use an Enum for the log level (doesn't have to be a strong enum, since we're already inside the logger namespace).

tests/logging/mpi_logger_tests.cpp
```diff
@@ -0,0 +1,97 @@
...
+    REQUIRE(main_log.get_logfile_name() == component_log.get_logfile_name());
+
+    // check that the main log will not output debug messages
+    main_log.debug("this will not be logged from main_log.");
+    REQUIRE( !main_log.should_log(Log::level::debug) );
+
+    // check that the component log will output debug messages
+    component_log.debug("this will be logged from component_log.");
+    REQUIRE( component_log.should_log(Log::level::debug) );
+
```
bartgol, line 92: A bunch of empty lines that can be removed.

bartgol (changes requested):
Great work Pete! I have some questions and some suggestions.

Regarding testing the log files output, we could complete the logging, destroy the logger (e.g., goinig out of scope), since I believe that's when the sink file is closed, and then re-open the file with normal ifstream objects, and inspect the content.

pbosler:
> EKAT _is_ using c++14. 

Good to know!  I remember talking about it but I missed the decision.  We can keep `quill` around as a plan b if `spdlog` shows some undesirable properties as we move forward.

src/ekat/logging/ekat_logger.hpp
```diff
@@ -0,0 +1,158 @@
...
+     spdlog::logger(LogMpiPolicy::name_append_rank(log_name))
+  {
+
+    const auto clevel = Log::level::from_str(level_str);
+    auto csink = std::make_shared<spdlog::sinks::stdout_color_sink_mt>();
+    csink->set_level(clevel);
+
+    std::string fname = LogMpiPolicy::name_append_rank(log_name) + "_logfile.txt";
```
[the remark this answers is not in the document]
pbosler: This is a good question.   I know concurrent logging is ok between threads, but I'm not sure about between ranks.   I will do some tests.

src/ekat/logging/ekat_logger.hpp
```diff
@@ -0,0 +1,158 @@
...
+    this->sinks().push_back(csink);
+    this->sinks().push_back(fsink);
+
+    LogMpiPolicy::update_sinks(this->sinks());
+
+    this->set_level(clevel);
+  }
+
+  spdlog::sink_ptr get_console_sink() {
+    return (this->sinks().size() > 1 ? this->sinks()[0] : spdlog::sink_ptr(nullptr));
```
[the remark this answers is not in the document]
pbosler: See previous comment.  A log always has a file sink.   It may or may not also have a console sink.

src/ekat/logging/ekat_log_mpi.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+/* Suppress console output from ranks > 0.  All ranks still write to files, if
+  file logging is enabled.
+*/
+struct LogOnlyRank0 {
+  // erase the console sink for all ranks != 0
+  static void update_sinks(std::vector<spdlog::sink_ptr>& sinks) {
+    const Comm mpicomm(MPI_COMM_WORLD);
+    if (!mpicomm.am_i_root()) {
+      sinks.erase(sinks.begin());
```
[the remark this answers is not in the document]
pbosler: This is a good option that could get around my hack for the problem, mentioned above, that I could not assign a null_sink shared pointer to what had been a console shared_ptr, even though they both inherit from `spdlog::sink_ptr`.

src/ekat/logging/ekat_logger.hpp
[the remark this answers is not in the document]
pbosler: Yeah, it would be nice if null_sink returned an empty string for its filename method, same for console sink.

src/ekat/logging/ekat_logger.hpp
pbosler: This confused me too, at first.  I had to write some extra tests to verify I understood the behavior.   It's an extra layer of flexibility that has some nice consequences.  In pseudocode, the logic flows like this:

```
def log_message(msg):
  if msg.priority >= log.priority:
    sink_msg(msg)
  else:
    return

def sink_msg(msg):
  if msg.priority >= sink.priority
    write_msg(msg)
  else:
   return
```

In hindsight, that pseudocode is probably obvious.  But here's how it becomes helpful.  Let's say Hommexx and Haero each have a logger, but they share the same sinks (console and file).   Since Hommexx is a fancy SC20 code, we have confidence in it and we don't  need to see all of its debug and trace output, so we set the hommexx logger level to `warn`.  Haero is just getting started though and it's largely unproven, so we set the Haero logger level to `debug`.   This means that the log files will only show warning or higher priority messages from Hommexx, but they'll show debug output from haero, all in the same sinks.

 If the file sink were the same level as the logger, then either all the haero debug output would be surpassed by the higher logging level of hommexx, or we'd see all of hommexx's debug output.

Btw, the test code I wrote to verify this is in `logger_tests.cpp`, lines 53--55:

```
      // the file level is trace, but the log level is debug (debug > trace); trace messages will be skipped.
      mylog.trace("this message won't show up anywhere.");
      REQUIRE( !mylog.should_log(Log::level::trace) );
```

src/ekat/logging/ekat_logger.hpp
pbosler: The file sink priority is set via the FilePolicy template class.  If you'd like a different design, I can rearrange.  I've been rereading Alexandrescu's book recently, to improve my meta programming skills -- it's possible I went a little too far in trying to follow his policy-based design suggestions.

src/ekat/logging/ekat_logger.hpp
bartgol: This makes sense. Thanks for the detailed explanation!

src/ekat/logging/ekat_logger.hpp
bartgol: > The file sink priority is set via the FilePolicy template class. If you'd like a different design, I can rearrange.

Ah, I missed that. This is fine, no need to change.

bartgol: > I've been rereading Alexandrescu's book recently, to improve my meta programming skills -- it's possible I went a little too far in trying to follow his policy-based design suggestions.

I was indeed wondering if you had been reading his stuff, given how you went the direction of policy design. :)

src/ekat/logging/ekat_logger.hpp
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  spdlog::sink_ptr get_file_sink() {
+    return (this->sinks().size() > 1 ? this->sinks()[1] : this->sinks()[0]);
```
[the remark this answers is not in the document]
bartgol: Ok, I think I understand it now. My suggestion would be to not delete the csink then, which causes this involved logic here. Getting the null_sink_mt to work would be the best option, but if that doesn't work, what about setting the csink;s log_level to off?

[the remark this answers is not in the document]
bartgol: Ah, sorry, I just noticed this option is discussed in another comment.

src/ekat/logging/ekat_log_mpi.hpp
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+
+  // append rank id to the log name
+  static std::string name_append_rank(const std::string& name) {
+    const Comm mpicomm(MPI_COMM_WORLD);
```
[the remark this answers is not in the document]
bartgol: Here's another point in favor of not using MPI_COMM_WORLD. Say we are running scream, using a subcomm of the world comm, with the rest doing, I don't know, land, or other stuff. When using the LogOnlyRank0 policy, we probably want one rank _in scream_ to log the output, but no process in scream might correspond to rank 0 in the world comm.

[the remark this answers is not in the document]
pbosler: I'll work on this.

src/ekat/logging/ekat_logger.hpp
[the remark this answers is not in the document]
pbosler: > You might do the second cast in here, and maybe throw if the second cast fails (since we expect a file sink to be either basic or rotatin, right?)

The file sink can be `spdlog::null_sink_mt`, which does not have a `filename()` method, so it would be incorrect to throw.  Maybe it's worth asking the `spdlog` folks to add a `filename()` method to `null_sink_mt` that just returns an empty string.

src/ekat/logging/ekat_logger.hpp
pbosler: > This seems like a lot of code duplication, mostly just to eliminate the creation of the file sink.

You're right. Thanks for this suggestion.  I was able to reproduce the behavior by just adding another constructor (rather than a whole specialized class).

src/ekat/logging/ekat_logger.hpp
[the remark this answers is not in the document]
pbosler: I agree with your recommendation to delete `LogIgnoreRank`.  I had put it in only for the case of running without MPI, but (a) I didn't make it impossible to use with MPI and (b) since EKAT already requires MPI, it's unnecessary.    Deleting this policy will fix the problem of multiple ranks writing to the same file.

pbosler:
Remaining to-dos:

- [ ] Pass an `ekat::Comm` instance to the MPI policy for cases where MPI_COMM_WORLD is incorrect.
- [ ] Improve the `LogOnlyRank0` hack for removing the console sink; either figure out how to get `null_sink_mt` to work (preferred) or set the console sink's level to `off`.

src/ekat/logging/ekat_logger.hpp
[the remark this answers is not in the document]
bartgol: Ah, good point.

src/ekat/logging/ekat_log_file.hpp
```diff
@@ -0,0 +1,74 @@
...
+namespace logger {
+
+#define EKAT_LOG_N_FILES 5
+#define EKAT_LOG_MAX_FILE_SIZE_MB 8
+
+// No file output; only console logging.
+struct LogNoFile {
+  static constexpr bool has_filename = false;
+  static std::shared_ptr<spdlog::sinks::null_sink_mt> get_file_sink(const std::string& logfilename="") {
+  return std::make_shared<spdlog::sinks::null_sink_mt>();}
```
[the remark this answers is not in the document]
bartgol: I meant the intentation of the return statement.

src/ekat/logging/ekat_log_mpi.hpp
```diff
@@ -15,35 +15,20 @@
...
 struct LogOnlyRank0 {
-  // erase the console sink for all ranks != 0
-  static void update_sinks(std::vector<spdlog::sink_ptr>& sinks) {
-    const Comm mpicomm(MPI_COMM_WORLD);
+  // nullify the console sink for all ranks != 0
+  static void update_sinks(std::vector<spdlog::sink_ptr>& sinks, const ekat::Comm& mpicomm=ekat::Comm()) {
     if (!mpicomm.am_i_root()) {
-      sinks.erase(sinks.begin());
+      //sinks.erase(sinks.begin()); // this is a hack, but it works
+      sinks[0] = std::make_shared<spdlog::sinks::null_sink_mt>();
```
bartgol, line 26: Nice. Now we can rely on the fact that each Logger has 2 sinks.

bartgol: Also, we can remove the commented line.

src/ekat/logging/ekat_log_mpi.hpp
```diff
@@ -15,35 +15,20 @@
...
-
-
 /* Suppress console output from ranks > 0.  All ranks still write to files, if
   file logging is enabled.
 */
 struct LogOnlyRank0 {
-  // erase the console sink for all ranks != 0
-  static void update_sinks(std::vector<spdlog::sink_ptr>& sinks) {
-    const Comm mpicomm(MPI_COMM_WORLD);
+  // nullify the console sink for all ranks != 0
```
bartgol, line 22: Question: should we nullify also the file sink? If the MPI policy is different for C and F sinks, then we should have a policy for each.

src/ekat/logging/ekat_log_mpi.hpp
```diff
@@ -15,35 +15,20 @@
...
+  static void update_sinks(std::vector<spdlog::sink_ptr>& sinks, const ekat::Comm& mpicomm=ekat::Comm()) {
```
bartgol, line 23: Related to the above: if we only deal with one sink, I think it'd be best to pass a single sink ptr to this routine, and let the Logger class distinguish the concept of file and console sinks.

src/ekat/logging/ekat_logger.hpp
```diff
@@ -50,36 +50,57 @@ namespace Log = spdlog;
...
 
-    std::string fname = LogMpiPolicy::name_with_rank(log_name) + "_logfile.txt";
+    std::string fname = LogMpiPolicy::name_with_rank(log_name, mpicomm) + "_logfile.txt";
     auto fsink = LogFilePolicy::get_file_sink(fname);
 
     this->sinks().push_back(csink);
     this->sinks().push_back(fsink);
 
-    LogMpiPolicy::update_sinks(this->sinks());
+    LogMpiPolicy::update_sinks(this->sinks(), mpicomm);
```
bartgol, line 71: See comment above: we should detach the concept of MPI policy from that of C vs F sinks. The Logger should be aware of C vs F, but the policy should only deal with a generic sink. So we should do
```
LogMpiPolicy::update_sink(get_console_sink(), mpicomm);
```

src/ekat/logging/ekat_logger.hpp
```diff
@@ -50,36 +50,57 @@ namespace Log = spdlog;
...
-          typename LogMpiPolicy = LogIgnoreRank> // see ekat_log_mpi.hpp
+          typename LogMpiPolicy = LogOnlyRank0> // see ekat_log_mpi.hpp
 class Logger : public spdlog::logger {
   public:
-    Logger() = delete;
 
-    Logger(const std::string& log_name, const std::string& level_str="debug") :
-     spdlog::logger(LogMpiPolicy::name_with_rank(log_name))
+  // primary constructor
+  Logger(const std::string& log_name, const Log::level::level_enum lev=Log::level::debug,
```
bartgol, line 58: I personally would not default the comm argument. Someone will use the version with defaulted comm (knowing what they are doing), then someone will copy that piece of code (not knowing what they are doing), and all of a sudden we'll have weird logs. If one is ok with a default comm (which wraps MPI_COMM_SELF, which is likely not desired), then they can do so explicitly:
```
Logger<FP,MP> logger("my_log","debug",ekat::Comm());
```

src/ekat/logging/ekat_logger.hpp
```diff
@@ -0,0 +1,137 @@
...
+    LogMpiPolicy::update_sinks(this->sinks(), mpicomm);
+  }
+
+  // Constructor for externally created sinks
+  Logger(const std::string& log_name, const Log::level::level_enum lev,
+    spdlog::sink_ptr csink, spdlog::sink_ptr fsink, const ekat::Comm& mpicomm=ekat::Comm()) :
+    spdlog::logger(LogMpiPolicy::name_with_rank(log_name, mpicomm), {csink, fsink}) {}
+
+  spdlog::sink_ptr get_console_sink() {
+    //return (this->sinks().size() > 1 ? this->sinks()[0] : spdlog::sink_ptr(nullptr));
```
bartgol, line 97: Since you got the null_sink to work, let's remove these commented lines.

bartgol (changes requested):
This looks better. I have a couple more suggestions/comments.

src/ekat/logging/ekat_log_mpi.hpp
pbosler: I removed the MPI Policy template parameter, and replaced that functionality with two class methods, `console_output_rank0_only` and `file_output_rank0_only` that can be called independently of each other.

pbosler:
I got rid of the `dynamic_cast` stuff by creating a `logfile_name_` private class variable for the `ekat::Logger`.  The cost is that it takes extra work to make sure it stays in-sync with the `spdlog` file sink, but I've written tests to make sure it works.

pbosler:
The difficulty with nullifying sinks using `null_sink_mt` pointers had to do with inconsistent `const`ness between references to `spdlog::sink_ptr` returned by various functions.  I put `EKAT_ASSERT`s in each nullify method to ensure it's working.

bartgol (dismissed):
I think this is fairly mature. Later, we might realize some design choices were not optimal, but that's only something we can establish after using this class. Thanks for taking the time to polish it Pete!

tests/logging/mpi_log_all_ranks.cpp
```diff
@@ -12,35 +12,33 @@ bool trace_msg_in_file(Logger<FP>& logger) {
   return (logger.should_log(Log::level::trace)) and (logger.get_file_sink()->should_log(Log::level::trace));
```
bartgol, line 12: So, this only checks that a log and sink should both log, but does not trace if a msg is found in the file, right?

You already spent a lot of time on this, so I'm ok merging. Later, we could do stuff like this:

```
TEST_CASE(blah) {
 ...
  std::string trace_msg = "trace msg";
  std::string debug_msg = "debug msg";
  std::string log_file;
  {
     Logger<...> mylog(...);
     mylog.set_level(Log::level::debug);
     log_file = mylog.logfile_name();
     mylog.trace(trace_msg); // Should not show up
     mylog.debug(debug_msg); // Should show up
  }
  std::ifstream ifile(log_file);
  std::stringstream ss;
  ss << ifile.rdBuf();
  REQUIRE(ss.str().find(trace_msg)==std::string::npos); // Trace output NOT present
  REQUIRE(ss.str().find(debug_msg)!=std::string::npos); // Debug output present  

}
```

pbosler: I'd be happy to put that in.  I'm off tomorrow, so it would be next week.

jeff-cohere (dismissed):
Agree that this is ready enough for merging, but if @pbosler wants to pick at Lucca's last item next week, that's fine too!

Thanks 10**6 for knocking this out so quickly, Pete!

pbosler:
@bartgol It looks like Weaver has a build failure but I can't reproduce it.  I'm trying to configure ekat on my own gpu machine and it failed at the CMake step with this configure file:

> #!/bin/bash
> 
> rm -rf CMake*
> 
> EKAT_SRC=$HOME/ekat
> 
> cmake \
> -DCMAKE_BUILD_TYPE=DEBUG \
> -DCMAKE_INSTALL_PREFIX=. \
> -DKokkos_ENABLE_SERIAL=ON \
> -DKokkos_ENABLE_CUDA=ON \
> -DKokkos_ENABLE_CUDA_LAMBDA=ON \
> -DKokkos_ARCH_MAXWELL52=ON \
> $EKAT_SRC

I've tried all my usual tricks for ensuring that `nvcc_wrapper` works, but I'm not sure what I'm doing wrong with `ekat_mpicxx`.

The output is: 

```
(base) [pabosle@s1024454 Fri Jun 11 10:57 ~/ekat/build-gpu (pbosler/spdlog)]$ ./configureEkat.sh
-- Kokkos_SOURCE_DIR not specified: using submodule version.
CMake Warning (dev) at /projects/sems/install/rhel7-x86_64/sems/utility/cmake/3.17.1/share/cmake-3.17/Modules/GNUInstallDirs.cmake:225 (message):
  Unable to determine default CMAKE_INSTALL_LIBDIR directory because no
  target architecture is known.  Please enable at least one language before
  including GNUInstallDirs.
Call Stack (most recent call first):
  cmake/mpi/EkatSetNvccWrapper.cmake:23 (include)
  CMakeLists.txt:34 (EkatSetNvccWrapper)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- The CXX compiler identification is unknown
-- The Fortran compiler identification is GNU 8.3.0
-- Check for working CXX compiler: /ascldap/users/pabosle/ekat/build-gpu/bin/ekat_mpicxx
-- Check for working CXX compiler: /ascldap/users/pabosle/ekat/build-gpu/bin/ekat_mpicxx - broken
CMake Error at /projects/sems/install/rhel7-x86_64/sems/utility/cmake/3.17.1/share/cmake-3.17/Modules/CMakeTestCXXCompiler.cmake:53 (message):
  The C++ compiler

    "/ascldap/users/pabosle/ekat/build-gpu/bin/ekat_mpicxx"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /ascldap/users/pabosle/ekat/build-gpu/CMakeFiles/CMakeTmp

    Run Build Command(s):/usr/bin/gmake cmTC_7432d/fast && /usr/bin/gmake  -f CMakeFiles/cmTC_7432d.dir/build.make CMakeFiles/cmTC_7432d.dir/build
    gmake[1]: Entering directory `/home/pabosle/ekat/build-gpu/CMakeFiles/CMakeTmp'
    Building CXX object CMakeFiles/cmTC_7432d.dir/testCXXCompiler.cxx.o
    /ascldap/users/pabosle/ekat/build-gpu/bin/ekat_mpicxx     -o CMakeFiles/cmTC_7432d.dir/testCXXCompiler.cxx.o -c /ascldap/users/pabosle/ekat/build-gpu/CMakeFiles/CMakeTmp/testCXXCompiler.cxx
    /ascldap/users/pabosle/ekat/build-gpu/bin/ekat_mpicxx: line 9: -o: command not found
    gmake[1]: *** [CMakeFiles/cmTC_7432d.dir/testCXXCompiler.cxx.o] Error 127
    gmake[1]: Leaving directory `/home/pabosle/ekat/build-gpu/CMakeFiles/CMakeTmp'
    gmake: *** [cmTC_7432d/fast] Error 2
```

bartgol:
It looks like you did not specify the CXX compiler. Ekat will use `ekat_mpicxx` to wrap mpicxx, but you need to tell cmake the name of the mpi compiler. So you need `-D CMAKE_CXX_COMPILER:STRING=mpicxx` as well. If you take a look at `bin/ekat_mpicxx.in` you will see immediately what's happening: with no CMAKE_CXX_COMPILER, the script has no MPICXX var defined, so when expanding `${MPICXX} ""$@""` you simply get `""$@""` (which are the args passed to the script. That's why cmake says `-o: command not found`, cause that's the first arg passed to ekat_mpicxx (the compiler wrapper).

src/ekat/logging/ekat_logger.hpp
```diff
@@ -125,9 +125,9 @@ class Logger : public spdlog::logger {
 
   Log::level::level_enum log_level() const {return this->level();}
 
-  Log::level::level_enum console_level() const {return this->get_console_sink()->level();}
+  Log::level::level_enum console_level() {return this->get_console_sink()->level();}
```
bartgol, line 128: I think instead of removing const here, you should add const to `get_console_sink` and `get_file_sink`. Those methods return ptrs, so even if someone changes the actual sinks, they won't change the ptrs stored by this class, so it's ok to make them const.

I think someone _might_ call `console_level` at some point from within a const method, and without the const here, that won't work.

bartgol (changes requested):
I think I have one last request. Sorry. :)

src/ekat/logging/ekat_logger.hpp
pbosler: Thanks @bartgol . This makes sense, and reminds me that I've been staring at my screen too long (should have thought of it myself).

bartgol (approved):

PR #118, opened 2021-07-12
Enable kokkos bounds checking by default if DEBUG is enabled

jgfouca:
I think enabling KOKKOS_DEBUG used to turn on bounds checking, but
that is no longer the case. I believe most, if not all, EKAT clients
will want bounds checking on Kokkos views in DEBUG mode, so I think it's
reasonable to turn this on by default if the CMake build type is DEBUG.

The SCREAM project relies on EKAT to configure Kokkos and out-of-bounds errors
have slipped into the code in the period since we stopped doing bounds checking.
Some of these errors were subtle and required significant debugging but would have
been much easier to detect and fix with bounds-checked views.


AaronDonahue (approved):
Looks good to me.

PR #119, opened 2021-07-20
Allow users to enable code coverage

jgfouca:
Allow users of ekat to add `-DEKAT_ENABLE_COVERAGE=ON` which will add coverage flags to compiler flags for C, CXX, and Fortran. This will be useful for developers and allow us to add a nightly build that uploads coverage.


AaronDonahue (approved):

CMakeLists.txt
```diff
@@ -78,6 +78,11 @@ set (EKAT_TPL_LIBRARIES ${EKAT_TPL_LIBRARIES_INTERNAL} CACHE INTERNAL "List of E
 option (EKAT_MPI_ERRORS_ARE_FATAL " Whether EKAT should crash when MPI errors happen." ON)
 option (EKAT_DEFAULT_BFB "Whether EKAT should default to BFB behavior whenever possible/appropriate." ${EKAT_IS_DEBUG_BUILD})
 option (EKAT_ENABLE_VALGRIND "Whether to run tests with valgrind" OFF)
+option (EKAT_ENABLE_COVERAGE "Whether to enable code coverage" OFF)
+
+if (EKAT_ENABLE_COVERAGE AND NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
```
bartgol, line 83: We should do a case-insensitive check of the build type. For instance, I always use `DEBUG` and `RELEASE` in my cmake scripts...

PR #121, opened 2021-08-04
Add better support for views of const packs in API

jgfouca:
Add support for having input args be Views of const Packs
for lin_interp. This required adding support for Views of const
Packs for all the repack functions as well.

## Motivation

@AaronDonahue is trying to use lin_interp on constant Field views.

## Testing
Added a lin_interp test that tries combinations of View<const Pack> inputs.

### DIFF
--- a/src/ekat/ekat_pack_kokkos.hpp
+++ b/src/ekat/ekat_pack_kokkos.hpp
@@ -178,6 +178,21 @@ repack (const Kokkos::View<Pack<T, old_pack_size>**, Parms...>& vp) {
     (old_pack_size / new_pack_size) * vp.extent_int(1));
 }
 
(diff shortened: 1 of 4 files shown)


jgfouca:
@bartgol , I thought it was unfortunate that I had to duplicate all the repack functions in order to get this to work. I'm wondering if there is some template magic I could have done instead. I tried changing Pack to a template argument, so that both `Pack` and `const Pack` would be accepted, but the compiler rejected that approach, at least the way I tried to do it.

jgfouca:
@tcclevenger , added you as a review since both Luca and I have done development on this branch.

tcclevenger (approved):
Looks good to me.

PR #122, opened 2021-08-05
Improved readme file

bartgol:
## Motivation

This is the first attempt to have a bit of documentation in EKAT. The readme file is the first entry point for a new user, so explain a bit what EKAT contains, and how to config+build.


jeff-cohere (approved):
This is a great start!

PR #123, opened 2021-08-05
Better handle the OpenMP dependency in ekat

bartgol:
## Motivation

Kokkos still relies on setting the flags. But that can be problematic. E.g., Kokkos adds its link options to the target link properties only if the link language is CXX (this helps avoiding cuda link options to be passed on to a non-cuda linker). However, when the main entry point is in F90, and the link language is Fortran, this causes the lack of openmp flags in the link line, and consequently undef refs to omp stuff.

Instead, if kokkos enables openmp, we explicitly link ekat against the openmp cmake target (for cxx and fortran). This will allow all compiler flags and linker flags to be correctly propagated, without the need to assume anything about the link/compiler language of the downstream applications.

## E3SM Stakeholder Feedback

@tcclevenger is currently testing this solution for a linker error in the scream repo, when building the whole e3sm executable (which uses Fortran link language).

## Testing

I tried to create a mini f90 executable, with Fortran linker language, and it failed building without this mod. It correctly builds now. I did not include the f90 executable, since it was just to test correctness of the cmake logic, not a _real_ test.


jgfouca (dismissed):

jgfouca (approved):

bartgol:
I'm gonna try to build SCREAM via cime on mappy, and see if this removes the build error

bartgol:
This PR seems to fix scream cime build. Merging.

PR #124, opened 2021-08-17
Jgfouca/lin interp imprv

jgfouca:
Make the LinInterp API more flexible by allowing user to provide a range policy to lin_interp and setup. This will allow it to be used inside of TeamThreadRange kernels. The user can also specify the column being interpolated if they cannot rely on the league_rank = col assumption.

Merges tests of these new capabilities into the main Lin-interp soak test.

Also, minor clean ups of whitespace, variable names, and code structure. Adds better documentation for the main LinInterp API.

### DIFF
--- a/src/ekat/util/ekat_lin_interp.hpp
+++ b/src/ekat/util/ekat_lin_interp.hpp
@@ -69,8 +69,10 @@ struct LinInterp
 
   LinInterp(int ncol, int km1, int km2, Scalar minthresh);
 
+  // Simple getters
   int km1_pack() const { return m_km1_pack; }
   int km2_pack() const { return m_km2_pack; }
+  const TeamPolicy& policy() const { return m_policy; }
 
   // Setup the index map. This must be called before lin_interp. By default, will launch a
   // TeamThreadRange kernel. By default, the column idx will be team.league_rank(); this can be
@@ -82,7 +84,7 @@ struct LinInterp
     const V1& x1,
     const V2& x2,
     const Int col=-1) const;
-  
+
   // Same as above except uses a user-provided range boundary struct. This will likely
   // be a TeamVectorRange.
   template<typename V1, typename V2, typename RangeBoundary>
@@ -124,6 +126,7 @@ struct LinInterp
   //
   // -------- Internal API, data ------
   //
+ private:
 
   template <typename RangeBoundary>
   KOKKOS_INLINE_FUNCTION
@@ -144,7 +147,6 @@ struct LinInterp
     const view_1d<const Pack>& x1, const view_1d<const Pack>& x2, const view_1d<const Pack>& y1,
     const view_1d<Pack>& y2,
     const Int col);
-
 
   int m_km1;
   int m_km2;
--- a/src/ekat/util/ekat_lin_interp.hpp
+++ b/src/ekat/util/ekat_lin_interp.hpp
@@ -86,7 +86,7 @@ struct LinInterp
     const Int col=-1) const;
 
   // Same as above except uses a user-provided range boundary struct. This will likely
-  // be a TeamVectorRange.
+  // be a ThreadVectorRange.
   template<typename V1, typename V2, typename RangeBoundary>
   KOKKOS_INLINE_FUNCTION
   void setup(
@@ -111,7 +111,7 @@ struct LinInterp
     const Int col=-1) const;
 
   // Same as above except uses a user-provided range boundary struct. This will likely
-  // be a TeamVectorRange.
+  // be a ThreadVectorRange.
   template <typename V1, typename V2, typename V3, typename V4, typename RangeBoundary>
   KOKKOS_INLINE_FUNCTION
   void lin_interp(
--- a/src/ekat/util/ekat_lin_interp.hpp
+++ b/src/ekat/util/ekat_lin_interp.hpp
@@ -110,7 +110,7 @@ struct LinInterp
   int m_km2_pack;
   Scalar m_minthresh;
   TeamPolicy m_policy;
-  view_2d<IntPack> m_indx_map; // [x2-idx] -> x1-idx
+  view_2d<IntPack> m_indx_map; // [x2_idx] -> x1_idx
 };
 
 } //namespace ekat
--- a/src/ekat/util/ekat_lin_interp.hpp
+++ b/src/ekat/util/ekat_lin_interp.hpp
@@ -72,49 +72,80 @@ struct LinInterp
   int km1_pack() const { return m_km1_pack; }
   int km2_pack() const { return m_km2_pack; }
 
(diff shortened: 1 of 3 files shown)


src/ekat/util/ekat_lin_interp.hpp
```diff
@@ -87,6 +87,15 @@ struct LinInterp
                   const V3& y1,
                   const V4& y2) const;
 
+  template <typename V1, typename V2, typename V3, typename V4, typename RangePolicy>
+  KOKKOS_INLINE_FUNCTION
+  void lin_interp(const MemberType& team,
+                  const RangePolicy& range_policy,
```
bartgol, line 93: I think `RangePolicy` might be a confusing name, leading someone to confuse team vs range policies (in the kokkos sense). Perhaps we could call the typename `RangeBoundariesStruct`. After all, in Kokkos, `TeamThreadRange(..)` and `ThreadVectorRange(..)` both return an `XYZRangeBoundaryStruct`, with `XYZ=TeamThread` and `XYZ=ThreadVector` respectively. Also, these structs are not reall "policies", per se...

src/ekat/util/ekat_lin_interp_impl.hpp
```diff
@@ -28,17 +28,17 @@ void LinInterp<ScalarT, PackSize, DeviceT>::setup(const MemberType& team,
...
-                                             const V3& y1,
-                                             const V4& y2) const
+                                                       const V1& x1,
+                                                       const V2& x2,
+                                                       const V3& y1,
+                                                       const V4& y2) const
 {
   lin_interp_impl(team,
+                  Kokkos::TeamThreadRange(team, m_km2_pack),
                   *this,
```
bartgol, line 42: This is not something that was done in this PR, but I can't help asking: what's the point of passing `*this`? The `lin_interp_impl` routine is probably not meant to be called by the user (possibly passing another LinInterp object). And what would be the use case of doing the following?
```
LinInterp li_1(..);
LinInterp li_2(...);
li_1.lin_interp_impl(team, li_2,...);
```
Unless I'm overlooking something, can we remove the `*this` parameter?

tests/algorithm/lin_interp_test.cpp
```diff
@@ -233,6 +225,45 @@ TEST_CASE("lin_interp_api", "lin_interp")
...
+                       vect.m_policy,
+                       KOKKOS_LAMBDA(typename LIV::MemberType const& team) {
+    const int i = team.league_rank();
+    vect.setup(team,
+               ekat::subview(x1kv, i),
+               ekat::subview(x2kv, i));
+    team.team_barrier();
+
+    Kokkos::parallel_for(
+      Kokkos::TeamThreadRange(team, 0), [&] (int k) {
```
bartgol, line 257: Perhaps we should do non-trivial sizes, just in case some counter is badly handled and ends up out of bounds. Not really interested in the numerical results, just making sure the code doesn't crash.

tests/algorithm/lin_interp_test.cpp
```diff
@@ -233,6 +225,45 @@ TEST_CASE("lin_interp_api", "lin_interp")
...
+
+        vect.lin_interp(team,
+                        Kokkos::TeamVectorRange(team, 0),
```
bartgol, line 260: Another interesting test would be to run a LI test on 2d views, using two team policies with two combinations of (team_size,vector_length): the first with (N,1). and the second with (1,N). Notice that the latter is effectively the same as (1,1) on non-GPU device.

The two should yield the same result, whatever it is, but if there's some race condition related to the ThreadVectorRange, we should get a failure.

bartgol (changes requested):
Thanks for doing this very quickly Jim! I have a couple of request, one related to naming, and the other to testing. If you're swamped, feel free to offload some work. However, while the first request might be personal taste, I really think we should test better the correctness of the TVR impl (see my comment on that).

src/ekat/util/ekat_lin_interp.hpp
jgfouca: Agree.

src/ekat/util/ekat_lin_interp_impl.hpp
jgfouca: @bartgol `liv` is needed in `lin_interp_impl` to provide the `m_indx_map` and other members which are used in the implementation. Calling a static function with a `*this` argument is a common technique I've used when writing classes/structs that launch Kokkos kernels. It's been a while since I've looked at the details, but I think Kokkos doesn't handle the implicit `this` pointer well.

bartgol: `lin_interp_impl` is not a static method though, so you don't need to pass `*this`.

Re: kokkos and *this. The issue is when dispatching a cuda kernel (meaning the outer-most one) using a lambda in the parallel_for. That's b/c the automatic vars need to be copied to device, and `this` is copied instead of `*this` (an issue solved with cuda 11, IIRC). Here, neither concern hold: we're already on device (we are inside a || region), and we are using a member function instead of a lambda in the parallel for).

jgfouca: @bartgol , it is: `static void lin_interp_impl`.

Thanks for clarifying the `this` situation. I will see if a regular member function works as the implementation.

bartgol: > @bartgol , it is: `static void lin_interp_impl`.

Ah, you're right, sorry. I was looking at the `_impl.hpp` file, not at the decl one.

Imho, we can do away with static, and remove the extra parameter `LinInterp&`. Since there's no need for a static fcn, emulating a non-static fcn via a static fcn is probably confusing to most people (it is to me, at least).

jgfouca:
@bartgol , I think this is ready for re-review (if the AT passes).

bartgol (approved):

AaronDonahue:
Just auditing my SCREAM PR's and wanted to check in on this one.  Any updates?

bartgol:
@AaronDonahue @jgfouca I can reproduce the blake failure with my env, which has different compilers/tpls versions, so it's a legit issue. It's got something to do with pack size, since tests pass if I change pack size to 1 on blake. But it's not just the pack size, since tests pass on mappy, which also uses pack size 16, like blake.

bartgol:
I'm trying to debug, but can't run valgrind due to unrecognized instruction (probably some intel stuff). Inspecting in gdb is also hard, since it showing nonsense stuff. E.g., right after `const int ncol = 10;`, printing ncol shows 0...

bartgol:
@jgfouca I think the failure might be due to `EKAT_TEST_POSSIBLY_NO_PACK_SIZE`, which gives pack size 1 on SKX, but all views are packed "normally".

jgfouca:
@bartgol , looking at it

bartgol (approved):

PR #126, opened 2021-08-21
Add wrappers to common MPI communication calls to Comm class

bartgol:
## Motivation

MPI calls can be clunky, and these interfaces make the signature of a call a tad shorter. In particular, Comm now offers: scan, all_reduce, broadcast, and barrier. The first three are templated on the values type, but only int, float, and double are currently supported.

## Testing

All communication functions are tested in `tests/mpi/comm.cpp`.


AaronDonahue (approved):
Nice set of changes.  Thanks Luca

PR #127, opened 2021-08-23
Fix lin_interp test on CUDA

jgfouca:
km2_pack is not a kokkos function, need to call it outside of kernels.


bartgol (commented):
Why not adding `KOKKOS_INLINE_FUNCTION` to both `km1_pack()` and `km2_pack()` ?

bartgol (approved):

PR #130, opened 2021-08-26
EkatCreateUnitTest now uses mpiexec (etc) to run tests with np = 1

### LINKED ISSUE #129: Let EkatCreateUnitTest use `mpiexec` (etc) for `nproc = 1`
Some platforms seem to require the use of `mpiexec` or `srun` even when running "serially" (`nproc = 1`). This is a wrinkle that drives some folks to support a fake stubbed MPI implementation for libraries like EKAT. Since we've decided not to fake MPI for serial builds, we always need it, and this means we have to accept the constraints that it poses on our target platforms.

Given this, I think we should modify `EkatCreateUnitTest` to use the appropriate MPI run command even for "serial" processes, since it is necessary in the general case. It's an easy modification to accommodate `mpiexec`--we'd just comment out the checks for `nproc > 1`. However, if we wanted to do it "correctly", we'd also have to detect a SLURM environment and use `srun` (or whatever other environments we support for E3SM).

jeff-cohere:
This PR changes  the `EkatCreateUnitTest` CMake function to use the provided name for the program that runs MPI (`mpiexec`, `mpirun`, `srun`, etc), even if the number of processes needed for the test is 1. This allows EKAT-generated tests to run on platforms that require this program to be specified always.

It also caches default values for MPI-related test info. 

Fixes #129


bartgol:
@jeff-cohere Sorry, I was off friday, and took a day off yesterday. Yes, I'll merge.

PR #131, opened 2021-08-31
EkatCreateUnitTest now runs unit tests serially if MPI_EXEC_NAME is omitted.

jeff-cohere:
This change allows us to run tests in environments like Docker containers where we aren't allowed to use `mpiexec` (because, for example, of the ignorant-but-ubiquitous tendency of Docker to require running as `root`).

### DIFF
--- a/cmake/EkatCreateUnitTest.cmake
+++ b/cmake/EkatCreateUnitTest.cmake
@@ -221,13 +221,11 @@ function(EkatCreateUnitTest target_name target_srcs)
     message (FATAL_ERROR "Error! THREAD_START < THREAD_END, but the increment is negative.")
   endif()
 
-  # Since we can't build without MPI, we must be prepared to supply the right
-  # MPI_EXEC.
-  if ("${ecut_MPI_EXEC_NAME}" STREQUAL "")
-    set (ecut_MPI_EXEC_NAME "mpiexec")
-  endif()
-  if ("${ecut_MPI_NP_FLAG}" STREQUAL "")
-    set (ecut_MPI_NP_FLAG "-n")
+  # If MPI_EXEC_NAME wasn't given, make sure we don't need more than one proc.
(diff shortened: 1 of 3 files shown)


bartgol (approved):

bartgol (dismissed):
CMake has the bad feature that, if the var is an empty string, the call to `cmake_parse_arguments` throws a warning. I wonder if we should default all these optional vars to `UNSET`, and make sure the user does not set them to empty strings. Maybe I'm overthinking this though, and definitely not something to do in this PR.

jeff-cohere:
I agree with your suggestion. I wouldn't even mind making the change in this PR if it's simple enough.

bartgol:
I now notice that with an empty MPI_EXEC we get errors. Perhaps we should check inputs consistency in `tests/CMakeLists.txt`, to ensure the user does not set wrong things. Even better: what if we offer the cmake option `EKAT_TEST_DISABLE_MPI`, and if ON, we set all mpi-related stuff to serial-looking values?

bartgol:
> I added MPI_EXEC_NAME to the logging tests to prevent issues. I'm open to an `EKAT_TEST_DISABLE_MPI` setting.

Maybe a more robust fix would be to change
```
set (EKAT_TEST_MAX_RANKS 4 CACHE STRING "Max number of ranks for tests")
```
to
```
if (EKAT_TEST_MPI_EXEC_NAME)
  set (EKAT_TEST_MAX_RANKS 4 CACHE STRING "Max number of ranks for tests")
else()
  set (EKAT_TEST_MAX_RANKS 1 CACHE STRING "Max number of ranks for tests")
endif()
```
So that we only run serial tests if no mpiexec was provided. If one still sets max_ranks>1 while asking for no mpiexec, then erroring out is the right thing, since they clearly have no idea what they are doing...

jeff-cohere:
I made the change you suggested.

bartgol (approved):

PR #133, opened 2021-09-13
Only print_config on rank 0

tcclevenger:
`ekat_initialize_test_session()` was printing configs on every rank when called through `ekat_catch_main()`, but print info contains no rank-specific info. I added a `print_config` input arg which is set `mpi_rank == 0`. So now `initialize_ekat_session()` only prints info if `rank=0`.

Another thing added (and vetos are welcome) is that the
```std::cout << "Starting catch session on rank " << comm.rank() << " out of " << comm.size() << "\n";```
call output becomes very messy and intertwined with other output. I added barriers around it and put it in a DEBUG ifdef.

### DIFF
--- a/src/ekat/util/ekat_catch_main.cpp
+++ b/src/ekat/util/ekat_catch_main.cpp
@@ -9,7 +9,7 @@
 
 #include <mpi.h>
 
-void ekat_initialize_test_session (int argc, char** argv);
+void ekat_initialize_test_session (int argc, char** argv, const bool print_config);
 void ekat_finalize_test_session ();
 
 int main (int argc, char **argv) {
@@ -64,10 +64,9 @@ int main (int argc, char **argv) {
   }
 
   // Initialize test session (initializes kokkos and print config settings).
-  // Ekat provides a defalt impl, but the user can choose
-  // Ekat provides a defalt one, but the user can choose
+  // Ekat provides a default impl, but the user can choose
   // to not use it, and provide one instead.
-  ekat_initialize_test_session(args.size(),args.data());
+  ekat_initialize_test_session(args.size(),args.data(),(comm.rank()==0));
 
   std::cout << "Starting catch session on rank " << comm.rank() << " out of " << comm.size() << "\n";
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by jeff-cohere — APPROVED** [review]
Looks good to me! I think the barriers around the init message are a good idea.

File: src/ekat/util/ekat_catch_main.cpp [hunk 1]
```diff
@@ -64,12 +64,15 @@ int main (int argc, char **argv) {
...
   // Initialize test session (initializes kokkos and print config settings).
-  // Ekat provides a defalt impl, but the user can choose
-  // Ekat provides a defalt one, but the user can choose
+  // Ekat provides a default impl, but the user can choose
   // to not use it, and provide one instead.
-  ekat_initialize_test_session(args.size(),args.data());
+  ekat_initialize_test_session(args.size(),args.data(),(comm.rank()==0));
 
+#ifndef NDEBUG
+  MPI_Barrier(comm.mpi_comm());
```
bartgol, line 72: What's the rationale for these barriers?

bartgol: Nvm, I think I get it: you don't want the following lines to be interleaved with the config print above or with other stuff from the test below. Makes sense.

src/ekat/util/ekat_catch_main.cpp
```diff
@@ -64,12 +64,15 @@ int main (int argc, char **argv) {
   }
 
   // Initialize test session (initializes kokkos and print config settings).
-  // Ekat provides a defalt impl, but the user can choose
-  // Ekat provides a defalt one, but the user can choose
+  // Ekat provides a default impl, but the user can choose
   // to not use it, and provide one instead.
-  ekat_initialize_test_session(args.size(),args.data());
+  ekat_initialize_test_session(args.size(),args.data(),(comm.rank()==0));
```
bartgol, line 69: Minor, but you could use `comm.am_i_root()`, if you wanted. Not a big deal though.

bartgol (approved):

src/ekat/util/ekat_catch_main.cpp
tcclevenger: Yes, that's much cleaner.

tcclevenger:
Thanks @jeff-cohere and @bartgol! I'll need 1 more approval after the latest changes.

bartgol (approved):

PR #137, opened 2021-09-14
Some mods to compiler flags and vectorization pragmas

### LINKED ISSUE #136: Add fopenmp-simd/qopenmp-simd to CXX flags, to use pragma omp simd
Both gcc and intel support the pragma `omp simd` for vectorization. This pragma is enabled with `-qopenmp-simd` and `-fopenmp-simd` for intel and gcc respectively. These flags do _not_ require OpenMP runtime, and no other omp feature is needed/linked. These flags simply allow to use that pragma for vectorization.

Note: this should get rid of the warning on serial intel builds

```
externals/ekat/src/ekat/ekat_pack.hpp(182): warning #3948: simd pragma has been deprecated, and will be removed in a future release.  Please refer to release notes for details and recommended alternatives
```

bartgol:
## Motivation

As explained in #136, Intel was complaining in Kokkos Serial builds, since the pragma simd is deprecated. In order to use pragma omp simd a special compiler flag is needed. The flag is `-qopenmp-simd` for intel, and `-fopenmp-simd` for GNU. Despite the name, pragma omp simd does _not_ require OpenMP features/libs. If OpenMP is also enabled (via `-qopenmp` or `-fopenmp`), the above flag has no effect (since it is already enabled by openmp).

Additionally:
* removed pragma ivdep. It was used only in the macro `ekat_masked_loop_no_force_vec`, which was never used (not even in scream), except for a unit test that only checked the macro itself...
* cleaned up the SetCompilersFlag macros, including removing unused code (brought in by the copy+paste from Homme's SetCompilersFlags).

## Related Issues
Closes #136

### DIFF
--- a/cmake/EkatSetCompilerFlags.cmake
+++ b/cmake/EkatSetCompilerFlags.cmake
@@ -50,6 +50,7 @@ macro (SetCompilerFlags)
     set (CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -Wall")
   endif()
 
+  # Language-specific flags to be used regardless of build type.
   if (DEFINED BASE_FFLAGS)
     set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} ${BASE_FFLAGS}")
   else ()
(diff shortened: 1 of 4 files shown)


jgfouca (approved):
I you are sure we won't ever need the no_force_vec loop, then I'm fine with this.

bartgol:
> I you are sure we won't ever need the no_force_vec loop, then I'm fine with this.

It seems like such a corner case. I don't feel the _need_ to have the macro in EKAT...

PR #138, opened 2021-09-14
Fix (and simplify) YAML parser

bartgol:
## Motivation

There was an issue with the YAML parser, where negative doubles were parsed as strings. For some reason in my implementation I was not using the `std::stod` and `std::stoi` functions. Using those, the implementation of the parser is much simpler.

## Testing

I modified the input file of the yaml parser unit test, to have also negative doubles.


AaronDonahue (approved):
Thanks for the quick fix!

PR #139, opened 2021-09-14
Fix EkatCreateUnitTest

bartgol:
## Motivation

Recent changes to the macro introduced a bug, only exposed when testing through ctest, with a resource specification file. In particular, the macro set the `RESOURCE_GROUPS` property to request $nranks+1 copies of `device:1`, rather than $nranks copies. When the resources are indeed limited (as it happens on weaver, e.g.), this caused tests to crap out.

Additionally, I modified the number of devices requested to $nranks*$nthreads, to guarantee more space to omp tests.

## E3SM Stakeholder Feedback

This is needed by scream, where weaver testing is failing due to the resources mis-specification.


tcclevenger (approved):
Thanks Luca!

PR #141, opened 2021-09-16
Reinstate running tests in a shell

bartgol:
## Motivation

Running tests in a shell is necessary to be able to use some shell-specific features. E.g., input/output redirection only works in a shell. Without a shell, something like `./my_exec < my_infile` is parsed as "run my_exec, and pass the two strings '<' and 'my_infile' to its argv inputs". For catch2 tests, this means that the test will try to use the two filters "<" and "my_infile", likely finding that no test has to be run.

## E3SM Stakeholder Feedback

The lack of input redirection was causing some tests to "pass" in SCREAM, because the string "<namelist.nl" was interpreted as a string to filter the tests with, causing no test to be run, hence the test "passed".

## Testing

I added a nano test, that simply contains `REQUIRE(false)`, and told ctest that the test should fail. Then, I added an input redirection as EXE_ARGS for this test. If the redirection works, the executable is run, a non-zero return code follows, and ctest is happy. However, if for some reason we break again the input redirection, the EXE_ARGS are interpreted as test filter, no test is run, the executable returns 0, and ctest will crap out.

### DIFF
--- a/tests/utils/regress_fail.cpp
+++ b/tests/utils/regress_fail.cpp
@@ -1,0 +1,11 @@
+#include <catch2/catch.hpp>
+
+#include "ekat/ekat_assert.hpp"
+
+namespace {
+
+TEST_CASE ("doomed") {
+  REQUIRE(false);
+}
+
+} // anonymous namespace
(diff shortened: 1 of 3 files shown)


tests/utils/regress_fail.cpp
```diff
@@ -0,0 +1,11 @@
+#include <catch2/catch.hpp>
+
+#include "ekat/ekat_assert.hpp"
+
+namespace {
+
+TEST_CASE ("doomed") {
```
jeff-cohere, line 7: 🤣

jeff-cohere (approved):
Looks good. Approved subject to test results.

PR #143, opened 2021-09-16
Do not use try-catch blocks when parsing YAML files

bartgol:
## Motivation

These blocks were inside functions checking if the input string was an int or a double. This makes debugging with 'catch throw' a nightmare. The new impl is exception free.

## Testing
No additional testing needed. Parser is already tested.


jeff-cohere (approved):
Sounds like a good idea.

bartgol:
@jeff-cohere I had to go through 50+ exceptions in gdb at the program startup, just to get through the YAML file parsing... I ended up having to disable exceptions catch, putting a breakpoint right after the parsing, then re-adding `catch throw`. Nonsense.

PR #144, opened 2021-09-22
Some safety fixes to the logic that generates the ekat_mpicxx wrapper script

bartgol:
## Motivation

The cmake logic to generate ekat_mpicxx (the scripts that wraps mpicxx, setting OMPI_CXX (or equivalent) to point to nvcc_wrapper) contains some guards to avoid infinite recursion. This logic boiled down to compare whether the CMAKE_CXX_COMPILER was already set to be the ekat_mpicxx wrapper. However, the logic was buggy in the (remote, but it happened to me) case where the two filenames compared were different string, yet pointing to the same file (due to some symlinks in the filesystem).

This micro PR fixes this.

## Testing

Hard to test this one. You would have to be on a particular filesystem, configure ekat, then modify something that triggers cmake to re-run with the next make. If you end up with `${CMAKE_CXX_COMPILER}` different from `${CMAKE_BINARY_DIR}/bin/ekat_mpicxx`, then you get the infinite recursion. I managed to trigger this bug on a gpu platform, and verified manually that this fix is indeed...a fix.

### DIFF
--- a/cmake/mpi/EkatSetNvccWrapper.cmake
+++ b/cmake/mpi/EkatSetNvccWrapper.cmake
@@ -3,11 +3,18 @@
 #       of ekat_mpicxx.in relative to this file.
 set (MPICXX_WRAPPER_SOURCE_DIR ${CMAKE_CURRENT_LIST_DIR}/../../bin)
 macro (EkatSetNvccWrapper)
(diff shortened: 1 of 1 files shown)


jeff-cohere (approved):
Must have been fun to find this bug!

PR #145, opened 2021-09-22
Allow --map-by as MPI res specification flag, plus test-launcher script to wrap tests

bartgol:
## Motivation

When running multiple tests, and/or MPI, and/or OpenMP tests, we get slow performance if 2+ tests and/or 2+ ranks and or 2+ threads end up on the same processing unit. This has become a big burden for SCREAM, where the nice test parallelization efforts are somewhat invalidated by having 20+ executables clashing on the same cores.

This PR does 2 things:
 - It allows (though not by default) to use `--map-by` instead of `-np` to specify mpi resources, by setting `MPI_NP_FLAG` to `--map-by` when calling `EkatCreateUnitTest`. If so, the macro will have ctest invoke the executable as `mpiexec --map-by ppr:$NRANKS:node:pe=$NTHREADS ...` instead of `mpiexec -np $NRANKS ...`.
  - It adds (and use) a python wrapper for the unit tests executables, called `test-launcher`. This script does very little in case ctest is a) not used (e.g., I run tests by hand) or b) invoked without the `--resource-spec-file <file>` option. But if one uses ctest with such option, this python wrapper ensures that all tests are run on different ranks, and all tests have enough resources to accommodate all ranks/threads.

## E3SM Stakeholder Feedback

This PR is highly desired by SCREAM.
## Testing

I manually tested this branch with SCREAM, and noticed a tremendous speedup. For instance, running only the valgrind tests build for SCREAM, since tests were spread out across cores, we had test runtime going down from 100s to 5s for "fast tests".

### DIFF
--- a/bin/test-launcher
+++ b/bin/test-launcher
@@ -112,8 +112,8 @@ def run_test(executable,exec_args):
 
     # Check if we're run with MPI
     if "OMPI_COMM_WORLD_SIZE" in env:
-        nranks = int(env["OMPI_COMM_WORLD_SIZE"])
-        myrank = int(env["OMPI_COMM_WORLD_RANK"])
+        nranks = int(env["${EKAT_MPIRUN_COMM_WORLD_SIZE}"])
+        myrank = int(env["${EKAT_MPIRUN_COMM_WORLD_RANK}"])
     else:
         nranks = 1
         myrank = 0
--- a/bin/test-launcher
+++ b/bin/test-launcher
@@ -156,7 +156,7 @@ def run_test(executable,exec_args):
         # variable TEST_LAUNCHER_ON_GPU to either True or False
         if ${TEST_LAUNCHER_ON_GPU}:
             expect (len(my_ids)==1, "Error! For GPU runs, each rank should use only one device.")
-            exec_args.append(" --kokkos-device-id={}".format(ids[0]))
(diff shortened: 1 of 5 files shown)


bin/test-launcher
```diff
@@ -0,0 +1,181 @@
...
+    # omp_env.append("OMP_DISPLAY_ENV=verbose")
+    # omp_env.append("OMP_DISPLAY_AFFINITY=true")
+    omp_env.append("OMP_PROC_BIND=spread")
+    omp_env.append("OMP_PLACES=threads")
+
+    env = os.environ
+
+    # Check if we're run with MPI
+    if "OMPI_COMM_WORLD_SIZE" in env:
+        nranks = int(env["OMPI_COMM_WORLD_SIZE"])
```
jgfouca, line 115: is `OMPI_COMM_WORLD_SIZE` portable across mpi types or only applies to openmpi?

bartgol: Ah, good catch! I noticed that in Kokkos, and meant to fix it, but I forgot. I'll fix it.

jgfouca (approved):
If MPICH also uses `OMPI_COMM_WORLD_SIZE`, then I approve.

bartgol:
> If MPICH also uses `OMPI_COMM_WORLD_SIZE`, then I approve.

It doesn't. But I think the last commit took care of making the logic portable to MPICH.

PR #148, opened 2021-09-27
Add support for enabling cuda-memcheck

jgfouca:
Add support for running cuda-memcheck on all tests. You can do this via cmake setting: `EKAT_ENABLE_CUDA_MEMCHECK=ON`.

## Motivation

Cuda memory problems have caused a lot of disruption to our weaver AT testing. Detecting and fixing cuda memory errors will help keep things running smoothly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,6 +148,10 @@ if (EKAT_ENABLE_CUDA_MEMCHECK AND NOT Kokkos_ENABLE_CUDA)
   message(FATAL_ERROR "Makes no sense to turn on cuda-memcheck without CUDA")
 endif()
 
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -143,6 +144,10 @@ if (EKAT_ENABLE_TESTS)
   add_subdirectory(tests)
 endif()
 
+if (EKAT_ENABLE_CUDA_MEMCHECK AND NOT Kokkos_ENABLE_CUDA)
```
bartgol, line 147: Should we also check that one does not request both valgrind and cuda-memcheck at the same time? The macro `EkatCreateUnitTest` would pick valgrind in that case, while that's probably not what the user wants.

jgfouca: Good call

bartgol (approved):

PR #154, opened 2021-10-20
Add params/sublists iterators to ParameterList

bartgol:
## Motivation

This PR allows to iterate over a ParameterList's stored parameters/sublists, in a const way. This allows to avoid the clunky construct of doing:

```
const auto& sublists_names = pl.get<std::vector<std::string>>("sublists names");
for (const auto& n : sublists_names) {
  const auto& sub_pl = pl.sublist(n);
  ...
}
```
and simply do

```
for (auto it = p.sublists_cbegin(); it!=p.sublists_cend(); ++it) {
  const auto& sub_pl.sublist = it.second;
  ...
}
```
And avoid forcing users to add the "sublist names" entry to the parameter list (often, the input yaml file) just for the sake of being able to access sublists.

## Testing

Added small test in `util_tests.cpp`.

### DIFF
--- a/src/ekat/std_meta/ekat_std_map_key_iterator.hpp
+++ b/src/ekat/std_meta/ekat_std_map_key_iterator.hpp
@@ -65,8 +65,8 @@ public:
   using iterator_category = std::bidirectional_iterator_tag;
   using difference_type   = std::ptrdiff_t;
   using value_type        = Key;
-  using pointer           = value_type*;
-  using reference         = value_type&;
+  using pointer           = const value_type*;
+  using reference         = const value_type&;
 
   using map_iterator = typename std::map<Key,Value>::const_iterator;
 


jeff-cohere (dismissed):

src/ekat/std_meta/ekat_std_map_key_iterator.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+  map_key_iterator ( ) : map_iterator ( ) { };
+  map_key_iterator ( map_iterator it_ ) : map_iterator ( it_ ) { };
+
+  Key *operator -> ( ) { return ( Key * ) &( map_iterator::operator -> ( )->first ); }
+  Key operator * ( ) { return map_iterator::operator * ( ).first; }
+};
+
+template<typename Key, typename Value>
+class map_key_const_iterator<std::map<Key,Value>> final : public std::map<Key, Value>::const_iterator
```
jeff-cohere, line 40: This is probably too nitpicky, but: do you want to `public`ly inherit from `std::map<K,V>::const_iterator` if you're trying not to expose the value content? Not a big deal. :-P

bartgol: Well, the class overloads `op*` and `op->`, which effectively 'hides' the base class operators, no? So one should not be able to get the std::pair stored in the map from this iterator... Unless they do `std::map<K,V>::const_iterator it = the_map_key_const_iter`. No?

bartgol: And even if I inherit privately, they can always do the aforementioned hack, I think.

jeff-cohere: All true. Looks good to me.

jeff-cohere (dismissed):

jeff-cohere (approved):

jeff-cohere (approved):

PR #158, opened 2021-10-29
test-launcher: more updates

jgfouca:
1) Refactor argument handling to avoid conflicts with options for
the executable. Use -- instead of -e.
2) Do not use taskset on GPU machines. taskset is for specifying CPUs
but the incoming data from the ctest resource file is specific to GPUs,
so it makes no sense to use that same data for tasksetting.

## Motivation

This is part of the effort to get things working on Lassen.

## Testing

./scripts-tests -f -m mappy TestTestAllScream.test_mpirank_and_thread_spreading

### DIFF
--- a/bin/test-launcher
+++ b/bin/test-launcher
@@ -75,20 +75,19 @@ def run_cmd(cmd, input_str=None, from_dir=None, verbose=None, dry_run=False,
 def parse_command_line(args, description):
 ###############################################################################
     parser = argparse.ArgumentParser(
-        usage="""\n{0} <executable>
+        usage="""\n{0} -- <executable> [<exe-args>]
 OR
 {0} --help
 
 \033[1mEXAMPLES:\033[0m
     \033[1;32m# Lightweight python wrapper for a test that handles thread binding.\033[0m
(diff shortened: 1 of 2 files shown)


bin/test-launcher
```diff
@@ -75,31 +75,31 @@ def run_cmd(cmd, input_str=None, from_dir=None, verbose=None, dry_run=False,
...
     return parser.parse_known_args(args[1:])
 
 ###############################################################################
-def run_test(executable,buffer_output,print_omp_affinity,exec_args):
+def run_test(buffer_output,print_omp_affinity,exec_args):
 ###############################################################################
 
-    print("execname: {}".format(executable))
+    expect(exec_args, "Nothing to run?")
+    print("execname: {}".format(exec_args[1]))
```
bartgol, line 102: I think I added this when I was debuggin. We can remove it, I think. `run_cmd` will already take care of printing what we're running.

bartgol (approved):
I have a cleanup request, but I'm gonna approve anyways.

bartgol (approved):

PR #160, opened 2021-11-05
Add support for reusing executables to EkatCreateUnitTest macro

jgfouca:
This will allow us to avoid having to unnecessarily rebuild multiple identical executables from the same source.

This PR is best viewed by remove whitespace-only changes. A large block of code needed to be indented but most of it was not otherwise changed.

## Testing
Tested by changing some of the resource tests on the scream side to use this feature and it appeared to work exactly as intended.

### DIFF
--- a/cmake/EkatCreateUnitTest.cmake
+++ b/cmake/EkatCreateUnitTest.cmake
@@ -2,8 +2,8 @@ include(CMakeParseArguments) # Needed for backwards compatibility
 include(EkatUtils) # To check macro args
 
 # This function takes the following mandatory arguments:
-#    - target_name: the name of the executable
-#    - target_srcs: a list of src files for the executable
+#    - test_name: the base name of the test. We create an executable with this name unless TEST_EXE is passed.
+#    - test_srcs: a list of src files for the executable. An empty list should be provided if TEST_EXE is passed.
 #      Note: no need to include ekat_catch_main; this macro will add it
 # and the following optional arguments (to be passed as ARG_NAME "ARG_VAL")
 #    - MPI_RANKS: the number of mpi ranks for the test.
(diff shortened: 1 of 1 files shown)


cmake/EkatCreateUnitTest.cmake
```diff
@@ -68,73 +69,78 @@ function(EkatCreateUnitTest target_name target_srcs)
...
-  #---------------------------#
-
-  # Set link directories (must be done BEFORE add_executable is called)
-  # NOTE: CMake 3.15 adds 'target_link_directories', which is superior (does not pollute other targets).
-  #       We should switch to that as soon as we can assume CMAKE_VERSION >= 3.15.
-  if (ecut_LIBS_DIRS)
-    link_directories("${ecut_LIBS_DIRS}")
-  endif()
-  add_executable (${target_name} ${target_srcs})
+  if (NOT ecut_TEST_EXE)
```
bartgol, line 72: The code inside the `if` could be extracted into a separate macro (something like `EkatCreateUnitTestExec`, which EkatCreateUnitTest can call) so that in the app cmake we can do
```
# Create executable
EkatCreateUnitTestExec (my_exec my_exec.cpp LIBS "scream_share")

# Create tests
EkatCreateUnitTest(my_test_1 TEST_EXEC my_exec EXE_ARGS input_file1)
...
EkatCreateUnitTest(my_test_2 TEST_EXEC my_exec EXE_ARGS input_fileN)
```

jgfouca: @bartgol , the usage based on the current state of the PR would be:

```
# Create tests
EkatCreateUnitTest(my_test_1 my_exec.cpp scream_share EXE_ARGS input_file1)
...
EkatCreateUnitTest(my_test_2 "" "" TEST_EXEC my_exec_1 EXE_ARGS input_fileN)
```

Dunno if this is better or not. It is a bit shorter and keeps the cmake unit test API at one function, so I'm leaning towards this approach.

bartgol (approved):

PR #163, opened 2021-11-10
For Catch2 unit tests, suppress colored output in ctest tests

bartgol:
## Motivation

While the colors are nice when running tests manually, they create awful strings when the output is redirected to a file (like the LastTest.log file that ctest generates). E.g.,

```
^[[1;31m^[[0m^[[1;33m^[[0m^[[1;32m===============================================================================^[[0m
^[[1;32mAll tests passed^[[0m
```
In the macro `EkatCreateUnitTest`, we suppress colors in the test output if we are linking `ekat_catch_main`. Notice that this only affects output when running `ctest`. If one runs the executable manually, the colors are still used (unless the `--use-colour no` option is used).

## Testing

I manually verified it works.

### DIFF
--- a/cmake/EkatCreateUnitTest.cmake
+++ b/cmake/EkatCreateUnitTest.cmake
@@ -254,6 +254,10 @@ function(EkatCreateUnitTest test_name test_srcs)
   else()
     set(invokeExec "./${target_name}")
   endif()
+  if (NOT ecut_EXCLUDE_MAIN_CPP)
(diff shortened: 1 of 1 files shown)


jgfouca:
@bartgol , I agree this is an issue. Most common linux tools that offer colorful output have solved this problem by checking to see if the output is going to the terminal. I think we should do the same since the color is nice to have for developers running in their terminals.

I can do some digging on this if you don't have time.

jgfouca (approved):

PR #166, opened 2021-11-22
Split EkatCreateUnitTest in two macros

bartgol:
## Motivation

A recent change made it possible to create a unit test with an existing executable. However, the user still had to pass an empty argument for the exec sources. This PR fixes this, by splitting the existing macro into two macros: one to create the executable (`EkatCreateUnitTestExec`), and the other to create the tests (`EkatCreateUnitTestFromExec`). The old macro `EkatCreateUnitTest` is still available (but does not support the `TEST_EXE` optional arg anymore), and simply calls the other two in sequence (plus some logic to split the optional arg among the two macros).

To create structurally different tests with the same exec, one can now do
```
EkatCreateUnitTestExec(my_exe src1 [src2..] <options_for_building_exec> )

EkatCreateUnitTestFromExec (test1 my_exe <options_for_test1>)
EkatCreateUnitTestFromExec (test2 my_exe <options_for_test2>)
```
or even
```
EkatCreateUnitTest (test1 src1 [src2] <options_for_building_exec> <options_for_test1>)
EkatCreateUnitTestFromExec (test2 test1 <options_for_test2>)
```
(`EkatCreateUnitTest` makes the exec name coincide with the root test name).

## Testing

I manually verified that the two macros work as expected. All existing EKAT tests use the old macro, which is now implemented in terms of the two new ones, making them "tested".

### DIFF
--- a/cmake/EkatCreateUnitTest.cmake
+++ b/cmake/EkatCreateUnitTest.cmake
@@ -1,11 +1,131 @@
 include(CMakeParseArguments) # Needed for backwards compatibility
 include(EkatUtils) # To check macro args
 
+# Note: we hace to set this variable here, so CMAKE_CURRENT_LIST_DIR gets the
+#       directory of this file. If we did it inside the function, it would get
+#       the directory from where the function is called
+set (CATCH_INCLUDE_DIR ${CMAKE_CURRENT_LIST_DIR}/../extern/catch2/include)
+
 # This function takes the following mandatory arguments:
-#    - test_name: the base name of the test. We create an executable with this name unless TEST_EXE is passed.
-#    - test_srcs: a list of src files for the executable. An empty list should be provided if TEST_EXE is passed.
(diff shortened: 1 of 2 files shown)


bartgol:
Note: the old macro could have been implemented easily as

```
macro (EkatCreateUnitTest test_name test_srcs)
  EkatCreateUnitTestExec (${test_name} ${test_srcs} ${ARGN})
  EkatCreateUnitTestFromExec (${test_name} ${test_name} ${ARGN})
endmacro ()
```
However, this would cause lots of cmake warning from the inner macros, which would receive many more args than what they expect (they would both receive both add-exec and add-test phase args). I could have dropped the check on the input args, but it seemed unsafe, so I had to add some logic to split the input lists of optional args into lists for exec-creation and tests-creation phases.

jgfouca (approved):

PR #168, opened 2021-11-24
Reduce duplication of argument lists for create unit test functions

jgfouca:
Use module-wide variables to store argument lists for the two functions that implement ekat's CreateUnitTest function.

## Motivation

Reduces lots of duplication.

## Testing

Did a full build/ctest cycle. Ran a few of the more-complex tests with `-V` to confirm that various options were passed successfully.


cmake/EkatUtils.cmake
```diff
@@ -113,3 +113,19 @@ macro (EkatDisableAllWarning targetName)
 
 endmacro (EkatDisableAllWarning)
 
+function(separate_cut_arguments prefix options oneValueArgs multiValueArgs return_varname)
```
bartgol, line 116: Since you plan to set stuff in the parent scope, you could use a macro instead of a function, and do away with the `result` temp var.

bartgol (approved):
Thanks, looks much cleaner. I have a suggestion for the helper function, but it's not really a big deal.

PR #171, opened 2021-12-17
Add default constructor and setup() to WSM

tcclevenger:
Add an empty constructor and a setup() function to WSM

## Motivation
In SCREAM AD, we were having to construct a WSM object inside the `run_impl()` of various processes, which often was a significant amount of time. This is unnecessary since at initialization we have all the info needed for the WSM construction. To make the WSM a class variable we need an empty constructor with a setup method.

## Testing
I added a test that makes sure the empty constructor+setup() correctly sets size and can take and release blocks

## Additional Information
I needed to add an empty constructor for `TeamUtils` and `TeamUtilsCommonBase` as well.

### DIFF
--- a/src/ekat/ekat_workspace_impl.hpp
+++ b/src/ekat/ekat_workspace_impl.hpp
@@ -15,23 +15,16 @@ namespace ekat {
 
 template <typename T, typename D>
 WorkspaceManager<T, D>::WorkspaceManager(int size, int max_used, TeamPolicy policy,
-                                         const double& overprov_factor) :
-  m_tu(policy, overprov_factor)
+                                         const double& overprov_factor)
 {
-  compute_internals(size, max_used);
-  m_data = decltype(m_data) (Kokkos::ViewAllocateWithoutInitializing("Workspace.m_data"),
-                             m_max_ws_idx, m_total*m_max_used);
-  init_all_metadata(m_max_ws_idx, m_max_used);
+  setup(size, max_used, policy, overprov_factor);
 }
 
 template <typename T, typename D>
 WorkspaceManager<T, D>::WorkspaceManager(T* data, int size, int max_used,
-                                         TeamPolicy policy, const double& overprov_factor) :
-  m_tu(policy, overprov_factor)
+                                         TeamPolicy policy, const double& overprov_factor)
 {
-  compute_internals(size, max_used);
-  m_data = decltype(m_data) (data, m_max_ws_idx, m_total*m_max_used);
-  init_all_metadata(m_max_ws_idx, m_max_used);
+  setup(data, size, max_used, policy, overprov_factor);
 }
 
 template <typename T, typename D>
--- a/src/ekat/ekat_workspace_impl.hpp
+++ b/src/ekat/ekat_workspace_impl.hpp
@@ -62,6 +62,8 @@ int WorkspaceManager<T, D>::get_total_bytes_needed(int size, int max_used, TeamP
 template <typename T, typename D>
 void WorkspaceManager<T, D>::report() const
 {
+  EKAT_ASSERT_MSG (is_initialized, "Error! WorkspaceManager not yet inited.\n");
+
 #ifndef NDEBUG
   auto host_num_used   = Kokkos::create_mirror_view(m_num_used);
   auto host_high_water = Kokkos::create_mirror_view(m_high_water);
@@ -128,6 +130,8 @@ void WorkspaceManager<T, D>::setup (int size, int max_used, TeamPolicy policy,
   m_data = decltype(m_data) (Kokkos::ViewAllocateWithoutInitializing("Workspace.m_data"),
                              m_max_ws_idx, m_total*m_max_used);
   init_all_metadata(m_max_ws_idx, m_max_used);
+
+  is_initialized = true;
 }
 
 template <typename T, typename D>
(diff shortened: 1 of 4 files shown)


tcclevenger:
@bartgol Is mappy not setup for the AT? Seems like the old module set is no longer relevant.

I ran all of `ctest` on mappy manually with 
```
module purge
module load sems-cmake/3.18.4
module load sems-gcc/10.1.0
module load sems-openmpi/4.0.5
module load sems-git/2.11.1
module load sems-netcdf-fortran/4.5.3
module load sems-netcdf-c/4.7.3
```
and
```
cmake   
  -D CMAKE_BUILD_TYPE:STRING=DEBUG \
  -D CMAKE_C_COMPILER:STRING=mpicc \
  -D CMAKE_CXX_COMPILER:STRING=mpicxx \
  -D CMAKE_Fortran_COMPILER:STRING=mpifort   
  -D EKAT_ENABLE_TESTS:BOOL=ON  \
  -D EKAT_TEST_DOUBLE_PRECISION:BOOL=ON  \
  -D EKAT_TEST_SINGLE_PRECISION:BOOL=ON   \
  -D EKAT_DISABLE_TPL_WARNINGS:BOOL=ON  \
  -D YAML_CPP_INSTALL_PREFIX_SUFFIX:STRING=yamlcpp  \
  -D Kokkos_ENABLE_OPENMP:BOOL=ON   \
  /path/to/ekat
```
and everything passed.

jgfouca:
@tcclevenger , the mappy GitHub issues have been resolved so we should expect it to work again for AT purposes. It looks like it is still having problems with the new modules which is something we (I) need to fix. Let's use this PR as an opportunity to get the AT working again.

jgfouca:
@tcclevenger , OK, this is an EKAT PR which is why we are seeing the stale modules. Fix coming.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -433,10 +435,22 @@ template <typename ValueType, typename ExeSpace = Kokkos::DefaultExecutionSpace>
...
 {
  public:
+  TeamUtils() = default;
+
   template <typename TeamPolicy>
   TeamUtils(const TeamPolicy& policy, const double& = 1.0) :
     TeamUtilsCommonBase<ValueType, ExeSpace>(policy)
   { }
+
+  TeamUtils& operator= (const TeamUtils& src)
```
jgfouca, line 445: Isn't this the default implementation? I assume the base class operator= is invoked in the default?

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -493,6 +521,21 @@ class TeamUtils<ValueType,Kokkos::Cuda> : public TeamUtilsCommonBase<ValueType,K
     }
   }
 
+  TeamUtils& operator= (const TeamUtils& src)
```
jgfouca, line 524: This also looks like default.

tests/kokkos/workspace_tests.cpp
```diff
@@ -129,6 +129,58 @@ static void unittest_workspace()
         {&ws1, &ws2, &ws3, &ws4});
     });
   }
+
+  {
+    // Test constructing the WSM with empty constructor
+    const int slot_length = 17;
```
jgfouca, line 135: Thanks for adding a test!

src/ekat/ekat_workspace_impl.hpp
```diff
@@ -124,6 +125,29 @@ void WorkspaceManager<T, D>::report() const
 #endif
 }
 
+template <typename T, typename D>
+void WorkspaceManager<T, D>::setup (int size, int max_used, TeamPolicy policy,
+                                    const double& overprov_factor)
+{
```
jgfouca, line 131: Can the constructors call setup to reduce a bit of duplication?

jgfouca (changes requested):
Some minor questions, requests. Overall looks good.

src/ekat/kokkos/ekat_kokkos_utils.hpp
bartgol: I agree, if you need a "default" behavior, you should simply declare `operator=(..) = default`. The defaulted impl, implicitly calls the base class op=.

src/ekat/ekat_workspace_impl.hpp
bartgol: Agreed

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -382,6 +382,8 @@ class TeamUtilsCommonBase
 protected:
   int _team_size, _num_teams, _max_threads, _league_size;
 
+  TeamUtilsCommonBase() = default;
```
bartgol, line 385: Now that default ctor is legal, it is possible that someone forgets to call setup before using the class.

Do you think it's worth it to add some checks (maybe just in debug mode) every time get methods are called? Something like

```
int get_blah () const {
  EKAT_REQUIRE_MSG (_team_size>0, "Error! TeamUtils not yet inited.\n");
  return blah;
}
```
And also change `int _team_size` to `int _team_size=0,...`, to _guarantee_ that it's inited to 0 (and trigger the check error if no setup method is called).

bartgol (commented):
I agree with Jim's comments. Additionally, I have another comment/question.

src/ekat/kokkos/ekat_kokkos_utils.hpp
tcclevenger: Yeah, I agree. I'll add checks.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -521,21 +501,6 @@ class TeamUtils<ValueType,Kokkos::Cuda> : public TeamUtilsCommonBase<ValueType,K
     }
   }
 
-  TeamUtils& operator= (const TeamUtils& src)
```
bartgol: If you are ok with the default op=, then I would suggest making it clear, by declaring op= defaulted.

tcclevenger: Ok yeah, I was about to ask that question. What is the difference in eliminating it vs. adding the default declaration?

bartgol: Morally, none. The compiler auto-generates the default op= anyways. However, imho there are a couple of benefits adding an explicitly defaulted one. The most important one is that the reader _knows_ you are ok with the default impl. The second is more personal: if down the road your class becomes non-copyable (e.g., it contains a member that is non-copyable), you would get a slightly better error message (at least with gcc):

without defaulted op=:
```
error: use of deleted function ‘B& B::operator=(const B&)’
   32 |   b2 = b1;
      |        ^~
note: ‘B& B::operator=(const B&)’ is implicitly deleted because the default definition would be ill-formed:
   23 | struct B {
      |        ^
```
with defaulted op=:
```
error: use of deleted function ‘B& B::operator=(const B&)’
   32 |   b2 = b1;
      |        ^~
note: ‘B& B::operator=(const B&)’ is implicitly deleted because the default definition would be ill-formed:
   24 |   B& operator= (const B&) = default;
      |      ^~~~~~~~
```
Not a big difference, but to me, it's always clearer when the compiler points to a problematic method (like in the second case) rather than to the class name (like in the first case).

bartgol (approved):

tcclevenger:
@bartgol The same is true of the WSM (default constructor available). I added some of its own checks that it is setup. Let me know if you would like changes (maybe checking an already defined local `int` instead of adding a new bool?)

bartgol (approved):

jgfouca (approved):

PR #172, opened 2021-12-22
Add LICENSE file with copyright notice and boost license assertion

bartgol:
## Motivation

EKAT was distributed without a copyright/license, which means people were technically not able to use EKAT in their project. This PR adds both copyright notice (Sandia's), as well as boost license assertion, allowing people to use the library in their projects.

## E3SM Stakeholder Feedback

Needed in other projects.

## Testing

Not needed.


jeff-cohere (approved):

bartgol:
Merging without AT run, since this PR only touches docs, and the AT would stall anyways due to weaver offline.

PR #173, opened 2022-01-05
Some meta-utils for type-list related metaprogramming

bartgol:
## Motivation

This kind of utilities are useful when handling several instantiation of a type, minimizing code duplication. For instance, I'm trying to use them in scream, to handle views of different arithmetic type in a structured way. One can do stuff like:
```
using scalars = TypeList<int,float,double>;
template<typename ST>
using view1d = KokkosTypes<MyDevice>::view_1d<ST>;
using views = ApplyTemplate<view_1d,scalars>::type; // -> TypeList<view1d<int>,view1d<float>,view1d<double>>;

using map_t = TypeMap<scalars,views>;
map_t my_views;
...
my_func(my_views.at<int>());
my_func(my_views.at<float>());
my_func(my_views.at<double>());
```
The last chunk can be done more generically with the `TypeListFor` util:
```
TypeListFor<scalars>([&](auto t){
  using T = decltype(t);
  my_func(my_views.at<T>());
});
```
which allows to add a new scalar type limiting the places in the code where an update is needed. In particular, all places that perform automatic operations on a bunch of different templated instantiations, without really caring about how many there are, or what the template instantiation is, can benefit from an agnostic implementation, which won't change if we add more supported types.

Basically, these meta-utility allow to have tuple-like structure, but access them by a type name rather than an index.

## E3SM Stakeholder Feedback

This is making my life easier in a refactor work in SCREAM. Nevertheless, I am putting the PR on hold until I can fully verify that it works well in scream, and that all pieces are indeed needed. Once that's verified (or if another repo/pr could benefit from this work), I will 'unlock' this pr.

## Testing

I added a test for all meta-utilities. The only things I could not easily test are the static asserts correctness. I could not figure out a simple _run-time_ way of checking that static asserts fire when the template arguments are bad (the only way would be to use a helper cpp file, and try to compile it). However, I did test their correctness by hand.

### DIFF
--- a/src/ekat/util/ekat_meta_utils.hpp
+++ b/src/ekat/util/ekat_meta_utils.hpp
@@ -14,7 +14,7 @@ using TypeList = std::tuple<Ts...>;
 template<typename TL>
 struct type_list_size : std::tuple_size<TL> {};
 
-// Cat two type lists
+// Concatenate two type lists
 template<typename TL1, typename TL2>
 struct CatTypeLists;
 
--- a/src/ekat/util/ekat_meta_utils.hpp
+++ b/src/ekat/util/ekat_meta_utils.hpp
@@ -1,0 +1,165 @@
+#ifndef EKAT_META_UTILS_HPP
+#define EKAT_META_UTILS_HPP
+
+#include <tuple>
+
+namespace ekat
+{
+
+// A parameter pack. To be used for template specializations
+template<typename... Ts>
+struct TypeList;
+
+template<typename T>
+struct TypeList<T> {
+  using head = T;
+  static constexpr int size = 1;
+};
+
+template<typename T, typename... Ts>
+struct TypeList<T,Ts...> {
+private:
+  using tail = TypeList<Ts...>;
+public:
+  using head = T;
+  static constexpr int size = 1 + tail::size;
+};
+
+// Cat two packs
+template<typename P1, typename P2>
+struct CatLists;
+
+template<typename... T1s, typename... T2s>
+struct CatLists<TypeList<T1s...>,TypeList<T2s...>> {
+  using type = TypeList<T1s...,T2s...>;
+};
+
+// Find position of first occurrence of type T in a TypeList, store in 'pos'.
+// If not found, set pos = -1.
+template<typename T, typename TypeList>
+struct FirstOf;
+
+template<typename T, typename H>
+struct FirstOf<T,TypeList<H>> {
+  static constexpr int pos = std::is_same<H,T>::value ? 0 : -1;
+};
+
+template<typename T, typename H, typename... Ts>
+struct FirstOf<T,TypeList<H,Ts...>> {
+private:
+  static constexpr int tail_pos = FirstOf<T,TypeList<Ts...>>::pos;
+public:
+  static constexpr int pos = std::is_same<H,T>::value ? 0 : (tail_pos>=0 ? 1+tail_pos : -1);
+};
+
+// Check if a TypeList contains unique types
+template<typename P>
+struct UniqueTypeList;
+
+template<typename T>
+struct UniqueTypeList<TypeList<T>> : std::true_type {};
+
+template<typename T, typename... Ts>
+struct UniqueTypeList<TypeList<T,Ts...>> {
+  static constexpr bool value = FirstOf<T,TypeList<Ts...>>::pos==-1 &&
+                                UniqueTypeList<TypeList<Ts...>>::value;
+};
+
+// Access N-th entry of a pack
+template<typename P, int N>
+struct AccessList;
+
+template<typename T, typename... Ts, int N>
+struct AccessList <TypeList<T,Ts...>,N> {
+private:
+  using pack = TypeList<T,Ts...>;
+  using tail = TypeList<Ts...>;
+  static_assert (N>=0 && N<pack::size, "Error! Index out of bounds.\n");
+
+public:
+  using type = typename AccessList<tail,N-1>::type;
+};
+
+template<typename T, typename... Ts>
+struct AccessList <TypeList<T,Ts...>,0> {
+  using type = T;
+};
+
+// A map of types
+template<typename KP, typename VP>
+struct TypeMap;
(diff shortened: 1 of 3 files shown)


bartgol:
@jeff-cohere This PR passes on my machine, so I added you as reviewer even before the AT runs, since mappy has been flaky, and I don't want to wait for it to come back online before getting any feedback.

I probably need to add some more inline documentation to `ekat_meta_utils.hpp`.

bartgol:
Had to rebase on master, to get the jenkins scripts folder in.

src/ekat/util/ekat_meta_utils.hpp
```diff
@@ -0,0 +1,235 @@
...
+
+template<typename T, typename H, typename... Ts>
+struct FirstOf<T,TypeList<H,Ts...>> {
+private:
+  static constexpr int tail_pos = FirstOf<T,TypeList<Ts...>>::pos;
+public:
+  static constexpr int pos = std::is_same<H,T>::value ? 0 : (tail_pos>=0 ? 1+tail_pos : -1);
+};
+
+// Check if a TypeList contains unique types
```
jeff-cohere, line 56: If this is supposed to be a meta-predicate, what do you think about sticking the word `Is` in there somewhere? Like `IsUniqueTypeLIst`?

src/ekat/util/ekat_meta_utils.hpp
```diff
@@ -0,0 +1,235 @@
...
+
+template<typename T, typename... Ts>
+struct UniqueTypeList<TypeList<T,Ts...>> {
+  static constexpr bool value = FirstOf<T,TypeList<Ts...>>::pos==-1 &&
+                                UniqueTypeList<TypeList<Ts...>>::value;
+};
+
+// Access N-th entry of a pack
+template<typename P, int N>
+struct AccessList;
```
jeff-cohere, line 71: Maybe `AccessTypeList`? Just a thought.

jeff-cohere: Or `AccessType`, even.

jeff-cohere:
C++ is such a strange language. It has really good library support in some ways, and in others it really makes you do everything. I feel like this PR has stuff in it that should be better supported out of the box, but for some reason the standards committee seems more interested in creating countless ways of doing easier things.

Anyway, let's do it! I think we will have to document this outside of the code itself, as templated code alone isn't really well supported by doc tools, and metaprogramming stuff is even less so.

jeff-cohere (approved):

jeff-cohere:
Looks like mappy had a network issue retrieving a submodule.

src/ekat/util/ekat_meta_utils.hpp
```diff
@@ -0,0 +1,235 @@
...
+  static constexpr int size = keys::size;
+private:
+
+  template<bool has_V_type, typename V>
+  typename std::enable_if<has_V_type,bool>::type
+  has_v_impl (const V& v, const self& map) const {
+    bool found = false;
+    TypeListFor<keys>([&](auto t){
+      using key_t = decltype(t);
+      const auto& value = map.at<key_t>();
```
jeff-cohere, line 217: This line seems to be tripping up Mappy's C++ compiler: 

```
In file included from /ascldap/users/e3sm-jenkins/jenkins-ws/workspace/EKAT_PullRequest_Autotester_Mappy/213/ekat-src/tests/utils/meta_utils.cpp:3:
/ascldap/users/e3sm-jenkins/jenkins-ws/workspace/EKAT_PullRequest_Autotester_Mappy/213/ekat-src/src/ekat/util/ekat_meta_utils.hpp: In lambda function:
/ascldap/users/e3sm-jenkins/jenkins-ws/workspace/EKAT_PullRequest_Autotester_Mappy/213/ekat-src/src/ekat/util/ekat_meta_utils.hpp:217:39: error: expected primary-expression before > token
  217 |       const auto& value = map.at();
      |                                       ^
/ascldap/users/e3sm-jenkins/jenkins-ws/workspace/EKAT_PullRequest_Autotester_Mappy/213/ekat-src/src/ekat/util/ekat_meta_utils.hpp:217:41: error: expected primary-expression before ) token
  217 |       const auto& value = map.at();
```

jeff-cohere: That said, I am not sure why line 217 reported in the error message doesn't include the template parameter for `at`.

bartgol: Really strange. I saw some (somewhat) similar errors while developing this branch, with the compiler reporting errors at seemingly unrelated lines, or with lines looking slightly different. I wonder if the heavy meta-programming is flipping out the compiler...

I'll manually build on mappy, and find the bug.

bartgol:
> C++ is such a strange language. It has really good library support in some ways, and in others it really makes you do everything. I feel like this PR has stuff in it that should be better supported out of the box, but for some reason the standards committee seems more interested in creating countless ways of doing easier things.

I was really bummed when I realized that a "type-map" is not in C++14 or 17 (that I know of). I don't even check C++20, cause god knows when we'll get to use that (and the language there has a big C++11-like evolution jump, so I'm not sure if I'm looking forward to that).

jeff-cohere (commented):
Nice work!

jeff-cohere (dismissed):

src/ekat/util/ekat_meta_utils.hpp
```diff
@@ -7,34 +7,25 @@
...
-  using head = T;
-  static constexpr int size = 1 + tail::size;
-};
+template<typename TL>
+struct type_list_size : std::tuple_size<TL> {};
 
-// Cat two packs
-template<typename P1, typename P2>
-struct CatLists;
+// Cat two type lists
```
jeff-cohere, line 17: Maybe the comment here could say "Concatenate" instead of "Cat", just to be clear.

jeff-cohere (approved):
Great. Just one minor suggestion.

jeff-cohere (approved):

PR #174, opened 2022-01-06
Manage jenkins jobs with repo scripts similar to scream

jgfouca:
Also, leverage utilities from SCREAM to make generalized jenkins script.

## Motivation
Recent churn in the SEMS modules revealed that EKAT was doing its own env loading, independent of what is in SCREAM. As I did some work to ensure that EKAT uses the SCREAM env in AT testing, I saw more opportunities to make the entire EKAT AT system a bit more like SCREAM's, with generalized (instead of machine-specific) test scripts.

I decided it was not worth taking the plunge and driving the scripting with python; things are much simpler on the EKAT side, so we can stick with bash for now, just using simple python tool to pick up needed python config items.

## Testing
Set up a mock Jenkins workspace and ran the new script; it worked.

Jenkins AT jobs have already been updated to reflect these changes.

### DIFF
--- a/scripts/jenkins/jenkins.sh
+++ b/scripts/jenkins/jenkins.sh
@@ -1,0 +1,7 @@
+#! /bin/bash -xe
+
+export JENKINS_SCRIPT_DIR=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
+DATE_STAMP=$(date "+%Y-%m-%d_%H%M%S")
+
+set -o pipefail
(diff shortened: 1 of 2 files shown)


bartgol (approved):
Nice. Are you going to (or did you already) modify our jenkins jobs to use these scripts?

Edit: probably best to do it after the PR goes in.

jgfouca:
@bartgol 

> Jenkins AT jobs have already been updated to reflect these changes.

bartgol:
> @bartgol
> 
> > Jenkins AT jobs have already been updated to reflect these changes.

LoL! One day I'll learn to read PRs all the way to the last line before commenting/reviewing...

PR #176, opened 2022-01-19
Add reset() for WSM

tcclevenger:
Adds a `WorkspaceManager::reset_internals()`. We need this for using the WSM inside an iterative process where `setup()` occurs in an initialization phase. This new function resets the internal data of the WSM to "start over" working through the allocated data at each iteration.

## Testing
Altered a WSM test case to run over multiple steps.


jgfouca (approved):
Looks good!

PR #177, opened 2022-01-27
Add correct --kokkos args when none provided

tcclevenger:
This accomplishes 2 things:
1. If the user inputs args with no `--kokkos-blah`, assume that the will want the base `--kokkos-num-devices=cudaGetDeviceCount` (on GPU) and `--kokkos-disable-warnings`. These match what was previously done for `initialize_ekat_session` with no args.  
2. Connects the 2 `initialize_ekat_session()` by having one call the other.

## Motivation
Previous, `--kokkos-num-devices` was only set if `initialize_ekat_session()` was called with with no args. This is not the case with the SCREAM standalone tests. The result was, when calling `mpirun -np 4 --map-by ppr:2:socket ./test` on a GPU node with 4 GPUs, all work from each mpi process was sent to GPU1 and GPU2-4 were idle.

### DIFF
--- a/src/ekat/ekat_session.cpp
+++ b/src/ekat/ekat_session.cpp
@@ -7,10 +7,13 @@
 #include <vector>
 
 namespace ekat_impl {
-// Since we're initializing from inside a Fortran code and don't have access to
-// char** args to pass to Kokkos::initialize, we need to do some work on our
-// own. As a side benefit, we'll end up running on GPU platforms optimally
-// without having to specify --kokkos-ndevices on the command line.
+// If we initialize from inside a Fortran code, we don't have access to
+// char** args to pass to Kokkos::initialize. In this case we need to do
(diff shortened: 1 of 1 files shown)


bartgol (approved):
To be clear, if the user does specify a kokkos arg, but doesn't specify `--kokkos-num-devices`, then they're on their own, right? They will still get the GPU0 overload, but we don't want to bother fixing that up. If you specify _some_ kokkos, then you know kokkos, and you should be taking care of that.

Sounds good to me.

tcclevenger:
> To be clear, if the user does specify a kokkos arg, but doesn't specify `--kokkos-num-devices`, then they're on their own, right? They will still get the GPU0 overload, but we don't want to bother fixing that up. If you specify _some_ kokkos, then you know kokkos, and you should be taking care of that.
> 
> Sounds good to me.

That is correct.

PR #178, opened 2022-02-03
Some minor mods to unit-test-related cmake functions

bartgol:
## Motivation

Sometimes, Ekat generate tests names that are a tad longer than needed, due to `_np1_omp1`. Additionally, it adds "_ut" (which stands for "unit test"), but it might be redundant if the test name provided by the user is something like "blah_test". This PR does two micro mods:

* Do not add "_ut" to the test name. Let the user decide the test name.
* Do not add _np1 (resp. _omp1) if test is only added for 1 rank (resp. thread). Add suffixes if ran for multiple ranks (resp threads) configurations, or if run with a single configuration with more than 1 rank (resp. thread).

So requesging a mpi/thread serial test named "blah_test" will yield just `blah_test` rather than `blah_test_ut_np1_omp1`.

## Testing

Manually verified generated tests. Before:
```
$ ctest -N
...
  Test #52: wsm_sp_ut_np1_omp7
  Test #53: wsm_sp_ut_np1_omp8
  Test #54: comm_ut_np1_omp1
  Test #55: comm_ut_np2_omp1
  Test #56: comm_ut_np3_omp1
  Test #57: comm_ut_np4_omp1
  Test #58: packs_dp_ut_np1_omp1
  Test #59: packs_sp_ut_np1_omp1
  Test #60: pack_utils_ut_np1_omp1
  Test #61: units_ut_np1_omp1
...
```
After:
```
...
$ ctest -N
  Test #52: wsm_sp_omp7
  Test #53: wsm_sp_omp8
  Test #54: comm_np1
  Test #55: comm_np2
  Test #56: comm_np3
  Test #57: comm_np4
  Test #58: packs_dp
  Test #59: packs_sp
  Test #60: pack_utils
  Test #61: units
...
```


jgfouca (approved):
Will cmake catch us if we try to add multiple tests with the same name? If so, I approve.

bartgol:
Yes, CMake throws a config error if you add multiple tests with the same name, just like for targets.

PR #179, opened 2022-02-04
Bartgol/remove exec launcher from cutfe

bartgol:
## Motivation

The concept (and need) of an exec launcher was really specific to scream. Also the addition of valgrind/cuda-memcheck commands in the call were project-specific. Now that EKAT is getting a larger use, it's best to defer to downstream codes the tuning of all these project-specific settings, while leaving hooks that allow to achieve the same results.

## Content
This PR removes the hard coding of valgrind/cuda-memcheck/test-launcher in the test invoke string. Instead, an additional argument is exposed by EkatCreateUnitTestFromExec: EXEC_LAUNCHER. This variable can be used to pass the string that needs to be pre-pended to the executable, in order to launch it. This can also be used to pass a profiler, or any script use to dispatch the executable.

This removal allowed to get rid of a bit of cmake stuff. The valgrind support to generate suppressions is still available, shipped as a function that the host app can invoke to generate suppression files (with a bit of flexibility on where to put them, or what mpiexec to use).

Also, `EXCLUDE_CATCH_MAIN` (new name for `EXCLUDE_MAIN_CPP`, which was a bit more confusing) is not handled by EkatCreateUnitTestFromExec, to add the `--use-colour no` flag for catch2. The user needs to pass that flag as an exec argument. If the test is created by the 'combo' function EkatCreateUnitTest, then it is automatically added when using catch2 main.

Impact for downstream apps:
- If they relied on test-launcher, they should copy the script from an old ekat version, and put it in their repo.
- If they need to run tests with cuda-memcheck and/or valgrind, they should pass `EXEC_LAUNCHER <tool-name> <tool-options>` to the cmake function that creates the tests.
- If they need to generate valgrind suppressions files, they can call
```
EkatGenerateValgrindSuppressions (MPIEXEC <mpiexec_name> SUPP_FILE_DIR <supp_file_bin_dir_location>)
```
Both arguments are optional, and if no MPIEXEC is passed, only serial suppressions are generated.

## Testing

None added.


bartgol:
Note: if/when approved, I will hold off the merge of this PR, until SCREAM is prepared.

jgfouca (dismissed):
Approved on the condition that all the removed capability is moved into SCREAM.

bartgol:
Yes, of course. I don't want to lose any coverage in scream.

bartgol:
Closing since this is stale and I have no time to revive it.

PR #181, opened 2022-02-07
Use native compilers, and link against MPI

bartgol:
## Motivation

The mpi compilers wrappers (e.g., `mpicxx` or `mpifort`) are simply invoking the native compilers (e.g., `g++` or `gfortran`), adding include paths in the compilation phase, and link flags/libs in the link phase. However, assuming that ekat is built with MPI compiler wrappers, poses certain limitations and adds some configuration complexity. For instance, we had to use `ekat_mpicxx` to wrap mpicxx on GPU, in order to automatically set the env var that mpicxx uses to find the background host compiler (which was nvcc_wrapper for CUDA builds). Another issue is that we made impossible to build EKAT without MPI.

With MPI as a linked lib, we become a "better" cmake project, and allow, through a following PR, to make MPI optional (of course, that will disable some features, like rank-based testing, or the `src/ekat/mpi` source folder).

## E3SM Stakeholder Feedback

I've been asked if EKAT worked without MPI by a downstream app (not SCREAM) developer. They were ultimately able to simply use the mpi wrappers, but at least it showed that separating EKAT from MPI might be useful.

## Testing

All current tests should still work, without any change in configuration settings.


jgfouca (approved):

jgfouca:
Nice! Lots of code removed.

jeff-cohere (approved):
👍

PR #184, opened 2022-02-10
Replaced catch2 header with a submodule.

jeff-cohere:
Updated CMake and testing code as needed.

This PR switches our Catch2 testing library to a submodule pointing at E3SM's fork of the Catch2 library. This allows us to keep Catch2 up to date and to customize it as needed.

Please let me know if I've messed up any CMake things. I usually rely on CMake's external package stuff to build third-party libraries instead of using `add_subdirectory`.


CMakeLists.txt
```diff
@@ -63,6 +64,10 @@ list (APPEND EKAT_TPL_LIBRARIES_INTERNAL ${YAMLCPP_LIBRARIES})
 BuildSpdlog()
 list (APPEND EKAT_TPL_LIBRARIES_INTERNAL ${SPDLOG_LIBRARIES})
 
+# EKAT offers Catch2 for unit testing
+BuildCatch2()
+list (APPEND EKAT_TPL_LIBRARIES_INTERNAL ${CATCH2_LIBRARIES})
```
bartgol, line 69: I was looking at this cmake var. It's used to then set the `EKAT_TPL_LIBRARIES` var, which is only used to link libs to ekat. I think that doing
```
targer_link_libraries (ekat PUBLIC
  ${Kokkos_LIBRARIES}
  ${YAMLCPP_LIBRARIES}
  ${SPDLOG_LIBRARIES}
  ${CATCH2_LIBRARIES}
)
```
might actually be clearer (and easier to maintain) than creating the two cmake vars (one of which is a cache one, for some reason without INTERNAL, so the user could modify it!)

src/ekat/util/ekat_catch_main.cpp
```diff
@@ -1,6 +1,6 @@
 #define CATCH_CONFIG_RUNNER
 
-#include "catch2/catch.hpp"
+#include "catch2/catch_all.hpp"
```
bartgol, line 3: Is this their "all-inclusive" header (corresponding to our old one)? Can we get away with including just a couple of headers?

Also, I _think_ the macro `CATCH_CONFIG_RUNNER` might not be needed with the new version. I remember trying catch as a lib, and being able to not use it. But I might just misremember...

cmake/EkatCreateUnitTest.cmake
```diff
@@ -4,7 +4,7 @@ include(EkatUtils) # To check macro args
 # Note: we have to set this variable here, so CMAKE_CURRENT_LIST_DIR gets the
 #       directory of this file. If we did it inside the function, it would get
 #       the directory from where the function is called
-set(CATCH_INCLUDE_DIR ${CMAKE_CURRENT_LIST_DIR}/../extern/catch2/include)
+set(CATCH_INCLUDE_DIR ${CMAKE_CURRENT_LIST_DIR}/../extern/Catch2/src)
```
bartgol, line 7: I don't think we need this var anymore. Linking against catch2 library, built within this cmake project, should already populate with Catch2 include paths. And I saw that their `src` folder is included in the PUBLIC section of their library.

src/ekat/CMakeLists.txt
```diff
@@ -129,7 +129,7 @@ install (DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/ekat_f90_modules
 add_library(ekat_test_main util/ekat_catch_main.cpp)
 target_link_libraries(ekat_test_main PUBLIC ekat)
 target_include_directories(ekat_test_main PUBLIC
-  $<BUILD_INTERFACE:${EKAT_SOURCE_DIR}/extern/catch2/include>
+  $<BUILD_INTERFACE:${EKAT_SOURCE_DIR}/extern/Catch2/src>
```
bartgol, line 132: Do we need this? That folder is already in the PUBLIC section of the catch 2 lib.

bartgol: Actually, I just noticed that we do not link against the Catch2 target. Any reason not to? I like to leverage cmake's targets property propagation...

src/ekat/util/ekat_test_utils.hpp
```diff
@@ -3,6 +3,8 @@
 
 #include "ekat/ekat_pack.hpp"
 
+#include "catch2/catch_approx.hpp"
```
bartgol, line 6: This is where Catch2 is used inside the ekat library. It would be nice if we moved the little function that needs catch (or compile test utils only with a separate target, maybe `ekat_catch_main`?).

bartgol (commented):
Good work. I have a few suggestions/thoughts.

src/ekat/util/ekat_catch_main.cpp
jeff-cohere: Yes, that's the all-inclusive header, and I didn't have to use anything else for EKAT to build and run its tests.

cmake/EkatCreateUnitTest.cmake
jeff-cohere: This is just me not understanding all the assumptions made by our build system. Happy to remove it.

src/ekat/CMakeLists.txt
jeff-cohere: Don't we? It's included in `EKAT_TPL_LIBRARIES`.

src/ekat/util/ekat_test_utils.hpp
jeff-cohere: Sure, I can look at this tomorrow.

src/ekat/CMakeLists.txt
bartgol: Ah, right. I didn't see it in the line above, but since it's linked to ekat, it's linked to ekat_catch_main as well.

So my first comment stands: we can remove the `target_include_directories` line (one less generator expression to mentally parse!)

src/ekat/util/ekat_test_utils.hpp
bartgol: I'm actually wondering if the utility `catch2_req_pk_sensitive` belongs in ekat, or is "test-specific". The fact that we may not get `==` if PackSIze>1 is very dependent on what kind of operation we did. I think that, among our use cases, only reductions/scans can be nonbfb when done exploiting packs, but the rest should stay bfb.

Perhaps we can punt on the expunction of this routine (and of all Catch2) fromt he ekat target. We may do it in a following PR.

CMakeLists.txt
jeff-cohere: I tried to change this, but I think it messed up the dependency graph for the `ekat` target. Maybe we can fiddle with this in another cleanup PR, like the one in which we might separate out the Catch2 logic.

jeff-cohere:
After switching back to Catch v2, I think most of the changes here become irrelevant. I'm opening a new PR for this.

PR #185, opened 2022-02-10
Added Catch2 as a submodule.

jeff-cohere:
Very few code changes here, as we're still using the single-header version of
Catch2 v2.


bartgol (approved):

bartgol:
@jeff-cohere before merging, would you mind quickly building scream on any machine, to make sure we are setting all the right properties? So we don't have to come back with another pr in ekat to adjust some cmake shenanigans... Thanks!

bartgol (approved):

PR #189, opened 2022-02-17
Add support for dumping parameter lists to YAML file

bartgol:
## Motivation

It might be desirable to dump parameter list into yaml files, to save some results and/or settings used in the simulation.

## E3SM Stakeholder Feedback

I would like to use this in CLDERA (to save data), as well as maybe in SCRAM (to dump default values of parametrizations parameters).

## Testing

Added a test that "loads-dumps-loads", and compares the param lists obtained by the first and second loads (allowing for lines scrambling within each sublist).

## Additional Information
Due to the evil nature of `std::vector<bool>`, so far we saved arrays of bools as `std::vector<int>`. However, that made arrays of ints indistinguishable form array of bools. So I switched internal storage for arrays of bools to `std::vector<char>`. I quickly grepped in ekat and SCREAM, and did not notice any place where we set/get arrays of bools into/from parameter lists (except for ekat parser unit testing), so I suspect this change will have no impact on existing code. But I wanted to point this out, so that we have a log of this change here, for if/when issues arise in downstream codes.


src/ekat/ekat_parameter_list.hpp
```diff
@@ -71,7 +71,7 @@ class ParameterList {
   // NOTE: this *requires* op<< to be overloaded for all the stored parameters.
   //       The code won't crash otherwise, but instead of the parameter, you
   //       will get a message informing you of the lack of op<< overload.
-  void print (std::ostream& out = std::cout, const int indent = 0) const;
+  void print (std::ostream& out = std::cout, const int indent = 0, const int indent_inc = 1) const;
```
jeff-cohere, line 74: Maybe we should briefly describe the indentation-related parameters, just for clarity.

src/ekat/io/ekat_yaml.hpp
```diff
@@ -0,0 +1,16 @@
+#ifndef EKAT_YAML_HPP
+#define EKAT_YAML_HPP
+
+#include "ekat/ekat_parameter_list.hpp"
+#include <string>
+
+namespace ekat {
```
jeff-cohere, line 7: I think we should mention that vectors of booleans are stored as `vector<char>` here, for lack of a better place to warn folks against `vector<bool>`.

jeff-cohere (commented):
Looks good. A few very minor doc suggestions for you to consider.

src/ekat/io/ekat_yaml.cpp
```diff
@@ -180,5 +181,103 @@ void parse_yaml_file (const std::string& fname, ParameterList& params) {
...
+    } else if (params.isType<int>(pname)) {
+      auto i = params.get<int>(pname);
+      out << i;
+    } else if (params.isType<double>(pname)) {
+      auto d = params.get<double>(pname);
+      out << std::showpoint << d;
+    } else if (params.isType<std::string>(pname)) {
+      auto s = params.get<std::string>(pname);
+      out << s;
+    } else if (params.isType<std::vector<char>>(pname)) {
```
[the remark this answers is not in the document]
bartgol: I did think about it. I'll see if I can avoid some code duplication.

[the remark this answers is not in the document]
jeff-cohere: Don't sweat it too much! I'm actually not a DRY zealot at all--I think that the blind application of the principle often harms readability/maintainability.

src/ekat/io/ekat_yaml.cpp
```diff
@@ -1,43 +1,95 @@
...
 #include <yaml-cpp/yaml.h>
 
 #include <sstream>
 #include <fstream>
+#include <iomanip>
 
 namespace ekat {
 
+// These are the values we can parse.
+using values_t     = TypeList<bool,int,double,std::string>;
```
jeff-cohere, line 14: Why the `snake_case` type names here? Just curious.

src/ekat/io/ekat_yaml.cpp
```diff
@@ -1,43 +1,95 @@
...
                  ParameterList& list);
 
-// Helpers
-bool is_bool (const std::string& s) {
+// Check if input string can be interpreted as given type
+template<typename T>
+bool is_type (const std::string& s);
+
+template<>
+bool is_type<bool> (const std::string& s) {
```
jeff-cohere, line 31: Probably not needed, but if we used `strcasecmp` (or whatever C++ equivalent exists, if there is one), we could also allow all the interpolations between `TRUE` and `true`.

bartgol:
@jeff-cohere I abstracted a bit the parse/write routines, so that they can be extended. I have a couple of thoughts on YAML parse/write, which I will put in a separate issue.

jeff-cohere (dismissed):
This is cool. I left a few remarks (consider `strcasecmp` to allow case-insensitive boolean recognition), but there's nothing here preventing us from merging if the tests pass.

bartgol:
> This is cool. I left a few remarks (consider `strcasecmp` to allow case-insensitive boolean recognition), but there's nothing here preventing us from merging if the tests pass.

I can also do `str==CaseInsensitiveString("true")`, using what we have in ekat. I can do that.

jeff-cohere (approved):

src/ekat/io/ekat_yaml.cpp
bartgol: I usually prefer `snake_case` for _local_ type names (i.e., aliases and typedefs). Not sure why I prefer that.

PR #192, opened 2022-02-22
Fix the include directories for EKAT's tpls

bartgol:
## Motivation

The tpls include directories were generating warnings, even if EKAT_DISABLE_TPL_WARNINGS=ON. This was easily fixed by adding the include dirs of the tpl with SYSTEM, in case tpl warnings were to be disabled.

## Testing

None needed.

## Additional Information
This fix would have been much easier if `target_link_libraries` supported the `SYSTEM` option, like `target_include_directories` does. Alas, CMake does not have that.

### DIFF
--- a/cmake/pkg_build/EkatBuildKokkos.cmake
+++ b/cmake/pkg_build/EkatBuildKokkos.cmake
@@ -58,10 +58,14 @@ macro(BuildKokkos)
 
     add_subdirectory(${Kokkos_SOURCE_DIR} ${Kokkos_BINARY_DIR})
 
-    set (Kokkos_INCLUDE_DIRS
-      ${Kokkos_SOURCE_DIR}/core/src
-      ${Kokkos_SOURCE_DIR}/algorithms/src
-      ${Kokkos_BINARY_DIR}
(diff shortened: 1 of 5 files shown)


cmake/pkg_build/EkatBuildKokkos.cmake
```diff
@@ -58,17 +58,40 @@ macro(BuildKokkos)
 
     add_subdirectory(${Kokkos_SOURCE_DIR} ${Kokkos_BINARY_DIR})
 
-    set (Kokkos_INCLUDE_DIRS
-      ${Kokkos_SOURCE_DIR}/core/src
-      ${Kokkos_SOURCE_DIR}/algorithms/src
-      ${Kokkos_BINARY_DIR}
+    #set (Kokkos_INCLUDE_DIRS
```
jeff-cohere, line 61: Do you anticipate that we'll need this commented code in the future?

bartgol: Ah, no, i just forgot to clean up completely. :)

jeff-cohere (dismissed):
Looks good to me.

jeff-cohere (approved):

PR #194, opened 2022-02-23
Fix tridiag tests

### LINKED ISSUE #193: The tridiag tests are not really run, but report pass
Chasing failures in a PR, I found out that the tridiag unit tests using `EXCLUDE_MAIN_CPP` are not really run. Such tests should not have the `--use-colour no` option, since that's only for catch2 executables, but due to buggy logic in splitting exec/test phase options of EkatCreateUnitTest, it's still in.

This can be seen by searching for this line in the ouput:

```
Unexpected arg: --use-colour
```
Unfortunately, the tridiag tests main is itself buggy, and returns zero exit code despite not having run any test.

The last piece of bad news is that, manually removing the `--use-colour no` option from the CTestTestfile.cmake file, and running the tests, I get failures on GPU.

bartgol:
## Motivation

The tridiag tests were not actually run, due to some bugs in both `EkatCreateUnitTest[XYZ]` as well as the argv parsing in `tridiag_tests.cpp`.

## Related Issues

* Closes #193 

## Testing

I manually verified that the tests are now run. Unfortunately, the tests are not passing on GPU. I'm working on tracking down the reason.


bartgol:
The failures were due to some important CUDA flags that got removed during the PR that changed the handling of MPI.

Also, I added some tests that check that invalid flags do trigger a test failure.

jgfouca (approved):

PR #196, opened 2022-02-24
Add property tests for LinInterp

bartgol:
## Motivation

So far, the unit test for LinInterp relied completely on BFB comparison against a "trusted" f90 implementation, specifically created for the unit test. However, there was no property test for the interpolation procedure.

This PR adds the following property tests

* Interpolation to same grid returns the same values of the input.
* Interpolation to grid midpoints gives the average of the input.
* Interpolation to a subsampled grid returns the subsample of the input.
* Interpolation to a a grid whose range is contained in the range of the input grid does not create new extrema.
  
NOTE: all tests are performed only for the TeamThreadRange case.

## E3SM Stakeholder Feedback

In SCREAM, we are tracking down a bug, where an interpolated field ends up being outside the bounds of the input field (despite the grid being the same). These tests allow to rule out bugs in the LinInterp routines.
## Testing

This PR is essentially just adding new tests, as described above.

### DIFF
--- a/tests/algorithm/lin_interp_test.cpp
+++ b/tests/algorithm/lin_interp_test.cpp
@@ -602,7 +602,7 @@ TEST_CASE("lin_interp_monotone", "lin_interp") {
   };
 
   // Mini lambda, to get min-max of a scalarized subview
-  using col_type = decltype(get_col(std::declval<packed_view_2d>(),0));
+  using col_type = decltype(get_col(std::declval<packed_view_2d::HostMirror>(),0));
   auto minmax = [](const col_type& v) -> std::pair<Real,Real> {
     std::pair<Real,Real> minmax {v[0],v[0]};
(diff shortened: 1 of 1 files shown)


bartgol:
@jgfouca Perhaps I do not get the point of the F90 implementation, but if it only served the purpose of giving a "gold reference", maybe we can purge it, now that we have more intrinsic tests (perhaps after adding a few more)?

jgfouca (approved):
Good tests! Thanks!

jgfouca:
@bartgol , 

> but if it only served the purpose of giving a "gold reference", maybe we can purge it

Yes, that was the only reason it got added to the repo. The testing is cheap so I'd be inclined to keep it just so we know when the behavior of the CXX impl changes such that it's not BFB.

jgfouca (approved):

PR #197, opened 2022-02-25
Refactor how EKAT handles compiler flags

bartgol:
## Motivation

The file `EkatSetCompilerFlags.cmake` was almost verbatim copied from Homme's `SetCompilersFlags.cmake`. This file sets a lot of global CMake flags, such as `CMAKE_<LANG>_FLAGS`, which are shared across all targets. Homme used a CMake build system only in a standalone setting, so this worked.

In EKAT, this is different. In particular, I am 99% sure that the pollution of global CMake flags var is the root under the error reported in E3SM-Project/scorpio#460.

Long explanation: on GPU builds, we set some CUDA-specific flags in CMAKE_CXX_FLAGS. If the target being built links against kokkos, some CMake magic ensures that Kokkos' compile launcher is invoked, which in turn calls nvcc_wrapper, which takes care of separating host and CUDA flags. HOWEVER, for tpls that do NOT link against kokkos (like SCORPIO), the kokkos compile launcher is not invoked, so the CUDA flags are passed to mpicxx's backend compiler which does not know what to do with them, and craps out. This causes a call to `find_package (MPI REQUIRED)` to fail (since, without specifying components, `find_package` tries to find all components, one for each currently enabled language).

Besides fixing the issue, this PR brings Ekat into the CMake 3.x era, by completely doing away with setting global vars like CMAKE_CXX_FLAGS.

## E3SM Stakeholder Feedback

This is needed in order to update EKAT in SCREAM, since EKAT did away with `ekat_mpicxx` (rightfully so), but when building in SCREAM, the issue with SCORPIO makes cmake crap out.

## Testing

Currently trying this branch with SCREAM.


bartgol:
Ok, I kind of backstepped here. It is convenient to put flags that are _truly_ general in `CMAKE_<LANG>_FLAGS` and `CMAKE_<LANG>_FLAGS_<BUILDTYPE>` (e.g., warning flags, optimization flags, fp-model flags, etc.). Setting flags exclusively on targets is appealing on projects that have a "pyramid" dependency style between targets, so that we can set basic flags on a "root" target, and have them propagated. However, SCREAM has a few separate "root" targets: ekat, pioc, share_util, homme, cprnc, and possibly more. A target-based approach requires to set basic flags on all of them, which can be tedious (to say the least).

OTOH, flags that are less generic (such as CUDA or OpenMP) are still to be set on a per-target basis.

Notice that the general flags (set in `CMAKE_<LANG>_FLAGS[_<BUILDTYPE]) can be reset upon entering a new scope (like a new subdir), without altering the flags in the parent folder.

This last commit is out of "convenience". I am willing to discuss whether to keep it or revert back to the "add-flags-to-targets-only" version.

jgfouca (approved):

tcclevenger (dismissed):

tcclevenger (approved):

jeff-cohere (dismissed):

bartgol:
I'll test this branch in SCREAM before merging.

bartgol:
This seems to work for SCREAM, at least on my laptop (gcc) and on weaver, when building a standard debug build.

jeff-cohere (approved):

PR #199, opened 2022-03-01
Remove include paths logic when calling BuildEkat()

bartgol:
## Motivation

These lines should be seemingly innocuous, but removing them seems to fix hard-to-debug fails in SCREAM. Also, these lines should actually have zero effect, so we might as well remove them.

## E3SM Stakeholder Feedback

Needed by SCREAM

## Testing

Testing SCREAM with this ekat version now.


jeff-cohere (approved):
Let's see what happens!

bartgol:
Ok, blake passes.

PR #201, opened 2022-03-02
Remove buggy generator expressions for tpls include directories

bartgol:
## Motivation

There were a few issues with these include directories:
    
 - The external packages are already handling build/install interfaces
 - The main goal was to add include dirs as SYSTEM, but since the directories were already added, and SYSTEM dirs are *always* parsed after normal ones, the feature was not achieving its goal.
 - Some (relative) include dirs were swapped (e.g., yaml-cpp/include rather than include/yaml-cpp).
 

## E3SM Stakeholder Feedback

The bug was spotted when linking CLDERA to e3sm. E3SM's cmake build system reported errors of the form
```
CMake Error in cmake/atm/CMakeLists.txt:
  Imported target "cldera-profiling" includes non-existent path

    "/ascldap/users/lbertag/workdir/cldera/cldera-install/gcc/debug/branch/yaml-cpp/include"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.

```

## Testing

I verified that this branch works correctly when EKAT is installed and linked to another app.

### DIFF
--- a/cmake/EkatConfig.cmake.in
+++ b/cmake/EkatConfig.cmake.in
@@ -8,4 +8,7 @@ find_dependency(yaml-cpp PATHS @CMAKE_INSTALL_PREFIX@)
 find_dependency(Kokkos PATHS @CMAKE_INSTALL_PREFIX@)
 find_dependency(spdlog PATHS @CMAKE_INSTALL_PREFIX@)
 
+# MPI is also a dependency, linked by ekat's main lib
+find_dependency(MPI)
+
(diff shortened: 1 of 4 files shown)


jeff-cohere (approved):
Net-negative PRs are pretty great!

bartgol:
If #200 is merged first, not only this PR will be net-negative; it would be absolute-negative. :)

PR #203, opened 2022-03-03
Remove lower bound limiter from LinInterp

bartgol:
## Motivation

The current impl of LinInterp forces a user to specify a (lower bound) limiter for the output, even if the application has no conceptual limit. In case the app does not have an a priori lower bound, the user is forced to set some very low number.

The user can easily implement a limiter after LinInterp is done (like a for-loop) or pad the input `x` and `y` values to implement a limiter. Notice that `y2` greater than `max(y1)` or smaller than `min(y1)` can only happen when extrapolation is needed. In this case, one could do
```
x1padded = [x1(:) very_big_value]
y1padded = [y1(:) y1(end)]
```
to enforece `y2=y1(end)` if `x2>x1(end)`, and similarly for the left of the interval. Ramp extrapolation can also be achieved by fixing the padding for y to a value that gives the desired slope, i.e., to control `(y_far_away-y1(end))/(x_far_away-x1(end))`.

## Testing

No additional testing is necessary

### DIFF
--- a/src/ekat/util/ekat_lin_interp.hpp
+++ b/src/ekat/util/ekat_lin_interp.hpp
@@ -67,7 +67,7 @@ struct LinInterp
   // ------ public API -------
   //
 
-  LinInterp(int ncol, int km1, int km2, Scalar minthresh);
+  LinInterp(int ncol, int km1, int km2);
 
   // Simple getters
   KOKKOS_INLINE_FUNCTION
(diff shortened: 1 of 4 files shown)


jgfouca (approved):

bartgol:
The failure on blake is hard to debug, since I don't get it when I run the tests.

PR #204, opened 2022-03-04
Restore CPP macro-based approach to detect MPI distribution

bartgol:
## Motivation

On some platforms, find_package(MPI) returns MPIEXEC_EXECUTABLE=srun, from which it's impossible to gather info on the backend mpi implementation. At this point, the most reliable approach remains to try to compile a C source file, and check for certain macros. Unfortunately, this rules out the case where a downsream project only enables Fortran, since the try-compile strategy needs a C or CXX mpi compiler. This should not be, imho, too big of a deal, since folks using EKAT are likely interested in using C/CXX.

## E3SM Stakeholder Feedback

SCREAM needs this to fix configuration on some LLNL platforms.

## Testing

Verified on mappy that both openmpi and mpich are identified correctly.


bartgol:
@AaronDonahue can you try to check out this version of ekat in scream on the LLNL platform and verify it works? It should be ok, since it's basically what we had before switching to what I thought was a more portable approach (sigh). Still, better to double check. If it works for you, please, add automerge.

cmake/EkatMpiUtils.cmake
```diff
@@ -1,28 +1,36 @@
...
-            "MPIEXEC_EXECUTABLE is not set. You must call find_package(MPI ..),\n"
-            "*before* calling GetMpiDistributionName (and in the same scope)")
+      "MPI_C_COMPILER and MPI_CXX_COMPILER are not set.\b"
+      "CMake logic to determine the distribution name requires C/CXX mpi compilers.\n"
+      "Please call find_package(MPI COMPONENTS [C|CXX]) *before* calling GetMpiDistributionName (and in the same scope)")
     message ("${MSG}")
     message (FATAL_ERROR "Aborting")
   endif()
-  execute_process(COMMAND ${MPIEXEC_EXECUTABLE} --version
+  execute_process(COMMAND ${COMPILER_NAME} -c ${EKAT_CMAKE_DIR}/TryCompileMPI.cxx
```
jeff-cohere, line 16: Have you ever used CMake's [try_compile](https://cmake.org/cmake/help/latest/command/try_compile.html) command? Just curious. It seems to be intended for this stuff.

jeff-cohere (approved):
Just a minor comment.

cmake/EkatMpiUtils.cmake
bartgol: Doesn't try-compile use `CMAKE_<LANG>_COMPILER` to compile? Right now, it is technically possible to build EKAT using standard compilers, and link MPI to them. Here, however, I want to use the MPI compiler wrappers. I guess I could tell cmake to include mpi when running this specific try compile...

I might switch to try_compile (with MPI include flags), since I suspect that that's more portable. Manually running commands always has the chance of causing shady config errors down the road.

jeff-cohere (approved):

AaronDonahue:
> @AaronDonahue can you try to check out this version of ekat in scream on the LLNL platform and verify it works? It should be ok, since it's basically what we had before switching to what I thought was a more portable approach (sigh). Still, better to double check. If it works for you, please, add automerge.

Confirmed, this fixes the build errors on quartz.

PR #208, opened 2022-03-17
WSM: allow passing refs rather than array of pointers to the take method

bartgol:
## Motivation

This PR is from an _old_ branch that I revived, and it's a quality-of-life improvement for the WorkspaceManager class.

In the following, the first syntax is heavier than the second:
```
   Kokkos::Array<const char*,3> names {"v1","v2","v3"};
   Unmanaged<view_1d> v1,v2,v3;
   
   // Old way
   Kokkos::Array<Unmanaged<view_1d>*,3> ptrs {&v1, &v2, &v3};
   wsm.take_many(names,ptrs);
   
   // New way
   wsm.take_many(names,v1,v2,v3);
```
This PR allows the second form (with a slightly different method name).

## Testing

I changed the wsm test to use `take_many_refs` instead of `take_many`.

## Additional Information
I would _really_ like to actually just do
```
wsm.take_many_refs(v1,v2,v3);
```
since I don't really care about the views names. However, the way I tried to implement it (see [here](https://github.com/E3SM-Project/EKAT/blob/118c6059976cdbfca524ede5feb2516413267d09/src/ekat/ekat_workspace_impl.hpp#L428)), cause the test to crash. I don't understand the wsm metadata stuff well enough to debug, and since it's not a _crucial_ feature, I can just let it go.

### DIFF
--- a/src/ekat/ekat_workspace.hpp
+++ b/src/ekat/ekat_workspace.hpp
@@ -4,6 +4,7 @@
 #include "ekat/util/ekat_arch.hpp"
 #include "ekat/kokkos/ekat_kokkos_utils.hpp"
 #include "ekat/kokkos/ekat_kokkos_types.hpp"
+#include "ekat/std_meta/ekat_std_utils.hpp"
 
 namespace unit_test {
 struct UnitWrap;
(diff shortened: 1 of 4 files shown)


jgfouca (approved):
Nice improvement!

tcclevenger (approved):
@jgfouca Should we apply automerge since Luca is on vacation?

jgfouca:
@tcclevenger , I just merged it manually.

PR #210, opened 2022-03-28
Refactor a bit the Logger class

bartgol:
## Motivation

The logger class had some hard-coded behaviors when it comes to file name. This PR makes the class a bit more flexible. In particular:
- use policies also for MPI-related behaviors, namely, whether to put rank in filename, and whether all ranks should output (as opposed to just rank 0).
- allow to specify file extension
- allow to set/disable format of output (the spdlog default prints messages as "[date time] [log name] [log lvl] <msg>")

## E3SM Stakeholder Feedback

This might be needed in SCREAM (see [this issue](https://github.com/E3SM-Project/scream/issues/1502)).
## Testing

Tests have been updated, and pass on my laptop.


jeff-cohere (commented):
Nice use of policies!

It looks like we have some logging-related (and tridiag solver?) test failures, but I can't see the results. I like this structure, though.

tests/logging/mpi_log_tests.cpp
```diff
@@ -0,0 +1,129 @@
...
+  std::ifstream without_rank_log("without.log");
+  REQUIRE( without_rank_log.is_open() );
+  std::vector<std::string> content;
+  for (int i=0; i<comm.size(); ++i) {
+    std::string line;
+    while (std::getline(without_rank_log,line)) {
+      content.push_back(line);
+    }
+  }
+  REQUIRE (content==std::vector<std::string> (comm.size(),"message"));
```
jeff-cohere, line 59: Here's where one of the tests is failing (on Blake, I believe).

bartgol:
@jeff-cohere The logger fails were due to multiple ranks writing on the same file. Turns out we cannot support that, unless we write a special "mpi-sink", which implements a "semaphore" logic for ranks. Otherwise, when two ranks write on same file, corruption is all but certain.

I rewrote a bit the policies, forcing the `LogAllRanks` policy to use different files for each rank.

I originally added some CPP/CMake logic to be able to use the `filesystem` stuff from `std`. The tedious logic comes from the fact that `std::filesystem` from `<filesystem>` only works if C++17 is enabled, while with C++14 you need to rely on `std::experimental::filesystem` from `<experimental/filesystem>`. But since we do have C++14 as standard, I decided to just remove that logic.

jeff-cohere (approved):
Looks great!

PR #212, opened 2022-03-31
Add utility to parse string representing nested lists

bartgol:
## Motivation

I wrote this for another library, but I can see it being used in the near future in SCREAM (and maybe beyond) to make specifying grouping of entities easier from input file. In particular, I want to group entities in one line in a command line, as follows:

```
My nested list: "[a, b, [c, d, [e]], f, [[g, h], i]]"
```
(i hope i got the parentheses correctly). This PR parses the string "[a, b, [c, d, [e]], f, [[g, h], i]]" and produces a nested ParameterList. Each list contains:

1. "Num Entries": how many entryes are in it. In this example, the outermost list has 4 entries
2. "Depth": how many lists are nested in the current list (counting the current list). So the outermost list has depth 3, while `[g,h]` has depth 1.
3. For i=0,...,NumEntries-1, "Type $i": whether the i-th entry is a "Value" or a "List".
4. For i=0,...NumEntries-1, "Entry $i": if the type is "Value", this is a string, otherwise it is another ParameterList.

So for the above string, the parsing the string, and printing the resulting parameter list would yield the following
```
[a,b,[c,d,[e]],f,[[g,h],i]]:
 Depth: 3
 Entry 0: a
 Entry 1: b
 Entry 3: f
 Num Entries: 5
 Type 0: Value
 Type 1: Value
 Type 2: List
 Type 3: Value
 Type 4: List
 Entry 2:
  Depth: 2
  Entry 0: c
  Entry 1: d
  Num Entries: 3
  Type 0: Value
  Type 1: Value
  Type 2: List
  Entry 2:
   Depth: 1
   Entry 0: e
   Num Entries: 1
   Type 0: Value
 Entry 4:
  Depth: 2
  Entry 1: i
  Num Entries: 2
  Type 0: List
  Type 1: Value
  Entry 0:
   Depth: 1
   Entry 0: g
   Entry 1: h
   Num Entries: 2
   Type 0: Value
   Type 1: Value
```
Notice that the name of the topmost list is the string that was parsed.

## E3SM Stakeholder Feedback

I think this might soon be helpful in SCREAM.

## Testing

Added testing that verify we can correctly parse valid strings, and throw errors for invalid ones.

### DIFF
--- a/src/ekat/util/ekat_string_utils.hpp
+++ b/src/ekat/util/ekat_string_utils.hpp
@@ -48,8 +48,8 @@ std::list<std::string> gather_tokens(const std::string& s,
 
 // Utils to verify/parse a string encoding nested lists,
 // such as '[a,b,[c,d],e]'
-bool validNestedListFormat (const std::string& str);
-ParameterList parseNestedList (std::string str);
+bool valid_nested_list_format (const std::string& str);
+ParameterList parse_nested_list (std::string str);
 
 // Computing similarity index between s1 and s2 using Jaro algorithm
 // For a quick description of the Jaro similarity index, see, e.g.,
--- a/src/ekat/util/ekat_string_utils.hpp
+++ b/src/ekat/util/ekat_string_utils.hpp
@@ -1,6 +1,8 @@
 #ifndef EKAT_STRING_UTILS_HPP
 #define EKAT_STRING_UTILS_HPP
 
+#include <ekat/ekat_parameter_list.hpp>
+
 #include <string>
 #include <vector>
 #include <list>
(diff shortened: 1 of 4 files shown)


src/ekat/util/ekat_string_utils.hpp
```diff
@@ -44,6 +46,11 @@ std::list<std::string> gather_tokens(const std::string& s,
                                      const std::vector<char>& delimiters,
                                      const std::string& atomic = "");
 
+// Utils to verify/parse a string encoding nested lists,
+// such as '[a,b,[c,d],e]'
+bool validNestedListFormat (const std::string& str);
```
jeff-cohere, line 51: Shouldn't these two methods use `snake_case`?

jeff-cohere (commented):
Looks good. Can you just rename the new functions using `snake_case` instead of `lowerCamel`?

jeff-cohere (approved):

PR #214, opened 2022-04-05
Avoid slamming the login node with compilation jobs during jenkins testing

bartgol:
## Motivation

It is not nice to use a lot of the login node resources to build code. Even though EKAT is relatively small, if a few users are doing the same thing, we can quickly clobber the node, slowing everything down for everyone.


jgfouca:
I'm not as familiar with EKAT jenkins infrastructure. Why are we not just submitting the entire job like we do for SCREAM (gather-all-data)?

jgfouca (approved):

PR #215, opened 2022-04-21
Support more types in comm operations, and allow users to extent to new types

bartgol:
## Motivation

Applications may need to perform comm operations with more types, or even user defined derived types. We now provide built-in support for more C++ types (including `short`, `long`, `long long`). Furthermore, the `get_mpi_type` function is now a free function, with no default implementation. Support for types is added via specialization, so the user can support its own types.

## E3SM Stakeholder Feedback

The CLDERA project needs to do reduction of `long long` integers.

## Testing

Added test for all newly added builtin types, as well as verified specialization logic works for new user-defined types.

### DIFF
--- a/src/ekat/mpi/ekat_comm.cpp
+++ b/src/ekat/mpi/ekat_comm.cpp
@@ -58,5 +58,39 @@ void Comm::check_mpi_inited () const
   assert (flag!=0);
 }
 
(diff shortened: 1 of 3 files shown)


jeff-cohere (approved):
Nice use of template specialization.

PR #216, opened 2022-04-29
Add query method to factory, and added factory tests

bartgol:
## Motivation

In some cases, is best to query the factory for a certain project before attempting to create it. That way, we can decide what to do, or print a nice and meaningful error message.

Yes, one could achieve the same result using try/catch blocks, but it's just cleaner to probe the factory first.

## E3SM Stakeholder Feedback

Might help SCREAM

## Testing

Added tests for factory (which was completely untested so far).


jeff-cohere (approved):
I read thru the SCREAM Slack stuff. Sounds like a good addition.

PR #234, opened 2022-08-04
Fast-forwarded spdlog to v1.10.0 tag.

jeff-cohere:
This seems to eliminate some irritating C++ warnings and also includes bugfixes.

(We on EAGLES have been on a little campaign to eliminate the myriad C++/CUDA warnings we've encountered in order to minimize the superstition on our team.)


bartgol (approved):
Nice! I've been looking the other way so many times when building on weaver. Does this take away those "statement is unreachable" and "GCC unrecognized pragma" warnings?

jeff-cohere:
🙂 Hard to predict exactly what it'll do on any given machine. But it does fix at least two warnings that have been distracting me. Let's see what this does.

PR #235, opened 2022-08-05
Adds EKAT_ENABLE_FPE option for enabling floating point exceptions.

jeff-cohere:
This is an option we can use to decide whether we want EKAT to interact with
the floating point exception system at all. It's ON by default, to keep
behavior unchanged. Might come in handy on architectures where it's less
obvious how to handle floating point exception trapping (such as Mac M1).


bartgol (commented):
It seems like `EKAT_FPE` is doing the same thing. It seems pointless to use two macros for the same thing. I do like "${PKG}_ENABLE_XYZ" better though, so maybe we should simply replace EKAT_FPE with EKAT_ENABLE_FPE.

That said, I am wondering if EKAT should be in charge of defining what the FPEs are. Imho, we should continue to provide the functions to turn on/off FPEs, but each application should decide what their FPEs are.

bartgol:
If the goal is to not compile FPE-related stuff on certain architectures, we can put all of this stuff in a header-only file, which downstream apps can include or not, based on their needs.

jeff-cohere:
Yeah, I thought the same about `EKAT_FPE`, but the option description mentioned bit-for-bit stuff, so I didn't want to mess with it.

The reason I think EKAT has a stake in FPEs is that there is no real standard way to get them working on all architectures. It seems to me that if we want to be able to build EKAT irrespective of how weird the system is, either we need to expunge all FPE code from it or we need to have the flexibility to disable it if it prevents a successful configuration.

jeff-cohere:
Do you want me to reorganize things into a header-only FPE package? Should be fairly simple.

bartgol:
> Do you want me to reorganize things into a header-only FPE package? Should be fairly simple.

That would be nice, thanks. I would also remove the `get_default_fpes` function. We should let the apps decide what those are. But I now see that we do use that fcn in a few places in EKAT, so maybe we can leave it there.

I guess the only mod I would do is remove `EKAT_FPE`. It affects scream, but we can adjust scream at the next ekat update.

jeff-cohere:
So just remove `EKAT_FPE` for now?

bartgol:
Yes, please.

jeff-cohere:
Done!

src/ekat/ekat_assert.cpp
```diff
@@ -44,23 +47,33 @@ int get_default_fpes () {
 }
 
 void enable_fpes (const int mask) {
+#ifdef EKAT_ENABLE_FPE
```
bartgol, line 50: I'm wondering if we should throw an error if this fcn is called without `EKAT_ENABLE_FPE` being defined. The user might wrongfully assume its mask is doing something, while it's not.

Also, we should silence compiler warnings if EKAT_ENABLE_FPE is not defined, with the usual `(void) mask;`.

bartgol: An alternative to the error: should we make certain function declared only if `EKAT_ENABLED_FPE` is defined? The end result is the same, we just catch it at compile time.

jeff-cohere: Maybe we can use `spdlog` to issue a run-time warning if these functions are called without `EKAT_ENABLE_FPE`. What do you think?

Agree about silencing compiler warnings (speaking of `spdlog`!)

jeff-cohere: Leaving the functions out if `EKAT_ENABLE_FPE` is indeed a compile-time fix, but might cause CPP propagation up through caller code. If that's okay with you, I'll go ahead and move the guards outside those functions.

bartgol: It's true that apps might have code like
```
auto m = ekat::get_default_fpes();
ekat::enable_fpes(m);
```
and they might not want to wrap it in some `#ifdef`. So You're right, we should not conditionally compile these fcns, but we should instead just warn the user.

I would use printf, and not spdlog, since we would have to build a logger on the fly. It would have to be console only, and on all ranks. So it would basically be just like a printf.

jeff-cohere: Any opinion on `stdout` vs `stderr`? Probably splitting hairs here. :-)

bartgol: I always print to stdout, but I guess stderr is more appropriate?

bartgol (approved):

PR #238, opened 2022-08-10
Add a non-templated base class for ekat::Logger, exposing all functionalities

bartgol:
## Motivation

The reason for having a non-templated base class is to allow customers to store a pointer to the logger, keeping all the ekat Logger functionalities. Without a non-templated class, customers had to store a pointer to an spdlog::logger object, and could not access the ekat getters/setters without trying to cast to all possible derived types.

## Testing

No additional test needed. Functionalities in the `Logger` class are the same, they are just moved in a base class.


pbosler (approved):
Thanks @bartgol, this makes a lot of sense; I probably should have done it this way from the beginning...

PR #240, opened 2022-08-18
Make Units usabe inside constant expressions

bartgol:
## Motivation

Units were not constexpr constructible b/c of the runtime string stored. It occurred to me that we can still use `const char*`, by requiring that the unit name is passed in at construction time, which is definitely not a restriction. This would open the door to compile-time checks on units, by templating stuff on units.

## Testing

Turned some vars in the units testing to constexpr vars.

### DIFF
--- a/src/ekat/util/ekat_units.hpp
+++ b/src/ekat/util/ekat_units.hpp
@@ -5,6 +5,8 @@
 #include "ekat/util/ekat_scaling_factor.hpp"
 #include "ekat/util/ekat_string_utils.hpp"
 
+#include <array>
+
 namespace ekat
 {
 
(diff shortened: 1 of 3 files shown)


src/ekat/util/ekat_units.hpp
```diff
@@ -97,23 +110,23 @@ class Units {
...
+  friend constexpr Units operator/(const Units&,const ScalingFactor&);
+  friend constexpr Units operator/(const ScalingFactor&,const Units&);
+  friend constexpr Units pow(const Units&,const RationalConstant&);
+  friend constexpr Units sqrt(const Units&);
+  friend constexpr Units root(const Units&,const int);
 
   friend std::string to_string(const Units&);
 
   const ScalingFactor     m_scaling;
   const RationalConstant  m_units[7];
```
jeff-cohere, line 127: Seems like Weaver is having trouble copying this ^^^ member in one of your `constexpr` constructors. Maybe it needs to be copied within the constructor body?

bartgol:
Apparently GNU is capable of seeing

```
const Units a = Units(1);
```
and turn it into a constant expression behind the scenes (since rhs is a constant expressino), but nvcc is not doing it. To make `a` a constant expression one has to beg nvcc for it:
```
constexpr Units a = Units(1);
```
You live you learn.

jeff-cohere:
> You live you learn.

Truth!

jeff-cohere (approved):

PR #241, opened 2022-08-20
Add utility to join entries of iterable container into a string

bartgol:
## Motivation

I've wished several times that C++ had an equivalent to Python's `string.join(iterable)` method. This PR adds the closest thing we can get to that.

## Testing

Added a test, with three different types of containers.


jeff-cohere (approved):
Nice!

PR #243, opened 2022-08-24
Prevent SetWarningFlags from overwriting existing CFLAGS/FFLAGS/CXXFLAGS

### LINKED ISSUE #242: Append to rather than set CXXFLAGS, etc.
Here: https://github.com/E3SM-Project/EKAT/blob/34d2130791dbe116a43bfcf1afaa605b74c9f114/cmake/EkatSetCompilerFlags.cmake#L127-L133 CXXFLAGS is being set rather than string-appended to. This overwrites flags from E3SM's cime_config cmake files.

bartgol:
## Motivation

The macro `SetWarningFlags` used very common names for the f/c/cxx flags vars. Being a macro, it overwrote any homonymus var already in scope (if any).

## Related Issues
Fixes #242 

## E3SM Stakeholder Feedback

Needed in SCREAM.

## Testing

I verified in scream that doing `set(CXXFLAGS "hello world")` at the top of the main CMakeLists.txt, and printing its value at the bottom will print 
- `-Wall` in master
- `hello world` with this branch.

That is, upon returning from the macro, the original value of the var CXXFLAGS is recovered.

### DIFF
--- a/cmake/EkatSetCompilerFlags.cmake
+++ b/cmake/EkatSetCompilerFlags.cmake
@@ -119,7 +119,7 @@ endmacro ()
 #############################
 # Warnings
 #############################
-macro (SetWarningFlags)
+function (SetWarningFlags)
   # enable all warning but disable Intel vectorization remarks like
   #    "remark: simd loop has only one iteration"
   # since we would get hit with 1000's of those anytime we use Packs with packsize=1.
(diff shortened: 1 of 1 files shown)


ambrad:
Did you test this in scream?

bartgol:
> Did you test this in scream?

Yes.

I verified in scream that doing `set(CXXFLAGS "hello world")` at the top of the main CMakeLists.txt, and printing its value at the bottom will print

-`-Wall` in master
- `hello world` with this branch.

ambrad:
Ok. I meant running scream, not just configuring it, but it's fine with me if you want to wait for a PR into scream to do that testing.

ambrad (approved):

PR #244, opened 2022-08-24
Include ekat's configuration where needed

bartgol:
## Motivation

The default fpes were always 0, since `ekat_assert.cpp` was never (directly or indirectly) including `ekat_config.h`, which is where `EKAT_ENABLE_FPE` is defined.

## E3SM Stakeholder Feedback

Needed to correctly build fpe testing for scream.
## Testing

I verified SCREAM's fpe builds are picking up FPEs with this branch.


jeff-cohere (dismissed):

bartgol:
Ugh, our lin_interp test has fpe's, since it uses non trivial packs over a randomly generated number of levels, so there can be padding, which gets filled with nan's.

I will change our testing, to use `EKAT_ENABLE_FPE=OFF` in our current sp/dp builds, and add an extra build with fpe=on and packsize 1.

jeff-cohere (approved):
What a pain!

PR #246, opened 2022-08-26
Cmake and FPE fixes

bartgol:
## Motivation

There are two parts to this PR, which handle separate issues.
1. The concept of default FPEs need to be separated from that of default FPEs. The reason is that a library might not want FPEs, but being called from a host app, it might be interested in temporarily disabling FPEs. Therefore, FPE machinery needs to be enabled in EKAT, but default fpe mask must be 0. With this PR, one can set `EKAT_ENABLE_FPE=ON` and `EKAT_ENABLE_DEFAULT_FPE=OFF`. I would like the first one to be called `EKAT_ENABLE_FPE_SUPPORT`, bbut I'm worried it woulbe too much non-backward compatible.
2. The cmake utility `BuildEkat` was overly complicated, and hid a potential bug: the util allowed the user to provide a PREFIX, and the macro would set EKAT_FOO from ${PREFIX}_FOO (if set). This was appealing, but if EKAT later changes the var name from FOO to BAR, the host code would not see this.

## E3SM Stakeholder Feedback

This helps scream v1 to correctly run with packsize>1, avoiding FPE traps set by E3SM.


bartgol:
Putting on hold, since the issues I was having might be due to something else.

bartgol:
Closing, cause one can still enable ekat's FPE machinery setting EKAT_ENABLE_FPE=ON, but never relying on ekat's default fpe mask.

As for the BuildEkat routine, I might clean that at a later time.

PR #248, opened 2022-08-26
Refactor a bit how FPE are handled in ekat

bartgol:
## Motivation

Downstream app should be free to enable FPE support without having EKAT enabling some FPE by default. This PR decouples the support and defaults features, with the latter being removed from the main ekat library. In other words, the production library ekat does not offer any support for default FPEs, and the downstream app is responsible for determining what those should be. However we do still provide some default FPE support in the ekat_test_session library, which can be used for unit tests driven by ekat's catch2 framework main.

## E3SM Stakeholder Feedback

This PR allows SCREAM to build with FPE support but no FPE enabled, which is necessary for scream to run with packs (which can contain garbage if padded) in an E3SM build that enables FPEs. In such a build, scream needs to be able to disable FPEs upon entry, and re-instate the original FPE mask before returning.

## Testing

With this branch, SCREAM (with proper mods due to new configuration) correctly builds with/without fpe enabled. It also builds correctly within CIME, with E3SM enabling fpe traps, but scream not throwing (i.e., successfully disabling fpes upon entry, and disabling them before returning).


bartgol:
Upon conversation with @jeff-cohere , I reinstated the possibility of having a reasonable default FPE mast set by ekat in `ekat::initialize_ekat_session`. This default mask is not on by default, and the downstream app are encouraged to decide for themselves what FPEs to use. However, if they don't have the knowledge/will to decide, they can ask EKAT to enable a reasonable default mask.

jeff-cohere (dismissed):
Looks great.

jeff-cohere (approved):

PR #253, opened 2022-09-19
Removed pointless const, which prevented assignment

bartgol:
## Motivation

The units classes (`Units`, `ScalingFactor`, `RationalConstant`) do not offer a way to alter the class content anyways. And constexpr object will still be fine. Without const, we can now use these classes also at runtime.

## E3SM Stakeholder Feedback

Needed in scream, to allow passing units objects inside parameter lists.


jeff-cohere (approved):
It's definitely not difficult to have too much `const`. :-)

PR #256, opened 2022-10-12
Fix view_reduction to work with output variable coming from either private or shared memory

### LINKED ISSUE #254: Potential problem with ExeSpaceUtils view_reduction and parallel_reduce
**Describe the bug**
This was discovered when porting shoc_energy_integrals to small kernels. I was getting large differences in the outputs of the view_reductions when num_threads>1. I suspect the problem is in the handling of the garbage of the last pack because the problem went away when I used nlev % pack_size = 0.

**To Reproduce**
1. Switch shoc_energy_integrals to the implementation it had before the small kernel PR. The one that uses view_reductions.
2. Build SCREAM with `-DSCREAM_SMALL_KERNELS=On -DCMAKE_BUILD_TYPE=Debug`
3. run `OMP_NUM_THREADS=16 ./shoc_tests shoc_main_bfb`
4. This should fail due to being non_bfb with fortran. You can add print statements to confirm that the se_int, ke_int, wv_int, and wl_int values do not match fortran, which causes different results later in shoc for the output views.

**Expected behavior**
view_reduction should have produced bfb results with fortran.

bartgol:
## Motivation

In case `result` came from shared memory (e.g., an entry of a view), race conditions caused wrong results. This PR does the same thing that was done in `impl::parallel_reduce`, namely uses an automatic local var, then copies back in the output var.

## Related Issues
* Closes #254 
## E3SM Stakeholder Feedback

Might help scream, even though there's already a fix there.


tests/kokkos/kokkos_utils_tests.cpp
```diff
@@ -364,10 +364,10 @@ TEST_CASE("view_reduction", "[kokkos_utils]")
   test_view_reduction<Real,false,false,3,4> ();
 
   // Sum subset of entries, non-zero starting value
-  test_view_reduction<Real, true,true,16,3> (1.0/3.0,2,11);
-  test_view_reduction<Real,false,true,16,3> (1.0/3.0,2,11);
-  test_view_reduction<Real, true,false,16,3> (1.0/3.0,2,11);
-  test_view_reduction<Real,false,false,16,3> (1.0/3.0,2,11);
+  test_view_reduction<Real, true,true,18,4> (1.0/3.0,2,11);
```
bartgol, line 367: I have no idea how 3 was not generating a compiler error. We don't allow packs of non-power-of-2 size. I could have looked into why this does not throw an error, but I figured I fixed it and call it a day.

jgfouca:
@bartgol , note that I also needed to use local variables when using `ExeSpaceUtils::parallel_reduce`

bartgol:
> @bartgol , note that I also needed to use local variables when using `ExeSpaceUtils::parallel_reduce`

Ah, it might be that `impl::parallel_reduce` needs a barrier, to avoid the race condition Andrew pointed out in the shoc small kernel PR in scream:
```
auto local_var = resullt
// perform reduction into local_var
result = local_var // thread A might execute this line *before* thread B executes the first line
```

bartgol:
Shoot, the AT beat me. I'll do a super quick separate PR

ambrad:
I didn't get a chance to review before this was merged, but I agree with Jim that this PR is not complete. parallel_reduce needs to be reworked, too.

bartgol:
Yeah, I'm working on it.

jgfouca:
@bartgol , another thought, do you think it would be worth adding a test that would have exposed the prior deficiency?

bartgol:
> @bartgol , another thought, do you think it would be worth adding a test that would have exposed the prior deficiency?

That's something I was thinking about, yes. I will add a manifactured test, and verify it would have failed before.

ambrad:
@bartgol @jgfouca another thing I wanted to discuss was my opinion that there should be a version of view_reduction that does not take `result` as an input, thus negating this issue of automatic variable vs team-global. One could argue that the design flaw was permitting result to contain a valid value on input to be accumulated. That flaw is now leading to team_barrier calls that are strictly needed only when `result` is a team-global variable.

oksanaguba:
@bartgol  what is this PR for? Or, what i really want to know, will there be EKAT changes to scream master soon? Asking because i then would need to pull them into my branch, do not want to miss something.

bartgol:
This PR requires no change in scream. The follow up one will require a small change in `shoc_energy_integrals_impl.hpp`. I'll let you know when that happens.

oksanaguba:
Not sure I was clear -- please, let me know when you change head commit of EKAT submodule in scream master. thanks!

PR #258, opened 2022-10-17
Fix reduction utilities, part 2

bartgol:
## Motivation

We want to make sure that the bug we had related to reduction utilities never shows up again. Therefore, we force using a version of reduction utilities that returns the value, rather than update an input var. We still _do_ have one utility that takes the result as a ref, namely one overload of `impl::parallel_reduce`. We need that in order to perform Serialized reductions in `impl::view_reduction`. The user could technically still get the bug by using that overload improperly (e.g., passing a view entry as result). My hope is that, being the method in the `impl` namespace, and putting a big WARNING in the inline documentation, the risk of this happening is very minimal.

~~Bonus: fix the TPL include paths, so that they are treated as SYSTEM paths (if `EKAT_DISABLE_TPL_WARNINGS=ON`). This should fix  all the warnings from spdlog we get on GPU builds.~~ The commit(s) that did this got lost. I will issue a separate PR.

### DIFF
--- a/src/ekat/kokkos/ekat_kokkos_utils.hpp
+++ b/src/ekat/kokkos/ekat_kokkos_utils.hpp
@@ -60,33 +60,30 @@ ValueType parallel_reduce (const TeamMember& team,
                            const int& begin,
                            const int& end,
                            const Lambda& lambda,
-                           const ValueType& result)
+                           const ValueType& init)
 {
-  ValueType temp = result;
+  // Note the version of single(..) that takes the result as last arg is handy on Cuda,
+  // since it also implements a shuffle to ensure all threads in a warp have the same
+  // result as the one in vector lane 0.
+  ValueType result = init;
   if (Serialize) {
-    // We want to get C++ on GPU to match F90 on CPU. Thus, need to
-    // serialize parallel reductions.
-
-    // Note the version of single(..) that takes the result as last arg is handy on Cuda,
-    // since it also implements a shuffle to ensure all threads in a warp have the same
-    // result as the one in vector lane 0.
-    Kokkos::single(Kokkos::PerThread(team),[&] (ValueType& r) {
+    // Perform the reduction serially, with just one thread.
+    Kokkos::single(Kokkos::PerTeam(team),[&] (ValueType& r) {
       for (int k=begin; k<end; ++k) {
         lambda(k, r);
       }
-    },temp);
+    },result);
   } else {
-    // Kokkos::parallel_reduce inits var to 0, so we need to perform reduction on a temp,
-    // then add to result.
-    Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, begin, end), lambda, temp);
+    // Use the whole team of threads
+    Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, begin, end), lambda, result);
 
-    // Only one vector lane should update the result
-    Kokkos::single(Kokkos::PerThread(team),[&](ValueType& r){
-      r += result;
-    },temp);
+    // Kokkos::parallel_reduce inits result to 0, so we have to add init back again
+    // Use single, so that, on GPU, only one thread updates the result
+    Kokkos::single(Kokkos::PerTeam(team),[&](ValueType& r){
+      r += init;
+    },result);
   }
-
-  return temp;
+  return result;
 }
 
 template <bool Serialize, typename ValueType, typename TeamMember, typename Lambda>
@@ -153,7 +150,7 @@ auto view_reduction (const TeamMember& team,
   if (has_garbage_begin) {
     const auto temp_input = input(pack_loop_begin-1);
     const int first_indx = begin % N;
-    Kokkos::single(Kokkos::PerThread(team),[&] (ValueType& r) {
+    Kokkos::single(Kokkos::PerTeam(team),[&] (ValueType& r) {
       for (int j=first_indx; j<N; ++j) {
         r += temp_input[j];
       }
@@ -188,7 +185,7 @@ auto view_reduction (const TeamMember& team,
     // The following is basically a const var, but there are issues with
     // gnu and std=c++14. The macro ConstExceptGnu is defined in ekat_kokkos_types.hpp.
     ConstExceptGnu int last_indx = end % N;
-    Kokkos::single(Kokkos::PerThread(team),[&] (ValueType& r) {
+    Kokkos::single(Kokkos::PerTeam(team),[&] (ValueType& r) {
       for (int j=0; j<last_indx; ++j) {
         r += temp_input[j];
       }
--- a/src/ekat/kokkos/ekat_kokkos_utils.hpp
+++ b/src/ekat/kokkos/ekat_kokkos_utils.hpp
@@ -64,39 +64,39 @@ void parallel_reduce (const TeamMember& team,
                       const Lambda& lambda,
                       ValueType& result)
 {
+  // All threads init result.
+  // NOTE: we *need* an automatic temporary, since we do not know where 'result'
+  //       comes from. In particular, we don't know if result is already a
+  //       thread-local var (and automatic var itself) or shared across threads
+  //       (e.g., an entry from a view). In the former case, we would be fine
+  //       without the local var; in the latter, however, not using a local var
+  //       could lead to result being updated multiple times.
+  auto local_tmp = result;
+
   if (Serialize) {
     // We want to get C++ on GPU to match F90 on CPU. Thus, need to
     // serialize parallel reductions.
 
-    // All threads init result.
-    // NOTE: we *need* an automatic temporary, since we do not know
-    //       where 'result' comes from. If result itself is an automatic
-    //       variable, using it would be fine (only one vector lane would
-    //       actually have a nonzero value after the single). But if
-    //       result is taken from a view, then all vector lanes would
-    //       see the updated value before the vector_reduce call,
-    //       which will cause the final answer to be multiplied by the
-    //       size of the warp.
-    auto local_tmp = result;
+    // Ensure all threads are done reading from result, to avoid the possibility
+    // that some thread might update result (a few line below) before all other
+    // threads are done.
+    team.team_barrier();
 
     Kokkos::single(Kokkos::PerThread(team),[&] {
-        for (int k=begin; k<end; ++k) {
-          lambda(k, local_tmp);
-        }
-      });
+      for (int k=begin; k<end; ++k) {
+        lambda(k, local_tmp);
+      }
+    });
 
 #ifdef EKAT_ENABLE_GPU
     // Broadcast result to all threads by doing sum of one thread's
     // non-0 value and the rest of the 0s.
     team.vector_reduce(Kokkos::Sum<ValueType>(local_tmp));
 #endif
-
-   result = local_tmp;
   } else {
-    const ValueType initial(result);
-    Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, begin, end), lambda, result);
-    result += initial;
+    Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, begin, end), lambda, local_tmp);
   }
+  result = local_tmp;
 }
 
 /*
--- a/src/ekat/kokkos/ekat_kokkos_utils.hpp
+++ b/src/ekat/kokkos/ekat_kokkos_utils.hpp
@@ -56,12 +56,13 @@ namespace impl {
  */
 template <bool Serialize, typename ValueType, typename TeamMember, typename Lambda>
 static KOKKOS_INLINE_FUNCTION
-void parallel_reduce (const TeamMember& team,
-                      const int& begin,
-                      const int& end,
-                      const Lambda& lambda,
-                      ValueType& result)
+ValueType parallel_reduce (const TeamMember& team,
+                           const int& begin,
+                           const int& end,
+                           const Lambda& lambda,
+                           const ValueType& result)
 {
+  ValueType temp = result;
   if (Serialize) {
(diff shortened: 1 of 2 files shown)


jgfouca (dismissed):
Thx, Luca!

jgfouca (approved):

ambrad:
I removed AT:AUTOMERGE while I review this PR.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -53,135 +53,159 @@ namespace impl {
...
-                     ValueType& result)
+                     const InputProvider& input)
+ -> typename std::enable_if<ekat::impl::ResultTraits<InputProvider>::is_simd,
+                            typename ekat::impl::ResultTraits<InputProvider>::scalar_type>::type
 {
-  using PackType = typename std::remove_reference<decltype(input(0))>::type;
-  constexpr int vector_size = PackType::n;
+  using PackType = ekat::impl::ResultType<InputProvider>;
+  using ValueType = typename ekat::impl::ResultTraits<InputProvider>::scalar_type;
+  constexpr int N = sizeof(PackType) / sizeof(ValueType);
```
ambrad, line 132: Is it no longer possible to use PackType::n here?

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -53,135 +53,159 @@ namespace impl {
...
- * added to the value 'result'.
+ * Computes a reduction over the range [begin,end) of items provided by the InputProvider
+ * object, which must support calls to operator()(int).
+ * We provide two overloads, one for InputProvider returning a simd-type, and one for
+ * InputProvider returning a non-simd type. In both cases, the result of the reduction
+ * is a non-simd type, and the [begin,end) range refers to the 'scalar' range.
+ * NOTE: for simd-type usage, the simd-type must a) have a specialization of ekat::ScalarTraits
+ *       that sets is_simd=true, b) support operator[](int), and c) have a compile-time size
+ *       equal to the length of the simd array times the size of the individual element type.
+ *       Furtermore, a specialization of Kokkos::reduction_identity<T> must be available for
```
ambrad, line 106: "Furt(h)ermore"

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -53,135 +53,159 @@ namespace impl {
  * If Serialize=false, this function simply calls Kokkos::parallel_reduce().
  * If result contains a value, the output is result+reduction
  * For typical application, begin and end are pack indices.
- * NOTE: we do not provide an overload with Serialized defaulted, since this fcn is an impl
- *       detail, and, normally, should not be used by customer apps
+ * WARNING! This fcn should *not* be used outside this file. The implementation *relies*
```
ambrad, line 56: This is fine. Technically, we could return `result` to enforce this comment, and probably the compiler would use return value optimization to make that free, but we can't be sure. See below for further comments concerning the issue of whether `results`'s input value is meaningful.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -53,135 +53,159 @@ namespace impl {
...
-    // Broadcast result to all threads by doing sum of one thread's
-    // non-0 value and the rest of the 0s.
-    team.vector_reduce(Kokkos::Sum<ValueType>(local_tmp));
-#endif
-
-   result = local_tmp;
+    // Note the version of single(..) that takes the result as last arg is handy on Cuda,
+    // since it also implements a shuffle to ensure all threads in a warp have the same
+    // result as the one in vector lane 0.
+    Kokkos::single(Kokkos::PerThread(team),[&] (ValueType& result) {
```
ambrad, line 75: I'm confused about the intent here. If `result` enters this function non-0, then this `single` loop will accumulate into that non-0 value. But in the non-`Serialize` case, the `parallel_reduce` will 0-init `result`. So the two conditions branching on `Serialize` I think have two different behaviors.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -53,135 +53,159 @@ namespace impl {
...
 }
 
+template <bool Serialize, typename ValueType, typename TeamMember, typename Lambda>
+static KOKKOS_INLINE_FUNCTION
+ValueType parallel_reduce (const TeamMember& team,
+                           const int& begin,
+                           const int& end,
+                           const Lambda& lambda)
+{
+  ValueType result = Kokkos::reduction_identity<ValueType>::sum();
```
ambrad, line 92: `Kokkos::parallel_reduce` will 0-init `result`, so this line isn't needed. Or am I mistaken? This issue of 0-initing `result` is at the crux of my comments concerning `impl::parallel_reduce`.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -53,135 +53,159 @@ namespace impl {
...
-                                       [&](const int k, ValueType& local_sum) {
-          // Sum over pack entries and add to local_sum
-          ekat::reduce_sum<Serialize>(input(k),local_sum);
-      }, temp);
+      impl::parallel_reduce<Serialize,ValueType>(
+          team, pack_loop_begin, pack_loop_end,
+          [&](const int k, ValueType& local_sum) {
+            // Sum over pack entries and add to local_sum
+            ekat::reduce_sum<Serialize>(input(k),local_sum);
+      },result);
```
ambrad, line 166: Here I believe you're relying on `impl::parallel_reduce` to _not_ 0-init `result`. I think this might be working b/c of a sneaky alignment in assumptions: The `Serialize` code path _doesn't_ 0-init `result` and you rely on that here, while the `not Serialize` code path _does_ 0-init `result`, but the code below is OK with that.

src/ekat/kokkos/ekat_kokkos_utils.hpp
bartgol: Yes, it is, but I didn't want to unnecessarily rely on the packtype implementation details. This way, the user can potentially swap a different impl (so long as a specialization of `ScalarTraits` exists).

src/ekat/kokkos/ekat_kokkos_utils.hpp
bartgol: Ah, good point on ||reduce init-ing to 0. Second time I'm reminded about it today. Then I believe I have to use an auto var in the reduce fcn as well, or else I would lose the initial condition.

src/ekat/kokkos/ekat_kokkos_utils.hpp
bartgol: Yes, you're right.

src/ekat/kokkos/ekat_kokkos_utils.hpp
bartgol: Yeah, I was relying on that. I will mod `impl::parallel_reduce` so that the input result is not lost, regardless of Serialize

bartgol:
@ambrad I _think_ the last commit should fix the wrong impl for non-serialized case.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -53,135 +53,164 @@ namespace impl {
...
   } else {
-    const ValueType initial(result);
-    Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, begin, end), lambda, result);
-    result += initial;
+    // Kokkos::parallel_reduce inits var to 0, so we need to perform reduction on a temp,
+    // then add to result.
+    ValueType temp;
+    Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, begin, end), lambda, temp);
+
+    // Only one vector lane should update the result
```
ambrad, line 83: You should add back the warning at the top that `result` needs to be an automatic variable, since `PerThread` here means each thread in a team will accumulate `temp` into `result`.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -53,135 +53,164 @@ namespace impl {
...
 }
 
+template <bool Serialize, typename ValueType, typename TeamMember, typename Lambda>
+static KOKKOS_INLINE_FUNCTION
+ValueType parallel_reduce (const TeamMember& team,
+                           const int& begin,
+                           const int& end,
+                           const Lambda& lambda)
+{
+  ValueType result;
```
ambrad, line 97: Note that the way you've chosen to impl `impl::parallel_reduce` above (now consistent in both cases of `Serialize`) means you _do_ need to 0-init `result` here.

As an aside, I would have chosen to make `impl::parallel_reduce` return a value, thereby making this code as simple as possible.

src/ekat/kokkos/ekat_kokkos_utils.hpp
bartgol: I thought about using `PerTeam`. That would allow to use views rather than automatic vars, I think, right?

ambrad: I agree. The broadcast version of `single` with `PerTeam` will make it work for both.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -56,35 +56,37 @@ namespace impl {
...
+  ValueType temp = result;
   if (Serialize) {
     // We want to get C++ on GPU to match F90 on CPU. Thus, need to
     // serialize parallel reductions.
 
     // Note the version of single(..) that takes the result as last arg is handy on Cuda,
     // since it also implements a shuffle to ensure all threads in a warp have the same
     // result as the one in vector lane 0.
-    Kokkos::single(Kokkos::PerThread(team),[&] (ValueType& result) {
+    Kokkos::single(Kokkos::PerThread(team),[&] (ValueType& r) {
```
ambrad, line 73: If you change the `PerThread` below, you might as well do it here, too.

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -56,35 +56,37 @@ namespace impl {
...
-void parallel_reduce (const TeamMember& team,
-                      const int& begin,
-                      const int& end,
-                      const Lambda& lambda,
-                      ValueType& result)
+ValueType parallel_reduce (const TeamMember& team,
+                           const int& begin,
+                           const int& end,
+                           const Lambda& lambda,
+                           const ValueType& result)
```
ambrad, line 63: I meant that you should consider returning a value without also taking one in: in other words, make this a pure function so it's easy to reason about. Doing both is in some ways even worse that just leaving it as it was.

bartgol: Well, it must take the initial value, in order to do a serialized reduction. I guess I can change the name of the input to avoid confusion. But I do need the input, or else view_reduction would not be bfb in case begin!=0.

ambrad:
I'm approving based on not seeing anything that's incorrect in the new version. However, I want to document here an alternative impl that I believe would be simpler to reason about and faster, the latter by avoiding broadcasts. The idea is to leverage the fact that we really don't need to accumulate into anything other than an automatic scalar. This impl would not use `Kokkos::single`.

```c++
ResultType parallel_reduce (team, begin, end, lambda) {
  ResultType result;
  if (Serialize) {
    // Every physical thread (henceforth just 'thread') computes 'result'.
  } else {
    parallel_reduce(policy, lambda, result);
  }
  return result;
}

Scalar view_reduction (team, begin, end, input) {
  Scalar result;

  // Every thread accumulates first pack sum into 'result'.

  // Main part:
  auto pr_result = parallel_reduce(team, begin_pack, end_pack, lambda);
  // Every thread accumulates reduce_sum(pr_result) into 'result'.
  
  // Every thread accumulates last pack sum into 'result'.

  return result;
}
```

ambrad (dismissed):

bartgol:
@ambrad that's a much simpler implementation, and I like simpler. I'll push a change.

bartgol:
One solution could be to call `ekat::scalarize` in `view_reduction` if `Serialize=true`... Then, I believe your implementation would work.

ambrad:
> The issue is that the serial reduction, in that case, should not start from 0, but from whatever the sum of non-garbage entries in the first pack is. 

Good point. And we can't make both implementations sum the first and final packs after the reduction on the middle part because the goal is to match the F90 code's sum order. What do you think of simply scalarizing and summing straightforwardly in the `Serialize=true` case? It's not meant to be performant, anyway.

bartgol:
> > The issue is that the serial reduction, in that case, should not start from 0, but from whatever the sum of non-garbage entries in the first pack is.
> 
> Good point. And we can't make both implementations sum the first and final packs after the reduction on the middle part because the goal is to match the F90 code's sum order. What do you think of simply scalarizing and summing straightforwardly in the `Serialize=true` case? It's not meant to be performant, anyway, while the `Serialize=false` case could then be I think a little faster without the broadcasts.

Yeah, that could be a reasonable compromise. I'll give it a try.

bartgol:
@ambrad Ok, I simplified parallel_reduce, leaving only the version that returns (without any initial arg). For view reduction, if `Serialize=true`, I scalarize the lambda, and call the non-packed version (which simply calls parallel_reduce). SInce we only operate on automatic vars, the original bug should now be gone. Tests pass on CPU and GPU, but I'll let the AT do its thing.

ambrad (approved):
This looks great.

bartgol:
@jgfouca @tcclevenger unless I hear from you, I will merge this by the end of the day.

jgfouca (approved):
LGTM

tcclevenger (approved):
Same

PR #260, opened 2022-10-20
Allow EKAT to be built without MPI

jeff-cohere:
This PR makes it possible to use EKAT/Kokkos without MPI. I know many people don't care about this option and/or view it as counterproductive, but there are actually several interesting cases where (a) MPI is not needed at all (like column physics) and (b) it complicates development environments and workflows. One ignores these circumstances at the cost of a large productivity hit.

These changes illustrate the type of effort needed to enable a no-MPI build. I'm happy to maintain the "no-MPI-ness" in the near term, because the EAGLES aerosol work is column physics and does not need MPI in its standalone mode. If there's anything I can do to make it less distracting to support this option, let's talk!

### DIFF
--- a/src/ekat/mpi/ekat_comm.hpp
+++ b/src/ekat/mpi/ekat_comm.hpp
@@ -1,6 +1,10 @@
 #ifndef EKAT_COMM_HPP
 #define EKAT_COMM_HPP
 
+#include <ekat_config.h>
+
+#ifdef EKAT_ENABLE_MPI
+
 #include <type_traits>
 
 #include <mpi.h>
@@ -143,5 +147,7 @@ void Comm::all_gather (T* inout_vals, const int count) const
 }
 
 } // namespace ekat
+
+#endif // EKAT_ENABLE_MPI
 
 #endif // EKAT_COMM_HPP
--- a/src/ekat/mpi/ekat_comm.hpp
+++ b/src/ekat/mpi/ekat_comm.hpp
@@ -1,7 +1,7 @@
 #ifndef EKAT_COMM_HPP
 #define EKAT_COMM_HPP
 
-#include <ekat_config.h>
+#include <ekat/ekat_config.h>
 
 #ifdef EKAT_ENABLE_MPI
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,21 +21,29 @@ include(EkatBuildSpdlog)
 # EKAT has mostly C++, but some Fortran too
 project (EKAT C CXX Fortran)
 
-# EKAT requires MPI.
-find_package(MPI REQUIRED COMPONENTS C)
+# Here's an option for disabling MPI (because no, not everyone needs it
+# all the time).
+option (EKAT_ENABLE_MPI "Whether EKAT requires MPI." ON)
 
-# NOTE: may be an overkill, but depending on which FindMPI module is called,
-#       if _FIND_REQUIRED is not checked, we may not get a fatal error
-#       if the required components are not found. So check the _FOUND var.
-if (NOT MPI_C_FOUND)
-  message (FATAL_ERROR "EKAT *requires* the C component of MPI to be found")
+if (EKAT_ENABLE_MPI)
+  find_package(MPI REQUIRED COMPONENTS C)
+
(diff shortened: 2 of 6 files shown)


tcclevenger (commented):
Seems reasonable to me. I'm not aware of the dissenting opinion's finer points, so I'll let others chime in before approving.

src/ekat/mpi/ekat_comm.hpp
```diff
@@ -1,6 +1,10 @@
 #ifndef EKAT_COMM_HPP
 #define EKAT_COMM_HPP
 
+#include <ekat/ekat_config.h>
+
+#ifdef EKAT_ENABLE_MPI
```
bartgol, line 6: You are already compiling this  only conditionally, so what's the point of this? There are downstream files that include ekat_comm, so I think the cleanest solution is to provide a "serial" implementation of Comm, if MPI is not enabled. Either that, or we need to make the usage of Comm conditional to EKAT_ENABLE_MPI. See, for instance, `ekat_logger.hpp`.

bartgol: In Trilinos, they have a base class `Comm`, and concrete derived classes called `SerialComm` and `MpiComm`. Do you think we should do the same?

jeff-cohere: I do agree with you that a "serial" implementation of the basic elements of MPI is the tried-and-true way to go. I was just trying to see what I could get away with in the EAGLES environment, where we don't invest in MPI communicators (because we're only computing on a single node in standalone mode).

I think we could certainly do what Trilinos does. Another idea is to put all of Comm's behavior in a `.cpp` file and have the build system select the right one.

jeff-cohere:
Thanks for looking at this @bartgol . I'll put something together tomorrow for the "serial" Comm and we can take it from there.

src/ekat/mpi/ekat_comm.hpp
bartgol: Maybe. There are a few places with `MPI_XYZ` in the hpp though. I think all the implementation section at the bottom of the hpp can be moved in the cpp file (since we use ETI anyways). You will still need an ifdef in the hpp when the type `MPI_Comm` is mentioned, and also we would have to declare MPI_Op and MPI_Datatype enums in the serial case, though.

I'm not sure what the best path is. Perhaps you could change all fcns that use Comm in ekat, and roll something like
```
#ifdef EKAT_ENABLE_MPI
void foo (const Comm& comm);
#else
void foo ();
#endif
```
(or even completely not define `foo` in the else branch, depending what it is). But I would remove the ifdefs from this file, since they are inconsistent with the approach of not compiling `ekat_comm.cpp` if no MPI is enabled.

jeff-cohere: Yeah, I think you're right, actually. It seems to me that the best approach is for people who don't need MPI not to refer to anything that touches MPI. I've removed the guards from `ekat_comm.hpp` and added guards to files within EKAT that refer to it. I've verified that EKAT builds in a serial Haero configuration. So maybe this is all we need?

jeff-cohere:
@overfelt pointed out some issues with logging that I had overlooked. I'm going to think about how to address these issues with minimal disruption to the MPI builds. I'm off next week, so there might not be much activity till afterward. Thanks for looking at this, guys!

src/ekat/mpi/CMakeLists.txt
```diff
@@ -1,4 +0,0 @@
-# Add
```
bartgol, line 1: Interesting that we had a stranded cmake files for so long...

jeff-cohere: The possibilities are endless in that line!

jeff-cohere:
Okay, I've pushed a serial implementation of `ekat::Comm` that I believe should be correct, and removed some of the guards preventing it from being used. I've also reenabled the Comm tests in the serial configuration.

Let's take a look at this and see what we do and don't like about it.

overfelt (approved):

bartgol (commented):
Looks good!

bartgol (approved):

PR #264, opened 2022-11-01
Fix initialize_kokkos to follow Kokkos::Impl::parse_command_line_arguments.

ambrad:
Kokkos::Impl::parse_command_line_arguments in Kokkos version >= 3.7.00 expects
argv to have one more entry than narg, as documented in a comment: 'Note that
argv has (argc + 1) arguments, the last one always being nullptr.' This is safe
to insert for < 3.7.00, as well, so no KOKKOS_VERSION switch is needed.

[BFB]

### DIFF
--- a/src/ekat/ekat_session.cpp
+++ b/src/ekat/ekat_session.cpp
@@ -76,6 +76,11 @@ void initialize_kokkos () {
   args.push_back(const_cast<char*>(silence));
 
   int narg = args.size();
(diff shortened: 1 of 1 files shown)


bartgol (approved):
I wonder what their rationale is. It breaks a basic and widestpread assumption that argc is the length of argv. Granted, `argc<=length(argv)` is still safe, so no big deal. Still, it seems odd.

PR #265, opened 2022-11-03
Fix cmake utility EkatDisableAllWarning

bartgol:
## Motivation

The current impl was changing the order of the include directories, by grabbing all interface dirs, and re-adding them as include dirs with SYSTEM specifier. Instead, the correct thing to do is to set the [INTERFACE_SYSTEM_INCLUDE_DIRECTORIES](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_SYSTEM_INCLUDE_DIRECTORIES.html) to contain all the include directories. However, this still doesn't fix failures we were observing in SCREAM. There seem to be no way to get the feature to work, so I'm going to roll it back.

## E3SM Stakeholder Feedback

Current impl was causing nonsensical failures in SCREAM on blake.
## Testing

I verified that, with this mod, SCREAM tests pass on blake.

### DIFF
--- a/cmake/EkatUtils.cmake
+++ b/cmake/EkatUtils.cmake
@@ -110,12 +110,6 @@ macro (EkatDisableAllWarning targetName)
     target_compile_options (${targetName} PRIVATE
       $<$<COMPILE_LANGUAGE:Fortran>:$<$<Fortran_COMPILER_ID:GNU>:-w> $<$<Fortran_COMPILER_ID:Intel>: -w>>)
   endif()
-
-  get_target_property (tgt_include_dirs ${targetName} INTERFACE_INCLUDE_DIRECTORIES)
-  if (tgt_include_dirs)
-    target_include_directories(${targetName} SYSTEM BEFORE PUBLIC ${tgt_include_dirs})
-  endif()
-
 endmacro (EkatDisableAllWarning)
 
 function(separate_cut_arguments prefix options oneValueArgs multiValueArgs return_varname)


bartgol:
Note: I am a little bit afraid the "fix" is simply hiding some incorrect cmake logic somewhere in SCREAM. Alas, this is a conjecture, and I can find no evidence of that, so I think this fix is the best we can do for now. Unfortunately, this will bring back some compiler warnings, notably spdlog warnings on GPU builds.

jgfouca (approved):

bartgol:
Something funky is happening with the AT. The crash report shows it failed to merge the PR due to conflicts, but fetching from the repo shows the PR was merged!
```
$ git log -1 origin/master
commit 44a1c7f26285a342bfa6b0dd75ced01aa443a70f (HEAD -> bartgol/fix-disable-warnings-cmake-util, origin/master, origin/HEAD)
Merge: 6e4fc15 2eaa544
Author: Autotester for E3SM related projects <56648600+E3SM-Autotester@users.noreply.github.com>
Date:   Thu Nov 3 10:09:47 2022 -0600

    Merge Pull Request #265 from E3SM-Project/EKAT/bartgol/fix-disable-warnings-cmake-util
    
    Automatically Merged using E3SM Pull Request AutoTester
    PR Title: Fix cmake utility EkatDisableAllWarning
    PR Author: bartgol
```

bartgol:
SEMS folks are looking into this. I don't want to act on this PR (e.g., rebase/merge) until they figured it out, so they can see all details. Meanwhile, I think master is safe to use, since it has this PR merged in.

PR #266, opened 2022-11-09
Fix bug in TeamUtils get/release_workspace_idx for GPU.

ambrad:
We need a memory_fence so that read/writes to work slots are sequenced properly w.r.t. the lock on the workspace slots.

Add a test that fails in ~2 out of 5 trials with the buggy code. Passes 100 times in a row with the fix.

Use Kokkos::single's broadcast instead of a separate reduction.

### DIFF
--- a/src/ekat/kokkos/ekat_kokkos_utils.hpp
+++ b/src/ekat/kokkos/ekat_kokkos_utils.hpp
@@ -573,31 +573,23 @@ class TeamUtils<ValueType,EkatGpuSpace> : public TeamUtilsCommonBase<ValueType,E
   KOKKOS_INLINE_FUNCTION
   int get_workspace_idx(const MemberType& team_member) const
   {
-    EKAT_KERNEL_ASSERT_MSG (this->_num_teams>0, "Error! TeamUtils not yet inited.\n");
+    EKAT_KERNEL_ASSERT_MSG (this->_num_teams > 0, "Error! TeamUtils not yet inited.\n");
 
-    if (!_need_ws_sharing) {
+    if ( ! _need_ws_sharing) {
       return team_member.league_rank();
-    }
-    else {
-      int ws_idx = 0;
-      Kokkos::single(Kokkos::PerTeam(team_member), [&] () {
+    } else {
+      int ws_idx_broadcast;
+      Kokkos::single(Kokkos::PerTeam(team_member), [&] (int& ws_idx) {
         ws_idx = team_member.league_rank() % _num_ws_slots;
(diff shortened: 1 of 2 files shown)


bartgol (approved):

jgfouca (approved):
Awesome! This looks like a hard one to find!

src/ekat/kokkos/ekat_kokkos_utils.hpp
```diff
@@ -573,31 +573,23 @@ class TeamUtils<ValueType,EkatGpuSpace> : public TeamUtilsCommonBase<ValueType,E
...
         ws_idx = team_member.league_rank() % _num_ws_slots;
-        if (!Kokkos::atomic_compare_exchange_strong(&_open_ws_slots(ws_idx), (flag_type) 0, (flag_type)1)) {
+        if ( ! Kokkos::atomic_compare_exchange_strong(&_open_ws_slots(ws_idx), (flag_type) 0, (flag_type) 1)) {
           rnd_type rand_gen = _rand_pool.get_state(team_member.league_rank());
           ws_idx = Kokkos::rand<rnd_type, int>::draw(rand_gen) % _num_ws_slots;
-          while (!Kokkos::atomic_compare_exchange_strong(&_open_ws_slots(ws_idx), (flag_type) 0, (flag_type)1)) {
+          while ( ! Kokkos::atomic_compare_exchange_strong(&_open_ws_slots(ws_idx), (flag_type) 0, (flag_type) 1))
             ws_idx = Kokkos::rand<rnd_type, int>::draw(rand_gen) % _num_ws_slots;
-          }
+          Kokkos::memory_fence();
```
bartgol, line 589: On a second thought, this memory fence might not be necessary, now that you switched to the impl of `single` that also does the team reduction.

ambrad: It is. I can explain in person. Briefly, memory fences are for read/write ordering. `volatile` takes care of visiblity. Both are needed. The broadcast change was cosmetic; I just did it since it's more concise.

ambrad: Based on this description, I realize I should place the memory fence in the release routine _before_ the write, not after. I'll look into this tomorrow. I suspect something is happening that makes either placement work in practice and so in the unit test and app (i.e., the h/w has stronger memory guarantees than the memory model), but I'd like this code to respect the memory model exactly.

bartgol: Ah, my comment was based on a peek in the `team_broadcast` method for Host teams (which occurs inside the `single` overload), which has a memory_fence already built in. I now see the Cuda version has a `__syncthreads` rather than a `__threadfence` in it. From the documentation, the latter appears to be the one that guarantees memory syncronization.

PR #268, opened 2022-11-10
Refine use of memory_fence in get/release_workspace_idx for GPU.

ambrad:
After thinking about read/write sequencing some more, I decided one memory_fence is not needed and the other should be placed earlier. I also strengthened the test. It now fails nearly 100% of the time in the buggy case, and multiple 100s of runs in the fixed case pass.

Edit: In fact, calling both memory_fences is best. But this PR still fixes placement of one, strengthens the test, and adds a bunch of comments.


ambrad:
@bartgol sorry for the second PR. Our conversation in PR #266 made me realize the fix could be improved.

ambrad:
Interesting. Failure on weaver. I'll investigate.

bartgol:
> @bartgol sorry for the second PR. Our conversation in PR #266 made me realize the fix could be improved.

No problem. I'll wait till AT passes, then will review.

ambrad:
Thanks, Luca. I pushed a fix. Very interesting case. Added a comment to explain. Turns out this new workspace test is really effective.

bartgol (approved):

PR #273, opened 2023-02-10
Use TeamVectorRange when possible

tcclevenger:
Use `TeamVectorRange` when no further parallelism is needed.

## Motivation
Avoids meaningless computation when vector dim > 1. 

## Related Issues

Part of https://github.com/E3SM-Project/scream/issues/2070

### DIFF
--- a/scripts/jenkins/jenkins.sh
+++ b/scripts/jenkins/jenkins.sh
@@ -17,6 +17,7 @@ export SCREAM_SCRIPTS=${WORK_DIR}/scream/components/eamxx/scripts
 export JENKINS_SCRIPT_DIR=${SCREAM_SCRIPTS}/jenkins # some scream env setups depend on this
 source ${SCREAM_SCRIPTS}/jenkins/${NODE_NAME}_setup
 source ${SCREAM_SCRIPTS}/source_to_load_scream_env.sh
+export SCREAM_MACHINE
 
 BATCHP=$(${SCREAM_SCRIPTS}/query-scream $SCREAM_MACHINE batch)
 
--- a/scripts/jenkins/jenkins.sh
+++ b/scripts/jenkins/jenkins.sh
@@ -9,10 +9,10 @@ fi
 
 cd $WORKSPACE/${BUILD_ID}/
 
-WORK_DIR=$(pwd)
+export WORK_DIR=$(pwd)
 
 # setup env, use SCREAM env
-SCREAM_SCRIPTS=${WORK_DIR}/scream/components/eamxx/scripts
+export SCREAM_SCRIPTS=${WORK_DIR}/scream/components/eamxx/scripts
 source ${SCREAM_SCRIPTS}/jenkins/${NODE_NAME}_setup
 source ${SCREAM_SCRIPTS}/source_to_load_scream_env.sh
 
--- a/scripts/jenkins/jenkins.sh
+++ b/scripts/jenkins/jenkins.sh
@@ -3,5 +3,20 @@
 export JENKINS_SCRIPT_DIR=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
 DATE_STAMP=$(date "+%Y-%m-%d_%H%M%S")
 
+if [ -z "$WORKSPACE" ]; then
+    echo "Must run from Jenkins job"
+fi
+
+cd $WORKSPACE/${BUILD_ID}/
+
+WORK_DIR=$(pwd)
+
+# setup env, use SCREAM env
+SCREAM_SCRIPTS=${WORK_DIR}/scream/components/eamxx/scripts
+source ${SCREAM_SCRIPTS}/jenkins/${NODE_NAME}_setup
+source ${SCREAM_SCRIPTS}/source_to_load_scream_env.sh
+
+BATCHP=$(${SCREAM_SCRIPTS}/query-scream $SCREAM_MACHINE batch)
+
 set -o pipefail
-$JENKINS_SCRIPT_DIR/jenkins_impl.sh 2>&1 | tee JENKINS_$DATE_STAMP
+$BATCHP $JENKINS_SCRIPT_DIR/jenkins_impl.sh 2>&1 | tee JENKINS_$DATE_STAMP
--- a/scripts/jenkins/jenkins.sh
+++ b/scripts/jenkins/jenkins.sh
@@ -1,22 +1,24 @@
 #! /bin/bash -xe
 
-export JENKINS_SCRIPT_DIR=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
+EKAT_JENKINS_SCRIPT_DIR=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
 DATE_STAMP=$(date "+%Y-%m-%d_%H%M%S")
 
 if [ -z "$WORKSPACE" ]; then
     echo "Must run from Jenkins job"
+    exit 1
 fi
 
-cd $WORKSPACE/${BUILD_ID}/
+cd $WORKSPACE/${BUILD_ID}
 
 export WORK_DIR=$(pwd)
 
 # setup env, use SCREAM env
 export SCREAM_SCRIPTS=${WORK_DIR}/scream/components/eamxx/scripts
+export JENKINS_SCRIPT_DIR=${SCREAM_SCRIPTS}/jenkins # some scream env setups depend on this
 source ${SCREAM_SCRIPTS}/jenkins/${NODE_NAME}_setup
 source ${SCREAM_SCRIPTS}/source_to_load_scream_env.sh
 
 BATCHP=$(${SCREAM_SCRIPTS}/query-scream $SCREAM_MACHINE batch)
 
 set -o pipefail
-$BATCHP $JENKINS_SCRIPT_DIR/jenkins_impl.sh 2>&1 | tee JENKINS_$DATE_STAMP
+$BATCHP $EKAT_JENKINS_SCRIPT_DIR/jenkins_impl.sh 2>&1 | tee JENKINS_$DATE_STAMP
--- a/scripts/jenkins/jenkins_impl.sh
+++ b/scripts/jenkins/jenkins_impl.sh
@@ -38,7 +38,7 @@ if [[ "$ISCUDA" == "False" ]]; then
     EKAT_THREAD_SETTINGS="-DEKAT_TEST_THREAD_INC=2 -DEKAT_TEST_MAX_THREADS=14"
 fi
 
-# Build and test double precision
+# Build and test single precision
 mkdir -p ekat-build/ekat-sp && cd ekat-build/ekat-sp && rm -rf *
 
 cmake -C ${WORK_DIR}/ekat-src/cmake/machine-files/${NODE_NAME}.cmake \
@@ -80,7 +80,7 @@ else
 fi
 cd ${WORK_DIR}
 
-# Build and test single precision
+# Build and test double precision
 mkdir -p ekat-build/ekat-dp && cd ekat-build/ekat-dp && rm -rf *
 
 cmake -C ${WORK_DIR}/ekat-src/cmake/machine-files/${NODE_NAME}.cmake \
--- a/scripts/jenkins/jenkins_impl.sh
+++ b/scripts/jenkins/jenkins_impl.sh
@@ -55,7 +55,7 @@ fi
 # Build and test double precision
 mkdir -p ekat-build/ekat-sp && cd ekat-build/ekat-sp && rm -rf *
 
-cmake -C ${WORK_DIR}/ekat-src/cmake/machine-files/${NODE_NAME}.cmake \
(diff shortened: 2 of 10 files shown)


jgfouca (approved):

tcclevenger:
@jgfouca I'm guessing this weaver fail is module related?
```
CMake Error at /projects/ppc64le/spack-installs/cmake/3.23.1/gcc/7.2.0/base/knlpf2j/share/cmake-3.23/Modules/CMakeTestCCompiler.cmake:69 (message):
The C compiler

"/home/projects/ppc64le-pwr9-nvidia/spack-installs/openmpi/4.1.1/gcc/8.3.1/base/elujn3i/bin/mpicc"

is not able to compile a simple test program.

It fails with the following output:

Change Dir: /home/e3sm-jenkins/weaver/workspace/EKAT_PullRequest_Autotester_Weaver/481/ekat-build/ekat-sp/CMakeFiles/CMakeTmp
```
Is it an issue that we use gcc8.3.1, but the Error is in a file in spack-installs/cmake/3.23.1/gcc/7.2.0/.. ?

jgfouca:
@tcclevenger , weird, there's been no change to weaver configs. Are you able to duplicate this error on weaver by hand?

tcclevenger:
@jgfouca No, I copied the module load sequence and the cmake command. But I didn't get the error. Next week I can try to copy the whole case exactly.

jgfouca:
@tcclevenger , maybe try test-all-scream for a dbg run?

ambrad (approved):

bartgol (approved):

tcclevenger:
@jgfouca Running `test-all-scream -m weaver -t dbg` with scream master and this ekat branch checked out passes. 

I'm wondering if this is a bug that got through and exists on master. How do you run the ekat testsuite?

tcclevenger:
The error
```
CMake Error at /projects/ppc64le/spack-installs/cmake/3.23.1/gcc/7.2.0/base/knlpf2j/share/cmake-3.23/Modules/CMakeTestCCompiler.cmake:69 (message):

  The C compiler

    "/home/projects/ppc64le-pwr9-nvidia/spack-installs/openmpi/4.1.1/gcc/8.3.1/base/elujn3i/bin/mpicc"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /home/e3sm-jenkins/weaver/workspace/EKAT_PullRequest_Autotester_Weaver/483/ekat-build/ekat-sp/CMakeFiles/CMakeTmp
    Run Build Command(s):/usr/bin/gmake -f Makefile cmTC_24a1c/fast && /usr/bin/gmake  -f CMakeFiles/cmTC_24a1c.dir/build.make CMakeFiles/cmTC_24a1c.dir/build
    gmake[1]: Entering directory `/home/e3sm-jenkins/weaver/workspace/EKAT_PullRequest_Autotester_Weaver/483/ekat-build/ekat-sp/CMakeFiles/CMakeTmp'
    Building C object CMakeFiles/cmTC_24a1c.dir/testCCompiler.c.o
    /home/projects/ppc64le-pwr9-nvidia/spack-installs/openmpi/4.1.1/gcc/8.3.1/base/elujn3i/bin/mpicc    -o CMakeFiles/cmTC_24a1c.dir/testCCompiler.c.o -c /home/e3sm-jenkins/weaver/workspace/EKAT_PullRequest_Autotester_Weaver/483/ekat-build/ekat-sp/CMakeFiles/CMakeTmp/testCCompiler.c
    /home/projects/ppc64le-pwr9-nvidia/spack-installs/openmpi/4.1.1/gcc/8.3.1/base/elujn3i/bin/mpicc: /lib64/ld64.so.2: version `GLIBC_2.22' not found (required by /home/projects/ppc64le-pwr9-nvidia/spack-installs/ucx/1.12.1/gcc/8.3.1/base/z5nsbpl/lib/libucs.so.0)
    /home/projects/ppc64le-pwr9-nvidia/spack-installs/openmpi/4.1.1/gcc/8.3.1/base/elujn3i/bin/mpicc: /lib64/libc.so.6: version `GLIBC_2.25' not found (required by /home/projects/ppc64le-pwr9-nvidia/spack-installs/libevent/2.1.12/gcc/8.3.1/base/jk5oygh/lib/libevent_core-2.1.so.7)
    /home/projects/ppc64le-pwr9-nvidia/spack-installs/openmpi/4.1.1/gcc/8.3.1/base/elujn3i/bin/mpicc: /lib64/ld64.so.2: version `GLIBC_2.22' not found (required by /home/projects/ppc64le-pwr9-nvidia/spack-installs/numactl/2.0.14/gcc/8.3.1/base/f26egpf/lib/libnuma.so.1)
    /home/projects/ppc64le-pwr9-nvidia/spack-installs/openmpi/4.1.1/gcc/8.3.1/base/elujn3i/bin/mpicc: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by /home/projects/ppc64le-pwr9-nvidia/spack-installs/libxml2/2.9.12/gcc/8.3.1/base/fjluxu7/lib/libxml2.so.2)
    gmake[1]: *** [CMakeFiles/cmTC_24a1c.dir/testCCompiler.c.o] Error 1
    gmake[1]: Leaving directory `/home/e3sm-jenkins/weaver/workspace/EKAT_PullRequest_Autotester_Weaver/483/ekat-build/ekat-sp/CMakeFiles/CMakeTmp'
    gmake: *** [cmTC_24a1c/fast] Error 2
```

jgfouca:
@tcclevenger , I think something may be running on the login node that shouldn't be running there. Let me check

jgfouca:
@tcclevenger , yep, I was right. The cmake commands in the jenkins script were not going through the batch system.

@bartgol , please see my update to the jenkins infrastructure.

tcclevenger:
Thanks @jgfouca!

oksanaguba:
Conrad asked above where does scream set vector size to 1, i do not see the answer here (sorry if i missed it). it seems to me that there is no VECTOR_SIZE in scream, unlike homme. There is ekat::pack size, and loops are organized for league ind looping thru columns, threads ind looping thru packs, and packs in ekat are implemented via some simd (but they do not use kokkos TeamVectorRange). Is this correct?

So there is pack size, which is not necessarily 1. If there is a need to have a flat loop in a code with pack size > 1, then i would assume the implementation would be to have another data structure, with pack size=1, and using TeamThreadRange would be sufficient? That is, I am confused by Conrad's description `Avoids meaningless computation when vector dim > 1.` Maybe, because I do not know what operation would require a flat loop. Could anyone elaborate on this? thanks!

oksanaguba:
Partially answering my questions -- TeamVectorRange is used in WSM and in linear interpotation. So how did TeamThreadRange worked before with interpolation and packs>1 ? I see something in the code about repacking, if there was repacking to pack=1, then why change loops now?

tcclevenger:
@oksanaguba My understanding is that, with `TeamThreadRange` work is given to teams of threads, which themselves may have vector lanes. If there then is no interior loop to use these vector lanes then they will only be used to duplicate computation done on the thread. On the other hand, `TeamVectorRange` will split the work over threads and their vector lanes, so no duplicated computation.

I _think_ this is only an issue on GPU, and on CPU number of vector lanes is assumed to be 1? I could be wrong..

In SCREAM, on GPU, `packsize = max_threads = 1`. On CPU I think the default is `packsize = max_threads = 16`.

@ambrad The question has made me realize that I am not as confident as I thought I was about the reason for this change. If I am correct, why shouldn't there be an expected performance gain on GPU? Since vector lanes were duplicating computation that could have been split up.

ambrad:
First, read through this: https://github.com/E3SM-Project/scream/issues/2070#issuecomment-1424889660

Second, if there is a TTR with a nested ThVR, leave it as is. Certainly there are a few sprinkled throughout the code.

Third, if there is a TTR with _no_ nested ThVR, convert it to TeVR. In the TTR version, the Kokkos vector dimension is hard-set to 1. The TeVR will make clear that there is no nesting.

Fourth, packs and pack size have nothing to do with any of this. From Kokkos's perspective, they are just another scalar type. Note that I like to use the word "pack" when talking about our pack type, not "vector", precisely to disambiguate it from Kokkos's vector dimension.

scripts/jenkins/jenkins.sh
```diff
@@ -3,5 +3,20 @@
 export JENKINS_SCRIPT_DIR=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
 DATE_STAMP=$(date "+%Y-%m-%d_%H%M%S")
 
+if [ -z "$WORKSPACE" ]; then
+    echo "Must run from Jenkins job"
```
bartgol, line 7: We might as well error out here, no?

scripts/jenkins/jenkins.sh
```diff
@@ -3,5 +3,20 @@
...
+# setup env, use SCREAM env
+SCREAM_SCRIPTS=${WORK_DIR}/scream/components/eamxx/scripts
+source ${SCREAM_SCRIPTS}/jenkins/${NODE_NAME}_setup
+source ${SCREAM_SCRIPTS}/source_to_load_scream_env.sh
+
+BATCHP=$(${SCREAM_SCRIPTS}/query-scream $SCREAM_MACHINE batch)
+
 set -o pipefail
-$JENKINS_SCRIPT_DIR/jenkins_impl.sh 2>&1 | tee JENKINS_$DATE_STAMP
+$BATCHP $JENKINS_SCRIPT_DIR/jenkins_impl.sh 2>&1 | tee JENKINS_$DATE_STAMP
```
bartgol, line 22: I don't understand. Why do we need `jenkins_impl.sh` to be executed on compute nodes? Besides, that script will itself invoke a batch job (via test-all-scream).

jgfouca: @bartgol , this is for ekat, not scream.

bartgol:
> @tcclevenger , yep, I was right. The cmake commands in the jenkins script were not going through the batch system.
> 
> @bartgol , please see my update to the jenkins infrastructure.

I don't see why we need to run CMake through the batch system. It is pointless. It should be perfectly fine to configure/build on login node, and get on compute node just for the run phase. And when the inner gather-all-data dispatches another batch call, we'll be blocking 2 nodes instead of 1. When machines are swamped, this might cause timing out.

bartgol:
Besides, I think the current master impl already runs cmake on the compute node, since it's inside a gather-all-data call, no?

jgfouca:
@bartgol , i'm not sure either, but we saw the same thing on the scream side on weaver which led to this PR: https://github.com/E3SM-Project/scream/pull/2137

jgfouca:
@bartgol , to reiterate, this is for EKAT, not SCREAM, so there is no gather-all/test-all-scream going on here.

scripts/jenkins/jenkins.sh
bartgol: Ah, sorry.

bartgol:
I digged a bit, and the issue with GLIBC is that on compute nodes, there's a libc available that is not available on login nodes:
```
lbertag@weaver5:[ciccio]$ ldd /home/projects/ppc64le-pwr9-nvidia/spack-installs/libevent/2.1.12/gcc/8.3.1/base/jk5oygh/lib/libevent_core-2.1.so.7
	linux-vdso64.so.1 (0x0000200000050000)
	libpthread.so.0 => /lib64/power9/libpthread.so.0 (0x00002000000f0000)
	libc.so.6 => /lib64/power9/libc.so.6 (0x0000200000140000)
	/lib64/ld64.so.2 (0x0000200000000000)
```
vs
```
$ ldd /home/projects/ppc64le-pwr9-nvidia/spack-installs/libevent/2.1.12/gcc/8.3.1/base/jk5oygh/lib/libevent_core-2.1.so.7
/home/projects/ppc64le-pwr9-nvidia/spack-installs/libevent/2.1.12/gcc/8.3.1/base/jk5oygh/lib/libevent_core-2.1.so.7: /lib64/libc.so.6: version `GLIBC_2.25' not found (required by /home/projects/ppc64le-pwr9-nvidia/spack-installs/libevent/2.1.12/gcc/8.3.1/base/jk5oygh/lib/libevent_core-2.1.so.7)
	linux-vdso64.so.1 =>  (0x00007fff80f90000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fff80ed0000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fff80ce0000)
	/lib64/ld64.so.2 (0x00007fff80fb0000)
```

I am going to write to weaver help, to see if this is indeed an issue. It makes no sense that we're not allowed to cross compile for a compute node from login node.

scripts/jenkins/jenkins.sh
jgfouca: Yes, fixed.

tcclevenger:
Thanks again @jgfouca. Looks like tests are passing.

jgfouca:
@tcclevenger , thanks goodness. It's a bit tricky to test jenkins stuff without just letting jenkins do its thing, so apologies for this taking a while.

ambrad (approved):

PR #274, opened 2023-02-16
Support subiew_1 for rank2 views, returning a view with LayoutStride

bartgol:
## Motivation

We implemented `subview_1` only for views with rank>2, pushed by the constraint that the return type should still have LayoutRight. However, this prevents us from doing something like
```
auto sv = ekat::subview_1(input_view);
```
in a generic function. While it's true that LayoutStride comes with a performance price, it is perfectly ok to use it in areas where the performance hit is negligible and the subview is treated in a way that doesn't care about layout type. For instance, we cannot use this if we need to store the subview in the class, since the class member type needs to be hardcoded; however, it's doable in small snippets of code.

## E3SM Stakeholder Feedback

Needed in EAMxx.
## Testing

I added a test that ensures output layout and values are the expected ones.


bartgol:
> Looks good. Just curious, where do you want to use this in EAMxx?

I need to deal with the U/V components of surf_mom_flux, which has layout (ncols,2).

PR #275, opened 2023-02-18
Add possibility to pass bool flags to catch main

bartgol:
## Motivation

The only way we had to pass arbitrary parameters to a test was via `--ekat-test-params key=val`. For simple bool flags, this syntax is heavy. So I added a shorthand version: `--flags a,b,c` will populate the `flags` member of `TestSession` (an std::map) with bool entries with those names, setting their value to true. This should allow making development easier. E.g., one could have very little output to its test, but if `--flags verbose` is used, print more output:
```
auto verb = ekat::TestSession::get().flags["verbose"];
if (verb) {
  // print extra stuff
}
```
Before, one would have had to pass `--ekat-test-params verbose=true`, and in the test, do
```
auto verb = ekat::TestSession::get().params["verbose"];
if (verb=="true" || verb=="yes") {
  // print extra stuff
}
```

## Testing

Added testing for params and flags.

### DIFF
--- a/src/ekat/util/ekat_catch_main.cpp
+++ b/src/ekat/util/ekat_catch_main.cpp
@@ -31,7 +31,7 @@ int main (int argc, char **argv) {
   };
 
   // Read possible ekat-specific arguments
-  auto const readCommaSeparaterParams = [] (const std::string& cmd_line_arg) {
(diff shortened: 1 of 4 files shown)


jeff-cohere (approved):
Looks very helpful!

PR #278, opened 2023-02-24
Move MPI settings into EKAT cache

jgfouca:
Refactor cmake mach files a bit, add MPI cache files.

Fix valgrind suppression stuff to handle MPI, or lack of, correctly.

MPI global settings no longer passed to CUT as arguments.

Blake needs to rely on test-launcher to manage resources since it can't use srun.

## Motivation

EKAT needs to know about MPI options in order to correctly support valgrind suppression files. Since it needs to know, it also seems easier to use mpi cache settings in CUT instead of passing them in every time as arguments.

## Testing

Extensive testing on mappy and Blake.


bartgol (dismissed):
Nice cleanup.

bartgol (approved):

PR #280, opened 2023-02-28
EkatCreateUnitTest: Store shell command as a custom property of the test

jgfouca:
Will allow users to easily create shell commands that follow the same convention as a unit test.

The usage on the scream side looks like this:
```
  CreateUnitTestFromExec(p3_baseline_f90_fake p3_run_and_cmp
    THREADS ${SCREAM_TEST_MAX_THREADS}
    EXE_ARGS "-f -g ${BASELINE_FILE_ARG}"
    PROPERTIES DISABLED True)

  get_test_property(p3_baseline_f90_fake_omp${SCREAM_TEST_MAX_THREADS} FULL_TEST_COMMAND P3_F90_GEN)
```

## Testing

Used this approach to change how baselines are generated for p3|shoc_run_and_cmp.

### DIFF
--- a/cmake/EkatCreateUnitTest.cmake
+++ b/cmake/EkatCreateUnitTest.cmake
@@ -314,7 +314,7 @@ function(EkatCreateUnitTestFromExec test_name test_exec)
 
       # Set test properties
       math(EXPR CURR_CORES "${NRANKS}*${NTHREADS}")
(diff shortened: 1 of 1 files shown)


bartgol (approved):
Interesting feature. Looks like you first need to define the new property, via [define_property](https://cmake.org/cmake/help/latest/command/define_property.html).

jgfouca:
@bartgol , it worked for me without define_property.

bartgol:
Interesting. I thought you were not allowed to add a test property without first declaring it. I assumed that's why CMake was complaining...

jgfouca:
@bartgol , no, I found the source of the error. Some of the EKAT tests have complex shell commands with quotes, redirections, etc. This was confusing the Cmake argument parsing in set_tests_properties. I just need to quote the FULL_TEST_CMD dereference.

bartgol (approved):

PR #281, opened 2023-02-28
Replacing --kokkos-num-devices with --kokkos-map-device-id-by=mpi_rank.

jeff-cohere:
This PR replaces a deprecated compiler flag as described in [SCREAM issue 2026](https://github.com/E3SM-Project/scream/issues/2026).

### DIFF
--- a/src/ekat/ekat_session.cpp
+++ b/src/ekat/ekat_session.cpp
@@ -17,70 +17,13 @@ static int get_default_fpes () {
 }
 #endif
 
-// If we initialize from inside a Fortran code, we don't have access to
-// char** args to pass to Kokkos::initialize. In this case we need to do
-// some work on our own. As a side benefit, we'll end up running on GPU
-// platforms optimally without having to specify
-// --kokkos-map-device-id-by=mpi_rank on the command line.
-// We also will use this function to generate kokkos args if the user
-// specified none.
+// This function provides a simple and correct way to initialize Kokkos from
+// any context with the correct settings.
 void initialize_kokkos () {
-  // This is in fact const char*, but Kokkos::initialize requires char*.
-  std::vector<char*> args;
-
-  //   This is the only way to get the round-robin rank assignment Kokkos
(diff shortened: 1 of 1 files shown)


bartgol:
I wonder if we should change this impl to use `InitializationSettings`, rather than mimicking cmd line argc/argv. Seems more readable. What do you think?

jeff-cohere:
Sure, that makes sense. Let me put that together.

jeff-cohere:
If we don't need to count available devices and we don't need to specify the number of threads, it looks like the new implementation of this function would be

```
// This function provides a simple and correct way to initialize Kokkos from
// any context with the correct settings.
void initialize_kokkos () {
  auto const settings = Kokkos::InitializationSettings()
    .set_num_threads(255)
    .map_device_id_by("mpi_rank")
    .set_disable_warnings(true);
  Kokkos::initialize(settings);
}
```

Does that look right to you, @bartgol ?

bartgol:
> If we don't need to count available devices and we don't need to specify the number of threads, it looks like the new implementation of this function would be
> 
> ```
> // This function provides a simple and correct way to initialize Kokkos from
> // any context with the correct settings.
> void initialize_kokkos () {
>   auto const settings = Kokkos::InitializationSettings()
>     .map_device_id_by("mpi_rank")
>     .set_disable_warnings(true);
>   Kokkos::initialize(settings);
> }
> ```
> 
> Does that look right to you, @bartgol ?

Why not doing the device count? The code should not change from what's already in master, we simply need to add `.set_num_devices(nd)`, no?

jeff-cohere:
True enough!

bartgol (dismissed):

ndkeen:
Fixes #2026

ambrad:
We definitely want the device count. The comment was referring to the fact that it's not fatal not to get it. But it can decrease performance not to round-robin with the full number of devices.

bartgol (approved):

PR #283, opened 2023-03-03
Fix debug_tools tests in release mode

jgfouca:
Apparently the compiler is smart enough to know `(void) var` is not really using `var`, so it was compiling-out the tests.

## Motivation

This test was failing for Release builds. This PR fixes this.

## Testing

By hand, Release and Debug.


jgfouca:
@ambrad , added you as reviewer since Luca is out.

ambrad (approved):

PR #284, opened 2023-03-08
Allow to pass a lambda to ekat::join

bartgol:
## Motivation

The `join` util is mimicking the one in python. In python, you can use list comprehension to make things even niftier, but C++ can't do that. The closest we can do is pass a lambda to `join`, which will perform additional ops on its elements, and/or a selector, i.e., a lambda that returns true if the element is to be included in the output.

E.g., given a list of integers (be that a vector, list, or any iterable), the string of the squares of all even integers separated by a comma can be built as
```
auto even = [](const int i) { return i%2==0; };
auto pow2 = [] (const int) { return i*i; };
std::list<int> ints = {....};
auto squares = ekat::join_if (ints, even, pow2, ",")
``` 

This is not much less code than a manual for, but a) the join function hides all the logic regarding not adding separator at the end, and b) it might make the code cleaner (lambdas can have a meaningful name).

## Testing

Added tests for selector version (`join_if`) as well as the version taking an "apply" function.

### DIFF
--- a/src/ekat/util/ekat_string_utils.hpp
+++ b/src/ekat/util/ekat_string_utils.hpp
@@ -80,6 +80,58 @@ std::string join (const Iterable& v, const Lambda& f, const std::string& sep) {
   return ss.str();
 }
 
+template<typename Iterable, typename Selector>
+std::string join_if (const Iterable& v, const Selector& check, const std::string& sep) {
+  auto it = cbegin(v);
+  auto end = cend(v);
+  if (it==end) {
+    return "";
+  }
+
+  // It is complicated to figure out if/when to add sep, since we don't
+  // know a priori if the following items will satisfy the check.
+  // Hence, build short list of ptrs beforehand
+  using elem_t = typename std::remove_reference<decltype(*it)>::type;
+  std::list<elem_t*> ptrs;
+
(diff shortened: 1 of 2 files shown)


src/ekat/util/ekat_string_utils.hpp
```diff
@@ -61,6 +61,77 @@ std::string join (const Iterable& v, const std::string& sep) {
...
+      ptrs.push_back(&elem);
+    }
+  }
+  auto deref = [](const auto& p) {
+    return *p;
+  };
+  return ekat::join(ptrs,deref,sep);
+}
+
+// Combo of prebious two
```
jeff-cohere, line 108: `prebious` -> `previous`

jeff-cohere (approved):
Found a very minor typo, but otherwise looks great!

src/ekat/util/ekat_string_utils.hpp
bartgol: whoops, i missed this. hopefully i'll remember to take care of it next time I do a PR.

jeff-cohere: Just make sure you refer to this PR as "a prebious PR" when you do it.

PR #286, opened 2023-03-10
Improve valgrind suppression file generation

jgfouca:
Include an executable that is linked the same way most catch exes are linked.

## Motivation
This will add some suppressions that will help to keep innocuous valgrind errors down.

## Testing
Scream ctests with valgrind enabled.

### DIFF
--- a/valgrind_support/CMakeLists.txt
+++ b/valgrind_support/CMakeLists.txt
@@ -1,6 +1,9 @@
+include(EkatCreateUnitTest)
 
 add_executable(serial_hw serial_hw.cxx)
 add_executable(mpi_hw mpi_hw.cxx)
+
+EkatCreateUnitTest(mpi_valg_catch_test mpi_catch_test.cpp)
 
 add_custom_target(serial_sup ALL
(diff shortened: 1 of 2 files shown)


valgrind_support/CMakeLists.txt
```diff
@@ -1,7 +1,10 @@
+include(EkatCreateUnitTest)
 
 add_executable(serial_hw serial_hw.cxx)
 add_executable(mpi_hw mpi_hw.cxx)
 
+EkatCreateUnitTest(mpi_valg_catch_test mpi_catch_test.cpp)
```
bartgol, line 6: I think you might have forgotten to include this file in the commit maybe?

jgfouca: Good catch! Fixed.

bartgol (approved):
Looks good. Do you think it makes sense to eliminate `mpi_hw` exec at this point? The new one seems to create a supp file that is a superset of the old one, so maybe the old one is pointless?

jgfouca:
@bartgol , interestingly, for the version of openmpi that I am going to set mappy to for scream, both are needed.

PR #287, opened 2023-03-11
Update spdlog submodule to v1.11.0

bartgol:
## Motivation

The current version has some issues with C++17 on blake. Since the newer version seems to work seamlessly, we might as well just upgrade...

## E3SM Stakeholder Feedback

Needed in order to upgrade SCREAM to C++17.


jeff-cohere (approved):
🚗 💨

PR #289, opened 2023-03-16
[SYCL] Fix failed tests for PVC

abagusetty:
- Address SYCL failing tests on PVC
- Add Sunspot (ANL) machine config for Kokkos/PVC
- Removed XEHP (ATS) redundant hardware config

### DIFF
--- a/cmake/machine-files/jlse-xehp.cmake
+++ b/cmake/machine-files/jlse-xehp.cmake
@@ -1,2 +1,0 @@
-# Load XEHP arch with SYCL backend for kokkos
-include (${CMAKE_CURRENT_LIST_DIR}/kokkos/intel-xehp.cmake)
--- a/cmake/machine-files/kokkos/intel-pvc.cmake
+++ b/cmake/machine-files/kokkos/intel-pvc.cmake
@@ -1,0 +1,5 @@
+include (${CMAKE_CURRENT_LIST_DIR}/generic.cmake)
+
+option(Kokkos_ARCH_INTEL_PVC "" ON)
+set(Kokkos_ENABLE_SYCL TRUE CACHE BOOL "")
+set(Kokkos_ENABLE_DEPRECATED_CODE_3 FALSE CACHE BOOL "")
--- a/cmake/machine-files/kokkos/intel-xehp.cmake
+++ b/cmake/machine-files/kokkos/intel-xehp.cmake
@@ -1,5 +1,0 @@
-include (${CMAKE_CURRENT_LIST_DIR}/generic.cmake)
-
-option(Kokkos_ARCH_INTEL_XEHP "" ON)
-set(Kokkos_ENABLE_SYCL TRUE CACHE BOOL "")
-set(Kokkos_ENABLE_DEPRECATED_CODE_3 FALSE CACHE BOOL "")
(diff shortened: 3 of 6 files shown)


abagusetty:
Hi @bartgol Just verifying SYCL backend on the Aurora TDS system and setting up a CI on our end. In the process, several tests that used to fail for SYCL backend in the past were address except 3 tests. So trying to looking into these 3 via this PR.

Question: This test seems to fail with a precision issue on the PVC hardware. It might just be lucky if the significant digits match for greater than 12 digits (double) or 8 decimals (single-prec) ? I was hoping if you have any insights into this float-pt comparisons.

```bash
, values: 53.775955383012388, 53.775955383012395

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
lin_interp is a Catch v2.13.8 host application.
Run with -? for options

-------------------------------------------------------------------------------
lin_interp_soak
-------------------------------------------------------------------------------
/lus/gila/projects/CSC249ADSE15_CNDA/abagusetty/EKAT/tests/algorithm/lin_interp_test.cpp:67
...............................................................................

/lus/gila/projects/CSC249ADSE15_CNDA/abagusetty/EKAT/src/ekat/util/ekat_test_utils.hpp:55: FAILED:
  REQUIRE( lhs == rhs )
with expansion:
  53.775955383 == 53.775955383
```

bartgol (approved):
I don't have any experience with Intel GPUs, but the PR seems fine from what I can tell.

abagusetty:
> 

Thanks! I was inclined to get some precision check in place there but will see a more robust solution.

bartgol:
> Question: This test seems to fail with a precision issue on the PVC hardware. It might just be lucky if the significant digits match for greater than 12 digits (double) or 8 decimals (single-prec) ? I was hoping if you have any insights into this float-pt comparisons.
> 
> ```shell
> , values: 53.775955383012388, 53.775955383012395
> 
> ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> lin_interp is a Catch v2.13.8 host application.
> Run with -? for options
> 
> -------------------------------------------------------------------------------
> lin_interp_soak
> -------------------------------------------------------------------------------
> /lus/gila/projects/CSC249ADSE15_CNDA/abagusetty/EKAT/tests/algorithm/lin_interp_test.cpp:67
> ...............................................................................
> 
> /lus/gila/projects/CSC249ADSE15_CNDA/abagusetty/EKAT/src/ekat/util/ekat_test_utils.hpp:55: FAILED:
>   REQUIRE( lhs == rhs )
> with expansion:
>   53.775955383 == 53.775955383
> ```

That's hard to tell. You should prob run inside a debugger, and get a precise backtrace of this. I think we do expect ==, there is no trascendental fcn involved, and no reduction, so device and cpu _should_ give the same number, with full precision.

@jgfouca btw, it's not what happens here, but do yo know why we do an approx comparison if PackSize>1? I think lin interp should give the same answer, regardless of packing, no?

PR #291, opened 2023-04-27
Allow for persistent valgrind suppression files

jgfouca:
Add new cache entry `EKAT_VALGRIND_SUPPRESSION_FILE`. If unset, we will revert to the previous behavior of generating a suppression file on the fly.

## Motivation

The generated ones are OK but there's an element of nondeterminism
in some of the valgrind interactions with some of our TPLs (like MPI), so
a persistent, maintained suppression file is best for consistency.

## Testing

By hand and scripts-tests.


cmake/EkatCreateUnitTest.cmake
```diff
@@ -289,12 +289,10 @@ function(EkatCreateUnitTestFromExec test_name test_exec)
 
       # Setup valgrind/memcheck commmand modifications
       if (EKAT_ENABLE_VALGRIND)
-        if (EKAT_ENABLE_MPI)
-          set(VALGRIND_SUP_FILE "${CMAKE_BINARY_DIR}/mpi.supp")
-        else()
-          set(VALGRIND_SUP_FILE "${CMAKE_BINARY_DIR}/serial.supp")
+        if (NOT EKAT_VALGRIND_SUPPRESSION_FILE)
```
bartgol, line 292: Why not resorting to the ones we generate (mpi.supp or serial.supp) if this is not present? It

jgfouca: @bartgol , that should be how it works. EKAT_VALGRIND_SUPPRESSION_FILE is FORCE-set by the valgrind_support CMake file if it's not already set.

bartgol: Ah, I missed the FORCE.

bartgol (approved):

PR #292, opened 2023-04-28
Allow to add FIXTURE_SETUP/COMMON on a per-test basis in EkatCreateUnitTest

bartgol:
## Motivation

Consider the case where you have a test A, run for several rank/thread combinations. For each of these tests, some artifacts are generated, which we later want to check in a follow-up test. For instance, each test creates two output files, which need to be compared using a tool. Right now, we can do this with
```
EkatCreateUnitTest (blah blah.cpp
   MPI_RANKS 1 10
   PROPERTIES FIXTURE_SETUP the_files)
 
foreach (RANK IN RANGE 1 10
  add_test (NAME check_${RANK} COMMAND my_check_script --suffix ${RANK})
  set_property (TEST check_${RANK} PROPERTY FIXTURE_REQUIRED the_files)
endforeach()
```
However, each of the `check_N` tests must wait for *all* of the `blah_N` tests to complete (*and complete successfully*!), since all the `blah_N` tests contribute to the setup of the fixture `the_files`. It would be nice to have separate fixtures for each MPI test. This PR addresses that. It adds 4 optional arguments to `EkatCreateUnitTestFromExec`:
- `FIXTURES_SETUP_COMMON`: sets a `FIXTURES_SETUP` on all `blah_N` tests, with the same name
- `FIXTURES_REQUIRED_COMMON`: sets a `FIXTURES_REQUIRED` on all `blah_N` tests, with the same name
- `FIXTURES_SETUP_INDIVIDUAL`: sets a `FIXTURES_SETUP` on all `blah_N` tests, with a test-specifi name
- `FIXTURES_REQUIRED_INDIVIDUAL`: sets a `FIXTURES_REQUIRED` on all `blah_N` tests, with a test-specific name

The first two are basically doing what we are doing now (though in a more verbose way, and without requiring the `PROPERTIES` keyword. The second two add FIXTURES to the properties whose names are given by `${inputName}_npN_ompM` (regardless of whether the test name has `_npN_ompM` attached, for simplicity).

Example:

before:
```
$ ctest -N -R check_AVERAGE_np4
Test project /home/lbertag/workdir/scream/scream-src/branch/components/eamxx/ctest-build/full_debug/src/share/io/tests
  Test #25: output_restart_AVERAGE_test_np1
  Test #26: output_restart_AVERAGE_test_np2
  Test #27: output_restart_AVERAGE_test_np3
  Test #28: output_restart_AVERAGE_test_np4
  Test #32: output_restart_check_AVERAGE_np4

Total Tests: 5
```
after:
```
$ ctest -N -R check_AVERAGE_np4
Test project /home/lbertag/workdir/scream/scream-src/branch/components/eamxx/ctest-build/full_debug/src/share/io/tests
  Test #28: output_restart_AVERAGE_test_np4
  Test #32: output_restart_check_AVERAGE_np4

Total Tests: 2
```

## Testing

I tested this feature in the scream repo, and it seems to be working correctly.

### DIFF
--- a/cmake/EkatCreateUnitTest.cmake
+++ b/cmake/EkatCreateUnitTest.cmake
@@ -16,8 +16,26 @@ set(CUT_EXEC_MV_ARGS
   COMPILER_C_FLAGS COMPILER_CXX_FLAGS COMPILER_F_FLAGS
   LIBS LIBS_DIRS LINKER_FLAGS)
 
+# NOTE: for FIXTURE properties, the COMMON suffix denotes a property that has
+#       to be set on all RANKS/THREADS combinations *with the same name*,
+#       while the INDIVIDUAL suffix denotes a property that has to be set on
+#       a per-test basis. In the latter case, the actual FIXTURE name set
+#       on each test will have the "_npN_ompM" suffix attached.
+#       This can help if the tests setup some fixture (e.g., an output file)
+#       that a downstream test checks. Without this options, *all* ranks/threads
+#       combinations would have to complete (successfully) before the single
+#       downstream test can check their results. With this options, we can have
+#       one individual downstream test for each rank/thread combination,
+#       allowing the user to debug possible errors with a test more quickly.
+# NOTE: we do not offer the FIXTURES_CLEANUP option, since it makes no sense
+#       to have multiple tests cleaning up a FIXTURE. If a certain test is
+#       in charge of cleanup (and it must be a single rank/thread combination)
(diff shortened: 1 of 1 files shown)


jgfouca (approved):
Nice!

bartgol:
@jgfouca Weaver is having issues, and the AT is out until Monday due to gitlab-ex being down. I manually ran the jenkins test jobs on mappy/blake, and they pass. Given that this is a cmake mod (not a source mod), are you ok merging?

PR #293, opened 2023-05-11
Mild upgrades to string utils

bartgol:
## Motivation

The `split` utility did not allow to specify a string (rather than a char) as a delimiter. This PR allows that. I also added the `starts_with` util (C++20 introduces it as a method of basic_string, so it will be obsolete one day). Although the code reduction is very limited, it makes for a clearer intent when reading the code.

## E3SM Stakeholder Feedback

Needed in EAMxx.

## Testing

Added tests for new features in `tests/utils/string_utils`.

### DIFF
--- a/src/ekat/util/ekat_string_utils.cpp
+++ b/src/ekat/util/ekat_string_utils.cpp
@@ -32,6 +32,10 @@ std::vector<std::string> split(const std::string& str, const std::string& delim)
   return tokens;
 }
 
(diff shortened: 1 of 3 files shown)


tcclevenger (approved):

bartgol:
Nothing in this PR depends on Cuda. Since mappy/blake passed, I'm going to merge. It may take some time for weaver testing to be back to a healthy state.

PR #296, opened 2023-07-24
Adding a flag to enable/disable EKAT's Fortran support

jeff-cohere:
## Motivation

In order to make EKAT easier to use for a wider range of E3SM developers, I think it would be good not to require a Fortran compiler in those environments in which it's not needed. This probably isn't very interesting to a lot of projects just yet, but the MAM4 C++ port doesn't actually have any Fortran in it, and we're interested in providing a development environment with only the essentials for our task. This PR adds an optional called `EKAT_ENABLE_FORTRAN` that allows you to choose whether you want to build EKAT's Fortran support and run the Fortran-related BFB tests.

"Why not just live with the Fortran code and make you you have a working Fortran compiler on your system?" I hear you ask. Good question. In a perfect world, everyone would use a Linux development environment, in which all these things are *usually* easy. In practice, many people have Macs on which they like to develop, and many others have Linux machines that are maintained by their IT departments, which creates all kinds of edge cases that make every additional thing a support issue. Also, as you can see in this PR, there are really only a few spots where EKAT makes use of Fortran, at least so far, and it wasn't very difficult to find them and put guards around them.

This is just a suggestion (with solution attached), so if we don't want to go in this direction, that's also fine. By default, EKAT's Fortran support is enabled, so we shouldn't see any difference in the testing output (unless I've messed up somewhere).

## Testing

We are planning to build EKAT with `-DEKAT_ENABLE_FORTRAN=OFF` in our Haero/mam4xx test configurations for EAGLES, so this feature (like `EKAT_ENABLE_MPI`) is tested in our environment.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,8 +18,14 @@ include(EkatBuildKokkos)
 include(EkatBuildYamlCpp)
 include(EkatBuildSpdlog)
 
-# EKAT has mostly C++, but some Fortran too
-project (EKAT C CXX Fortran)
+# EKAT has mostly C++, but (optionally) Fortran too
+option (EKAT_ENABLE_FORTRAN "Whether EKAT Fortran support is enabled." ON)
+enable_language(C)
+enable_language(CXX)
+if (EKAT_ENABLE_FORTRAN)
+  enable_language(Fortran)
+endif()
+project (EKAT)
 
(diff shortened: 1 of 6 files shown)


jgfouca (approved):

jeff-cohere:
Looks like the build timed out on Blake.

PR #297, opened 2023-07-24
HIP: Omit aggressive vectorization flag

sarats:
While auditing all machine configuration options on Frontier, encountered this flag.
Evidently, this is not used by the GPU backends. Removed to avoid confusion.

Per Slack conversation,
> @ambrad : The vec flag triggers ivdep on CPU on ThreadVectorRange loops. I don't think any of the GPU backends do anything with it.


sarats:
@oksanaguba concurs with removal of this.

@bartgol Re-assign to whoever takes care of these on the EKAT side.

PR #300, opened 2023-08-16
Make YAML parsing an optional feature

bartgol:
## Motivation

Some applications do not need YAML, so no point in building this stuff. This is actually needed by MMF folks, who use EKAT but no yaml, since they are hitting some link error due to another library building YAML internlly. Unfortunately, the other library (ADIOS), does not ship its tpls nicely, and instead consolidate all tpls inside its main library. This makes it impossible for EKAT to even try to use an existing installation of the library.

## E3SM Stakeholder Feedback

Needed by MMF folks on Summit.

## Testing
Manually verified it works.


jeff-cohere:
Yeah, I saw this issue on Slack as well. This is a good fix.

PR #301, opened 2023-08-17
Add support for tags in YAML parsing

bartgol:
This allows the user to request the value to be stored with a specific type.

A big advantage is that we can have empty sequences stored with the correct type.

## Motivation

The main motivation for this is to be able to specify empty sequences of a specific type. So far, an entry of the form `foo: []` would be parsed as a sequence. But the parser deduces the inner type by trying to cast all the entries to, in order, `bool`, `int`, `double`, and `std::string`. Since an empty sequence has nothing to cast, the parser has nothing to check, and the first type (bool) was used.

With this PR, we can provide a tag to a node, to force the parser to use the desired type. The parser accepts `!bool`, `!int`, `!double`, and `!string`. Hence, a user can provide an empty list and force its inner type to be, say, `double` by using the syntax `foo: !double []`.

## Testing

In the yaml file parsed in the unit tests I added the line `ints_as_strings: !string [1,2]`, and verified it gets stored as `std::vector<std::string>`. I also added an empty sequence with `!double` tag, and verified it is stored as `std::vector<double>`, and the stored size is 0.

### DIFF
--- a/src/ekat/io/ekat_yaml.cpp
+++ b/src/ekat/io/ekat_yaml.cpp
@@ -106,15 +106,40 @@ void parse_node<YAML::NodeType::Scalar> (
   // Extract scalar as string, then try some casts
   // First int, then double, and finally fall back to string
   std::string str = node.as<std::string>();
+  const auto& tag = node.Tag();
 
-  TypeListFor<values_t>([&](auto t) -> bool{
-    using vtype = decltype(t);
-    if (is_type<vtype>(str)) {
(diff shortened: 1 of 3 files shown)


jeff-cohere (approved):
Looks good! I didn't know about YAML tags.

bartgol:
> Looks good! I didn't know about YAML tags.

Me either. It was a pleasant discovery. Apparently there are a bunch of special characters that trigger some special behavior in YAML. Had to go over the docs to find out which one was right for our case.

PR #302, opened 2023-08-22
Change tag names used in YAML parser

bartgol:
## Motivation

YAML already supports special tags for built-in types. This PR avoid confusion (and enhances interoperability with existing codes) by using native tags `!!bool`, `!!int`, `!!float`, and `!!str` for boolean, integer, floating point, and string parameters. Notice the double !! for the YAML built-in tags; `!!blah` actually is a short for `tag:yaml.org,2002:blah`. For arrays of parameters, we changed the tag names to be plural, for clarity and to minimize bugs.

## Testing

Modified (and improved) the tests accordingly.


jeff-cohere (approved):
👍

PR #303, opened 2023-08-24
Fix bug in YAML parser

bartgol:
Non-specific tag uses a single exlamation mark, not a question mark

## Motivation

Testing in EAMxx revealed that explicitly quoted strings do get the tag `!` (a single exlamation mark). Since quotes were explicitly used, such entries should be parsed as strings.

## Testing

Modified the yaml parser unit test to exercise this use case.


jeff-cohere (approved):
You're becoming quite the YAML expert, Luca.

PR #304, opened 2023-08-24
In ParameterList::get method, print more info in case of wrong type

bartgol:
## Motivation

Currently, if the provided template type does not match the parameter real type, an exception is thrown inside the `any_cast` function, which however has no knowledge about the fact that this `ekat::any` instance comes from a parameter list, let alone the name of the parameter. This makes it complicated for the user to realize which parameter was accessed with the wrong type (or even that this error comes from a `ParameterList::get` call)

This PR adds a check on the template type (the same that `any_cast` would do) before attempting the cast. If the cast fails, it prints the name of the parameter, as well as the name of the enclosing parameter list.

## Testing

Added small test to verify exception is thrown, and expected error message is printed.

### DIFF
--- a/src/ekat/ekat_parameter_list.hpp
+++ b/src/ekat/ekat_parameter_list.hpp
@@ -103,8 +103,15 @@ inline T& ParameterList::get (const std::string& name) {
   // Check entry exists
   EKAT_REQUIRE_MSG ( isParameter(name),
       "Error! Key '" + name + "' not found in parameter list '" + m_name + "'.\n");
+  auto p = m_params[name];
+  EKAT_REQUIRE_MSG ( p.isType<T>(),
+      "Error! Attempting to access parameter using the wrong type.\n"
+      "   - list name : " + m_name + "\n"
+      "   - param name: " + name + "\n"
+      "   - param type: " + std::string(p.content().type().name()) + "\n"
+      "   - input type: " + std::string(typeid(T).name()) + "'.\n");
 
-  return any_cast<T>(m_params[name]);
+  return any_cast<T>(p);
 }
 
 template<typename T>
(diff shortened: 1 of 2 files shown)


jeff-cohere:
I like this. But have you tried this within EAMxx to see if it causes any unexpected issues? How permissive do you think we've been with typed YAML parameters?

bartgol:
> I like this. But have you tried this within EAMxx to see if it causes any unexpected issues? How permissive do you think we've been with typed YAML parameters?

I am not sure what kind of issue it could give. The check I added is exactly the same check that it would be performed inside the `any_cast` call. I did run eamxx, and was able to change the uninformative
```
 FAIL:
src.isType<ConcreteType>()
/home/lbertag/workdir/scream/scream-src/branch/externals/ekat/src/ekat/std_meta/ekat_std_any.hpp:141
Error! Invalid cast requested.
   - actual type:    NSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
   - requested type: i'. 
```
to
```
 FAIL:
p.isType<T>()
/home/lbertag/workdir/scream/scream-src/branch/externals/ekat/src/ekat/ekat_parameter_list.hpp:107
Error! Attempting to access parameter using the wrong type.
   - list name : output_control
   - param name: Frequency
   - param type: NSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
   - input type: i'.
```
while debugging another PR failure

jeff-cohere:
Okay--I got confused and thought maybe this fix involves more stringent type checking, but really it just provides more informative errors. Sounds great.

jeff-cohere (dismissed):

jeff-cohere (approved):

PR #305, opened 2023-10-19
Remove submodules, and use find_package + FetchContent instead

bartgol:
## Motivation

Submodules can quickly increase the cloning time, as well as the size of the repo. Things get more pronounced if EKAT is used as a submodule of other submodules of a project (e.g., E3SM has ekat as submodule, as well as MAM4xx, which also has ekat as a submodule. 

Furthermore, there really is no need to download submodule repos if the user already has an installation of the library. E.g., yaml-cpp is often (not always, but a growing number of times) provided on clusters. In that case, users should be able to simply point to that location.

This PR relies on `find_package` to first attempt to find an installation (hence, users can point to an existing installation by setting `<PackageName>_ROOT` env vars). If not found, we proceed to download and build the pakcage as part of EKAT (and will subsequently install it along side EKAT). In this process, Kokkos needs a bit more care: if an installation is found, before accepting it we must verify that the correct devices/archs/options are enabled (correct here means "the ones that were requested via Kokkos_ENABLE_XYZ when configuring EKAT).

## E3SM Stakeholder Feedback

This is desired in E3SM, and may make life easier in general. 
## Testing
Manually tested on my laptop.


bartgol:
Note: right now I am passing GIT options to `FetchContent_Declare`. If a tpl is available as a release on github, then we can provide the release tarball URL, which is in general faster to download than the actual repo. I know for sure this is not possible with Kokkos (though we could, probably), but maybe we should explore it for spdlog and yaml-cpp.

CMakeLists.txt
```diff
@@ -10,23 +10,20 @@ endif(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.12.0")
...
 )
 
-# Scripts containing macros needed to handle TPLs
-include(EkatBuildKokkos)
-include(EkatBuildYamlCpp)
-include(EkatBuildSpdlog)
-
 # EKAT has mostly C++, but (optionally) Fortran too
 option (EKAT_ENABLE_FORTRAN "Whether EKAT Fortran support is enabled." ON)
+message ("cm search path: ${CMAKE_PREFIX_PATH}")
```
jeff-cohere, line 18: Do you want to keep messages like this one?

bartgol: Whoops!

jgfouca (changes requested):
Luca, you did a lot of work here so I am sorry to say that I don't think this approach will work; I think these dependencies have to submodules. The issue is authentication on CI systems like Jenkins. The Jenkins Git plugin has a feature that allows you to re-use your main repo authentication (usually an ssh key) to grab submodules. Without this feature, arbitrary git clones over ssh will not work because the ssh keys are never installed in the jenkins entity account's ${HOME}/.ssh directory.

jeff-cohere:
Before I provide detailed feedback, I want to give one more plug to CMake's [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html) mechanism, which I feel is superior to in-tree embedding and FetchContent because it creates isolated project namespaces for external libraries, which decreases the amount of knowledge required to stitch everything together. But if we ever switched to that approach (which I've used for Haero and mam4xx), it would mean changing SCREAM and EKAT's build systems a bit, especially in how they interact with Kokkos.

bartgol:
> Luca, you did a lot of work here so I am sorry to say that I don't think this approach will work; I think these dependencies have to be submodules. The issue is authentication on CI systems like Jenkins. The Jenkins Git plugin has a feature that allows you to re-use your main repo authentication (usually an ssh key) to grab submodules. Without this feature, arbitrary git clones over ssh will not work because the ssh keys are never installed in the jenkins entity account's ${HOME}/.ssh directory.

I don't think we should put the needs of our jenkins testing above the quality of the build system we have. But I think it's a valid concern, so I want to try to circumvent it. I see two options:

- why can't we have jenkins add its own ssh keys? It seems like a natural thing to do for an entity account...
- If a tpl is available with a release, we can grab the tarball, which is done via curl/wget (not sure what cmake uses, but not ssh).

jgfouca:
@bartgol , there may be way to work around the issue but they will be disruptive. We are currently running Jenkins jobs on a lot of machines. When it comes to wget, it opens up other issues with http proxies and SSL issues. Again, nothing we can't workaround if we really want to, but I'm wondering if slightly faster clones is worth making a big change like this? I agree that the build system should allow custom installations of TPLs, I'm working on that now in E3SM, but I think that is independent from the issue of how the code is obtained if it needs to be built.

jgfouca:
Follow on thought, if most of this PR is going towards supporting pre-installed TPLs, then I think we can keep most of it and just make some minor changes to reinstate the submodules.

bartgol:
Yes, we can keep the submodules, and do find_package+add_subdirectory.

My main driver was complains we hard from e3sm about the exponential increase of submodules. EKAT is a submodule of E3SM, as well as MAM4xx, which is also a submodule of E3SM.

Regarding ssh: I find it hard to believe that jenkins does not allow to propagate credentials to the build step. I think the [SSH Agent plugin](https://plugins.jenkins.io/ssh-agent/) is what we're looking for? I may try to use it and see if it works.

jeff-cohere:
Are either of you aware of any mechanism that would allow us to conditionally activate/deactivate submodules based on the existence of installed libraries on a machine? I feel like if such a thing existed, it could alleviate a lot of the "clone" burden. But I haven't ever found any such thing.

jgfouca:
@jeff-cohere , there are the various hook scripts but Git doesn't let you push those. They'd have to be something the user manually installed in each repo.

jeff-cohere:
Hmm. There is [`git submodule deinit`](https://git-scm.com/docs/git-submodule#Documentation/git-submodule.txt-deinit-f--force--all--ltpathgt82308203):

> Unregister the given submodules, i.e. remove the whole submodule.$name section from .git/config together with their work tree. Further calls to git submodule update, git submodule foreach and git submodule sync will skip any unregistered submodules until they are initialized again, so use this command if you don’t want to have a local checkout of the submodule in your working tree anymore.

It might be worth experimenting with `git submodule init/deinit` within the build system in tandem with the determination of the presence of libraries. It doesn't seem like this command affects the repo itself, but one never knows! I understand if this seems a bit too magical, though.

bartgol:
But by the time you run that command, we are already in the build step, so we either already downloaded the submodule, or we may need to download it (in which case we're back to the FetchContent issue)

I tried using SSH Agent plugin and it seems to work. The build does use the provided ssh key for ssh ops during build step.

jgfouca:
Interesting that you were able to get SSH Agent to work. We could turn off strict host key checking globally with a minor change to E3SM_test_scripts.

bartgol:
> Interesting that you were able to get SSH Agent to work. We could turn off strict host key checking globally with a minor change to E3SM_test_scripts.

We can just do it in the build step of each jenkins job, and only if we see that (for some reason) the agent does not work.

I mean, it makes no sense to get that error, cause it's got nothing to do with which ssh key we're using to connect. It only has to do with how ssh checks host keys, and I thought that was gonna be the same for "pre-build-phase" ssh (via git plugin) and "build-phase" ssh (via manual git commands). The fact that later on it was working without hacking host key verification makes me think it is an orthogonal issue to the ssh key used..

bartgol:
Ah, on mappy we still see the host key verification error. It is a mystery to me. I looked at sshd_config on mappy and I don't see any bad setting. I don't know why we get this error on mappy and not on weaver. I will set the GIT_SSH_COMMAND env var on the mappy ekat PR testing job.

bartgol:
Interesting: I actually put the `GIT_SSH_COMMAND` env var in the post-build step shell script. However, the job ran when I clicked on "rebuild last". I am not sure what's the difference between when I click it and when it gets triggered. I don't think there should be any...

jgfouca:
@bartgol , I forgot we also have some gitlab CI running jobs at ORNL.

bartgol:
> @bartgol , I forgot we also have some gitlab CI running jobs at ORNL. 

Ok, then I will close this or and open one where the submodules dirs are added only after an unsuccessful find_package call. We can revisit the reduction of the number of singles another time

PR #306, opened 2023-10-23
Refactor handling of TPLs

bartgol:
## Motivation

Currently, Ekat does not allow using existing installations for its TPLs (except for Kokkos). This PR fixes that, as well as improves a few things. Summary of changes
- For all TPLs, try the `find_XYZ` route first. If that fails, proceed to build locally. Note: we want to use `find_package`, but `yaml-cpp` appears to not ship a CMake config file, so we must find libs/headers separately.
- If building locally, always use the submodule version. Now that we allow pointing to an existing installation, it seems overkill to allow users to provide an ad-hoc source folder. If you have a src folder you'd like to try, install it first, then set the <PackageName>_ROOT env var.
- Changes to `EkatConfig.cmake`:
  - Correctly handle cases where tpls were found vs locally built
  - Add REQUIRED to ensure CMake craps out if the tpl is not found
  
When using `find_package` for Kokkos,  we accept the installation found (if any) only if the settings match what was enabled in the cache. E.g., if building Ekat with `Kokkos_ENABLE_CUDA=ON`, any installation that has `Kokkos_ENABLE_CUDA=OFF` (or not set) is rejected.
 

## E3SM Stakeholder Feedback

There have been issues with ADIOS when building MMF (interestingly, not when building SCREAM), where ADIOS would link one installation of yamlcpp, and ekat was building its own, leading to duplicated symbols. We changed ekat dep on yaml-cpp to be optional, but that is not an option for EAMxx. Being able to point to an existing yaml installation may help with this.

## Testing

I tested locally on my machine that Ekat is found and used correctly. Since CMAKE_INSTALL_PREFIX is part of the paths scanned by `find_package`, configuring Ekat a second time after installation will find deps in the install dir, so this verifies that finding a package works. If you need to change the installation (e.g., you update a submodule), you can nuke the install dir to prevent CMake from picking it up during `find_package`.


jgfouca (approved):

jeff-cohere (dismissed):
Thanks, Luca!

bartgol:
I will do a few test with EAMxx, and then merge.

bartgol:
I tested more scenarios (use some package from submodule, not others) and uncovered a few extra things to fix. Also, yaml-cpp _does_ ship as a CMake package, it's just that 1) it put its config file in a different folder from the one I was looking in, and 2) the presence of a hyphen in the pkg name is throwing off cmake on my machine (CMake 3.26.3), even though it _should_ convert `-` to `_` when forming the `<PACKAGENAME>_ROOT` cmake/env var names to use in `find_package`.

I tested this version of the PR with a downstream miniapp. I verified that it works when
- all submodules are built by ekat
- all submodules are found elsewhere
- some submodules are found and some are built by ekat.

jeff-cohere (approved):
I really wish CMake would establish better conventions for upper/lower-case variables in these situations. It's so needlessly complicated as it is in practice.

PR #312, opened 2023-11-15
Do not use new find_package option with old cmake

### LINKED ISSUE #311: EKAT's use of `find_package` for Kokkos requires CMake 3.24
**Describe the bug**

When I try to build EKAT in a standalone configuration, it uses `find_package` to look for Kokkos, providing it with a number of parameters to limit its search. One of these parameters, `NO_CMAKE_INSTALL_PREFIX`, appears not to be available in versions of CMake before 3.24. The version of CMake that EKAT claims it needs is 3.3.

Ordinarily this wouldn't be a big deal to fix--we could update the required version of CMake for EKAT to 3.24. However, this version of CMake doesn't seem to have made it to various systems yet, so I don't think we should be quite as aggressive with our use of CMake features. I recommend we remove the offending parameter from this package search until CMake 3.24 becomes more prevalent. Hopefully this wouldn't break anything.

**To Reproduce**

Try to build EKAT in a standalone configuration using a version of CMake < 3.24. This results in an error message during the configuration step:

```
-- Looking for a Kokkos installation ...
CMake Error at ext/ekat/cmake/tpls/EkatFindKokkos.cmake:47 (find_package):
  find_package called with invalid argument "NO_CMAKE_INSTALL_PREFIX"
Call Stack (most recent call first):
  ext/ekat/cmake/tpls/EkatTPLs.cmake:44 (include)
  ext/ekat/CMakeLists.txt:37 (include)
```

**Expected behavior**
EKAT should configure itself correctly using any version of CMake that complies with our specified requirements.

bartgol:
## Motivation

The cmake logic to find Kokkos is using a cmake 3.24 issue. Since we don't want to push for such a recent version as a minimum required, use that option only if cmake version is recent enough.

## Related Issues

* Closes #311

### DIFF
--- a/cmake/tpls/EkatFindKokkos.cmake
+++ b/cmake/tpls/EkatFindKokkos.cmake
@@ -44,15 +44,28 @@ endforeach()
 
 # Try to find the package
 message (STATUS "Looking for a Kokkos installation ...")
-find_package(Kokkos QUIET
-  # Skip all cmake defaults paths, except for env/cmake vars Kokkos_ROOT (if any)
-  NO_CMAKE_PATH
-  NO_CMAKE_ENVIRONMENT_PATH
-  NO_SYSTEM_ENVIRONMENT_PATH
-  NO_CMAKE_PACKAGE_REGISTRY
-  NO_CMAKE_SYSTEM_PATH
-  NO_CMAKE_INSTALL_PREFIX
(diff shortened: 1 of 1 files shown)


jeff-cohere (approved):
:+1:

jeff-cohere (approved):
:+1:

PR #313, opened 2023-11-16
Ensure that yaml is always linked using the scoped target name

bartgol:
## Motivation

On Chrysalis, if e3sm-unified is loaded, we were having issues with yaml-cpp includes. The issue was that the imported cmake target was "yaml-cpp::yaml-cpp", while we were linking "yaml-cpp". Since that wasn't the name of the target, CMake was simply adding `-lyaml-cpp` to the link line, without importing the target properties, such as the include dirs.

This PR ensures that, regardless of whether we build yaml-cpp in house or we import it via find_package, the target `yaml-cpp::yaml-cpp` is always defined, and it is what we link against.

## E3SM Stakeholder Feedback

Should fix some EAMxx v1 build issues on chrysalis.


bartgol:
Testing it on chrysalis now.

bartgol:
I had to add a fix regarding usage of `try_compile` for CMake versions older than 3.25. But with this branch, I can successfully build EAMxx cases on chrysalis, with e3sm-unified loaded.

jgfouca (approved):

PR #316, opened 2024-01-10
Fix GetMpiDistributionName

bartgol:
## Motivation

The error message in the function was already hinting at the issue, which was going to bite us at some point: one had to call `find_pacakge(MPI ...)` _within a scope visible from the scope where `GetMpiDistributionName` was called_. That's because the function checked which CMake var `MPI_<LANG>_FOUND` is `TRUE` to establish which source file to use with `try_compile`, but this vars, set by `find_package` are not cached, so they disappear as soon as we exit the scope where `find_package` was called.
In EAMxx, we let Ekat find the MPI package, so when control comes back to eamxx, we are out of scope, and those vars above are undefined when we call `DisableMpiCxxBindings` (which calls `GetMpiDistributionName`. Sure, we could add a call to `find_package(MPI)` in EAMxx's main `CMakeLists.txt` file, right before `DisableMpiCxxBindings`, but that seems to make things more complicated.

The solution is simple: we don't need to check if `MPI_<LANG_FOUND` is true. Instead, we simply call `find_package(MPI REQUIRED COMPONENTS <LANG>)` once we detect that `CMAKE_<LANG>_COMPILER` is set. Recall that `find_package` caches all the important stuff (libs, lib dirs, targets, include dirs,...), so subsequent calls are no ops anyways. So if MPI was already found, it doesn't cost anything to call this again.

## E3SM Stakeholder Feedback

Required by EAMxx
## Testing

Manually verified that EAMxx builds with this mod (well, it goes past the error, other things need to be adjusted in EAMxx to use current EKAT's master).

### DIFF
--- a/cmake/EkatMpiUtils.cmake
+++ b/cmake/EkatMpiUtils.cmake
@@ -1,20 +1,19 @@
 # Detect the library that provides MPI
 set (EKAT_CMAKE_DIR ${CMAKE_CURRENT_LIST_DIR})
 macro (GetMpiDistributionName DISTRO_NAME)
-  if (CMAKE_CXX_COMPILER AND MPI_CXX_FOUND)
+  if (CMAKE_CXX_COMPILER)
+    find_package (MPI REQUIRED COMPONENTS CXX)
     set (LINK_LIB MPI::MPI_CXX)
     set (SOURCE_FILE ${EKAT_CMAKE_DIR}/TryCompileMPI.cxx)
-  elseif (CMAKE_C_COMPILER AND MPI_C_FOUND)
+  elseif (CMAKE_C_COMPILER)
+    find_package (MPI REQUIRED COMPONENTS C)
     set (LINK_LIB MPI::MPI_C)
     set (SOURCE_FILE ${EKAT_CMAKE_DIR}/TryCompileMPI.c)
   else ()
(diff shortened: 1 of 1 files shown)


cmake/EkatMpiUtils.cmake
```diff
@@ -1,20 +1,19 @@
 # Detect the library that provides MPI
 set (EKAT_CMAKE_DIR ${CMAKE_CURRENT_LIST_DIR})
 macro (GetMpiDistributionName DISTRO_NAME)
-  if (CMAKE_CXX_COMPILER AND MPI_CXX_FOUND)
+  if (CMAKE_CXX_COMPILER)
+    find_package (MPI REQUIRED COMPONENTS CXX)
```
jeff-cohere, line 5: Do we want `CXX` components for MPI? Have the C++ MPI bindings ever been anything but trouble? I suppose we disable them elsewhere, so maybe this doesn't matter.

jeff-cohere (approved):
Thanks for fixing this, Luca! I left a minor note, but this looks great to me.

cmake/EkatMpiUtils.cmake
bartgol: That's a good point. I will change this to C component.

PR #317, opened 2024-01-10
Use check_symbol_exists instead of try_compile to detect MPI distro name

bartgol:
@jeff-cohere sorry, I did not push this commit in time to beat the AT.

I think using `check_symbol_exists` is more solid, since we don't run the risks of linking against `libmpi_cxx`, as you pointed out in the previous PR. I could not simply link against the `MPI::MPI_C` target, since the host project may not have enabled the C language. Since all we need is to check which macro is defined in the header, this approach is probably more reliable.


jeff-cohere (approved):
I like it! Idly, I wonder whether this means we can get rid of the logic to explicitly disable C++ MPI bindings, but somehow I doubt it. :-)

bartgol:
> I like it! Idly, I wonder whether this means we can get rid of the logic to explicitly disable C++ MPI bindings, but somehow I doubt it. :-)

I took off the automerge label while I quickly inspected this (while at it, might as well...).

I think I also doubt that's the case. The "skip cxx bindings" stuff is needed when `mpi.h` is parsed. It contains stuff like this (for OMPI)
```c++
#if !defined(OMPI_SKIP_MPICXX) && OMPI_BUILD_CXX_BINDINGS && !OMPI_BUILDING
#if defined(c_plusplus) || defined(__cplusplus)
#include "openmpi/ompi/mpi/cxx/mpicxx.h"
#endif
#endif
```
I suppose even if you don't define `OMPI_SKIP_MPICXX`, you can still avoid the cxx bindings if the other macros have the right values. However, in order to avoid surprises, I think it's best to keep the logic, so that we don't accidentally include `mpicxx.h`, without later linking against it.

bartgol:
Ugh, running manually the cmake command printed by the jenkins job does not reproduce the fail. But I can reproduce it by manually running jenkins.sh. Perhaps something gets amiss with env vars.

jeff-cohere (approved):
Nice work!

PR #318, opened 2024-01-25
Change default constructor of Pack to init to 0

bartgol:
## Motivation

While leaving entries to be invalid seems like a nice idea (allows to track uninited values better), it plays badly with Kokkos parallel_scan, where the accumulation variable is initialized as `ValueType accum = ValueType();`, which _assumes_ that the value initialization inits to 0 (or, to be precise, to the identity value for the sum operation). It would be nice if Kokkos did
```
  ValueType accum = reduction_identity<ValueType>::sum();
```
but this may break existing code (e.g., code that uses some `ValueType` for which there is no specialization of `reduction_identity`). I still think they _could_ use some SFINAE magic to use `reduction_identity` if the right specialization exists, and resort to `ValueType()` otherwise. Alas, I am not a kokkos dev.

So the solution is to make our default constructor init to 0.

I threw in also the change of a few ctor, to call the corresponding `set` method, to avoid duplication of code.

## E3SM Stakeholder Feedback

The PR in E3SM that updates to Kokkos 4.2 has exposed this issue. In the version of Kokkos in current master, the accum var is inited as `value_type accum = 0`, which works correctly. But updating to Kokkos 4.2 will require this change.
## Testing

I tested EAMxx with the Kokkos 4.2 branch on chrysalis. Without this change, I get NaN's from scan operation, while with this change I get correct numbers.

### DIFF
--- a/src/ekat/ekat_pack.hpp
+++ b/src/ekat/ekat_pack.hpp
@@ -176,7 +176,7 @@ struct Pack {
   KOKKOS_FORCEINLINE_FUNCTION
   Pack () {
     vector_simd for (int i = 0; i < n; ++i) {
-      d[i] = ScalarTraits<scalar>::invalid();
+      d[i] = scalar(0);
     }
   }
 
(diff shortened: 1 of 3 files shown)


jgfouca (dismissed):

bartgol:
@jgfouca I killed the weaver job cause it was doing the cleanup action, which, due to the FS issues, was taking forever. But the build passed, as you can see from the log. If you approve, go ahead and merge using the force.

PR #319, opened 2024-01-30
Update YAML to remove warnings

jgfouca:
Update YAML subm to remove warning flags.

## Motivation

Some of them were breaking Nvidia builds.


bartgol:
This PR was tested while also e3sm was tested. It probably took long b/c the case.build phase uses the whole machine, so the 2 jobs collided.

PR #320, opened 2024-01-30
Fix GetMpiDistributionName for Cray mpi compilers

bartgol:
## Motivation

Cray mpi compilers wrappers do not support the `-show` option, like `mpicxx` does. So in order to get the compiler line that shows the include paths for MPI headers, we need to use a Cray-specific option. Simply pick the option to use based on whether `CRAYPE_VERSION` is defined in the env.

## E3SM Stakeholder Feedback

Needed in order to use current EKAT on Perlmutter.
## Testing

I verified with EAMxx that with this branch CMake runs successfully on pm-cpu.

### DIFF
--- a/cmake/EkatMpiUtils.cmake
+++ b/cmake/EkatMpiUtils.cmake
@@ -32,7 +32,13 @@ macro (GetMpiDistributionName DISTRO_NAME)
     else()
       set (COMPILER ${MPI_Fortran_COMPILER})
     endif()
-    execute_process (COMMAND ${COMPILER} -show OUTPUT_VARIABLE TEMP)
+
+    if (DEFINED ENV{CRAYPE_VERSION})
+      # Cray wrappers have different options from mpicxx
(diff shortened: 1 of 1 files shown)


jeff-cohere (approved):
LGTM, pending successful testing. ;-)

bartgol:
Weaver is quite slow lately, and it seems it failed after testing completed (all pass), and right before installing the SP build. Still, to be extra safe, I will let the AT re-run the tests (I increased the jenkins job timeout to make sure it completes). Adding automerge so this goes in quickly.

jgfouca (approved):

bartgol:
The fail on Weaver is a timeout, caused by the increasingly problematic filesystem (35min to install the headers/libs!).

The testing had passed, and most of the installation was completed. Also, the DP build had installed with no problems, so I'm very confident everything is fine. I'm going to manually merge this.

PR #322, opened 2024-02-02
Add missing quotes around FIXTURES options in create unit test utilities

bartgol:
## Motivation

Without quotes, there is no way to pass 2+ fixtures to the routines.

## Testing

I manually tested it in EAMxx.


jgfouca (approved):

jgfouca:
LGTM. Do we want this included in the EKAT update for scream?

bartgol:
> LGTM. Do we want this included in the EKAT update for scream?

We may as well.

PR #326, opened 2024-02-13
Allow to call scalarize on non-packed views

bartgol:
## Motivation

Since
 - if we have a piece of code that needs non-packed views, calling `scalarize` is the easy solution;
 - we write generic code templated on scalar types, which should be usable with either pack types or scalar types
it is desirable for `scalarize` to work also on views that are already scalarized (meaning that `value_type` is not a pack). Ideally, `scalarize` should be an "idempotency", meaning that `scalarize(scalarize(v))=scalarize(v)`. This is, of course, assuming that `v` is a view of packs of a scalar type, and not a "double packed" view like`View<Pack<Pack<T,N>,M>*>` (in that case, applying `scalarize` once should return `View<Pack<T,N>*>` and a second application of `scalarize` should return `View<T*>`).

This PR extends the scalarize utilities to support this.

## E3SM Stakeholder Feedback

I think there are some areas in EKAT and EAMxx that could benefit from being able to call `scalarize(v)` in some generic code areas.
## Testing

Added tests that verify that `scalarize(scalarize(v))==scalarize(v)`.

### DIFF
--- a/src/ekat/ekat_pack_kokkos.hpp
+++ b/src/ekat/ekat_pack_kokkos.hpp
@@ -241,9 +241,14 @@ typename std::enable_if<NewPackT::packtag && OldPackT::packtag &&
 repack_impl (const Kokkos::View<OldPackT**, ViewProps...>& vp) {
   constexpr auto new_pack_size = NewPackT::n;
   constexpr auto old_pack_size = OldPackT::n;
-  static_assert(new_pack_size > 0 &&
-                old_pack_size % new_pack_size == 0,
-                "New pack size must divide old pack size.");
+  static_assert(new_pack_size > 0, "New pack size must be positive");
+
+  // It's overly restrictive to check compatibility between pack sizes.
+  // What really matters is that the new pack size divides the last extent
+  // of the "scalarized" view.
+  // This MUST be a runtime check.
+  assert ( (vp.extent_int(1)*old_pack_size) % new_pack_size == 0);
+
   return Unmanaged<Kokkos::View<NewPackT**, ViewProps...> >(
     reinterpret_cast<NewPackT*>(vp.data()),
     vp.extent_int(0),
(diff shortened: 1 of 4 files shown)


bartgol:
@jgfouca I decided to add also the equivalent for repack. My goal is to allow code that is templated on scalar type to call scalarize/repack without having to first check that the input view is indeed a view of packs.

Note: while for scalarize it's always ok, repack needs to check that the requested pack size is compatible with the view's extents. On a side node, I think we were not doing the right checks. In particular, starting from a pack size of 1, all our compile-time checks passed, but of course, one cannot repack to an arbitrary pack. The correct check is a runtime one, and needs to check that the pack size divides the extent of the scalarized view.

jgfouca (approved):

src/ekat/ekat_pack_kokkos.hpp
```diff
@@ -205,13 +271,17 @@ typename std::enable_if<NewPackT::packtag && OldPackT::packtag &&
...
+  // We actually need to check that the new pack size divides the last extent
+  // of the "scalarized" view.
+  // This MUST be a runtime check.
+  assert ( (vp.extent_int(1)*old_pack_size) % new_pack_size == 0);
+
   return Unmanaged<Kokkos::View<NewPackT**, ViewProps...> >(
     reinterpret_cast<NewPackT*>(vp.data()),
     vp.extent_int(0),
-    (new_pack_size / old_pack_size) * vp.extent_int(1));
+    vp.extent_int(1) / (new_pack_size / old_pack_size) );
```
bartgol, line 284: This was backward. I guess we were never using it. I'm a bit puzzled as of why our tests were running fine, but I won't lose sleep over it.

src/ekat/ekat_pack_kokkos.hpp
```diff
@@ -229,11 +299,31 @@ repack_impl (const Kokkos::View<OldPackT**, ViewProps...>& vp) {
...
+    Unmanaged<Kokkos::View<typename RepackType<N,OldPackT>::type**,ViewProps...>>
+>::type
 repack (const Kokkos::View<OldPackT**, ViewProps...>& vp) {
   return repack_impl<typename RepackType<N,OldPackT>::type>(vp);
 }
 
+// 2d pack from a scalar view
+template <int N, typename ScalarT, typename... ViewProps>
+KOKKOS_FORCEINLINE_FUNCTION
+typename std::enable_if<!IsPack<ScalarT>::value,
```
bartgol, line 312: This is why I needed `IsPack`. For the "positive" case I could have used `OnlyPackReturn<Unmanaged<...>>`, but for the negative case I had no tool to use. I could not do
```
enable_if<not ScalarT::packtag>
```
cause that fails if ScalarT is not a pack (which is the case I need).

jgfouca (approved):

PR #327, opened 2024-02-14
Some optimizations to LinInterp main method

bartgol:
## Motivation

I didn't like the fact that the main interpolation method had a big if/else, reducing readability. Also, the `index_and_shift` function is a bit cryptic (to me) when I see it. So I decided to see if I could clean up a bit, perhaps unifying the two branches. I ended up achieving some small perf improvements. For a 10 col, 128 levs benchmark, running 100 samples per run, and running the code 10 times, I ended up clocking the `lin_interp` method at ~270ms for this branch, versus ~350ms for master. This was on my laptop (Intel Tiger Lake), without threads. Of course, I need to verify perf does not deteriorates on GPU and other machines.

Likely, the most important mods are:
- no `end_mask` and `not_end` temp masks. I do need the extra `k1ph` int pack, but that's compensated by the fact that `k1` (same as `indx_pk` in master) is a ref rather than a copy.
- temp packs creation moved outisde loop (I did this in master too, and it already improved perf)
- break packed update formula in multiple steps, to minimize temps creation

### DIFF
--- a/src/ekat/util/ekat_lin_interp_impl.hpp
+++ b/src/ekat/util/ekat_lin_interp_impl.hpp
@@ -90,36 +90,67 @@ KOKKOS_INLINE_FUNCTION
 void LinInterp<ScalarT, PackSize, DeviceT>::lin_interp_impl(
   const MemberType& team,
   const RangeBoundary& range_boundary,
-  const view_1d<const Pack>& x1, const view_1d<const Pack>& x2, const view_1d<const Pack>& y1,
-  const view_1d<Pack>& y2,
+  const view_1d<const Pack>& x1,
+  const view_1d<const Pack>& x2,
+  const view_1d<const Pack>& y1,
+  const view_1d<      Pack>& y2,
   const Int col) const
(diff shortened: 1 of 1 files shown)


bartgol:
On weaver, I had to time the whole kernel_dispatch+setup+lin_interp combo, rather than _just_ the call to `lin_interp` (like I did on my laptop). But since everything else is the same, I think it's fair to assume that any difference in timing is to be attributed to lin_interp impl.

I ran on 1GPU with 200 cols and 128 levs, with randomized data. I ran the executable 100 times, with each run doing 100 experiments. I threw away the longest and shortest run, and computed an average. The branch seems to perform within 1% of master (sometimes faster, sometimes slower), so we can conclude that perf is not impacted on GPU.

I ran the same caimpaign on mappy, and this branch is measurably faster: 65ms vs 74ms, which is a bit less than what I observed on my laptop, but still noticeable.

jgfouca (approved):
lgtm

PR #328, opened 2024-02-14
Revert one change in LinInterp

bartgol:
## Motivation

Automatic variables created outside a nested pfor cannot be written onto from inside the pfor. See [this](https://kokkos.org/kokkos-core-wiki/ProgrammingGuide/HierarchicalParallelism.html#team-loops) kokkos wiki paragraph on this. Thanks @tcclevenger for catching this!

That said, with 1 thread on CPU, I _did_ notice the code being faster creating the temps before the loop, and recycling them inside. It would be nice to maybe use scratch memory for this stuff, but Pack's cannot be created from pre-allocated memory (as of now).

### DIFF
--- a/src/ekat/util/ekat_lin_interp_impl.hpp
+++ b/src/ekat/util/ekat_lin_interp_impl.hpp
@@ -104,9 +104,10 @@ void LinInterp<ScalarT, PackSize, DeviceT>::lin_interp_impl(
 
   const int i = col == -1 ? team.league_rank() : col;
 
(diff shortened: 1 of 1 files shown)


tcclevenger (approved):

bartgol:
For context, after chatting with kokkos devs, I think the code was working correctly _as long as the policy vector length is one_. That is the case for the places in EAMxx that use LinInterp. However, for generality, and to avoid subtle bugs to find, it was best to stick with correct usage.

PR #329, opened 2024-02-15
Add update method for packs

bartgol:
## Motivation

This snippet is a good example of why this feature would be helpful:
```c++
    auto& T_k = some_view(k); // T_k is a pack
    T_k = some_func();

    // Update: at_top, at_bot, and at_mid are "complementary" masks
    T_k.set(at_top, T_k - coeff*pack2);
    T_k.set(at_bot, T_k - coeff*pack3);
    T_k.set(at_mid, T_k - coeff*(pack2+pack3));
```
Each call to `set` is creating temporaries on the fly, to store the result of the pack to be used as rhs where the mask is true. Compare this with
```c++
    auto& T_k = some_view(k); // T_k is a pack
    T_k = some_func();

    // Update: at_top, at_bot, and at_mid are "complementary" masks
    T_k.add(not_at_bop, pack2, -coeff);
    T_k.add(not_at_top, pack3, -coeff);
```

This PR adds two sets of methods, `update` and `add`. The former is more general, and allows to do `y = beta*y + alpha*x`, while the latter is a shortcut for `y += alpha*x`. Both methods have 3 overloads: one for when the update is to be performed on the whole pack, and two for a masked update.

Notice that the set method is now a particular case of `y.update(mask,x,1,0)`. However, I'm not going to remove it, b/c a) it's clearer to see `a.set(m,val)` than `a.update(m,val)` or `a.update(m,val,1,0)`, and b) set does not have to do additions/multiplications, so it's pointless to resort to an impl that does them.

## Testing

Added small unit tests.

### DIFF
--- a/src/ekat/ekat_pack.hpp
+++ b/src/ekat/ekat_pack.hpp
@@ -292,6 +292,58 @@ struct Pack {
     return *this;
   }
 
+  // In the update methods, return *this, so we can pipe updates.
+  // By default, the update uses alpha=1 and beta=0, which means it's the equivalent of op= and set methods
+  template<typename T, typename CoeffT = scalar>
+  KOKKOS_FORCEINLINE_FUNCTION
+  Pack& update (const Pack<T,n>& x, const CoeffT alpha = 1, const CoeffT beta = 0) {
+    vector_simd
+    for (int i=0; i<n; ++i) {
+      d[i] = beta*d[i] + alpha*x[i];
+    }
+    return *this;
+  }
+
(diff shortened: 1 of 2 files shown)


src/ekat/ekat_pack.hpp
```diff
@@ -292,6 +292,58 @@ struct Pack {
     return *this;
   }
```
tcclevenger, line 294: I would add a short, general comment for update methods that has the `y = beta*y + alpha*x` pseudo code.

tcclevenger (approved):

PR #333, opened 2024-04-05
Add `get_subview()` Implementations for Multi-Sliced Subfields

mjs271:
## Motivation

This PR is associated with https://github.com/E3SM-Project/scream/pull/2785 and provides `get_subview_<N>()` implementations for getting a multi-sliced `subfield`'s underlying view. I've created a separate overload function for each view rank of `N = 1, ... 6` and employ if statements within to differentiate between which dimension is being sliced. Using the logic blocks isn't my favorite solution from an efficiency standpoint, so I'd be happy to hear any suggestions about a better approach.

## Testing

The added functions are tested within EAMxx's[ tests for `Field`-related functionality](https://github.com/E3SM-Project/scream/blob/1439884714a394c5227b535de4b4a98c8614443a/components/eamxx/src/share/tests/field_tests.cpp). As of submitting this PR, some more tests are required, but I wanted to prioritize getting this out with the draft-submission for the EAMxx PR.

### DIFF
--- a/src/ekat/kokkos/ekat_subview_utils.hpp
+++ b/src/ekat/kokkos/ekat_subview_utils.hpp
@@ -330,6 +330,7 @@ subview(const ViewLR<ST*, Props...>& v,
         const Kokkos::pair<int, int> &kp0,
         const int idim = 0) {
   assert(v.data() != nullptr);
+  assert(idim == 0);
   assert(kp0.first >= 0 && kp0.first < kp0.second);
   return Unmanaged<ViewLS<ST*,Props...>>(Kokkos::subview(v, kp0));
 }
--- a/src/ekat/kokkos/ekat_subview_utils.hpp
+++ b/src/ekat/kokkos/ekat_subview_utils.hpp
@@ -317,7 +317,10 @@ subview_1(const ViewLR<ST******,Props...>& v,
 // ================ Multi-sliced Subviews ======================= //
 // e.g., instead of a single-entry slice like v(:, 42, :), we slice over a range
 // of values, as in v(:, 27:42, :)
-// this means that the subview has the same rank as the source view
+// Note that this obtains entries for which in dimesion 2 is in the
+// range [27, 42) == {v(i, j, k), where 27 <= j < 42}
+// Note also that this slicing means that the subview has the same rank
+// as the source view
 
 // --- Rank1 multi-slice --- //
 template <typename ST, typename... Props>
@@ -325,15 +328,10 @@ KOKKOS_INLINE_FUNCTION
 Unmanaged<ViewLS<ST*, Props...>>
 subview(const ViewLR<ST*, Props...>& v,
         const Kokkos::pair<int, int> &kp0,
-        const int idim) {
+        const int idim = 0) {
   assert(v.data() != nullptr);
-  assert(idim == 0);
-  // NOTE: the final comparison is originally int <= long unsigned int
-  // the cast silences a warning, but may be unnecessary
-  assert(kp0.first >= 0 && kp0.first < kp0.second
-         && kp0.second <= v.extent_int(idim));
+  assert(kp0.first >= 0 && kp0.first < kp0.second);
   return Unmanaged<ViewLS<ST*,Props...>>(Kokkos::subview(v, kp0));
-
 }
 
 // --- Rank2 multi-slice --- //
@@ -345,10 +343,8 @@ subview(const ViewLR<ST**, Props...>& v,
         const int idim) {
   assert(v.data() != nullptr);
   assert(idim >= 0 && idim <= v.rank);
-  // NOTE: the final comparison is originally int <= long unsigned int
-  // the cast silences a warning, but may be unnecessary
   assert(kp0.first >= 0 && kp0.first < kp0.second
-         && kp0.second <= v.extent_int(idim));
+         && kp0.second < v.extent_int(idim));
   if (idim == 0) {
     return Unmanaged<ViewLS<ST**,Props...>>(Kokkos::subview(v, kp0, Kokkos::ALL));
   } else {
@@ -366,10 +362,8 @@ subview(const ViewLR<ST***, Props...>& v,
         const int idim) {
   assert(v.data() != nullptr);
   assert(idim >= 0 && idim <= v.rank);
-  // NOTE: the final comparison is originally int <= long unsigned int
-  // the cast silences a warning, but may be unnecessary
   assert(kp0.first >= 0 && kp0.first < kp0.second
-         && kp0.second <= v.extent_int(idim));
+         && kp0.second < v.extent_int(idim));
   if (idim == 0) {
     return Unmanaged<ViewLS<ST***,Props...>>(
       Kokkos::subview(v, kp0, Kokkos::ALL, Kokkos::ALL));
@@ -392,10 +386,8 @@ subview(const ViewLR<ST****, Props...>& v,
         const int idim) {
   assert(v.data() != nullptr);
   assert(idim >= 0 && idim <= v.rank);
-  // NOTE: the final comparison is originally int <= long unsigned int
-  // the cast silences a warning, but may be unnecessary
   assert(kp0.first >= 0 && kp0.first < kp0.second
-         && kp0.second <= v.extent_int(idim));
+         && kp0.second < v.extent_int(idim));
   if (idim == 0) {
     return Unmanaged<ViewLS<ST****,Props...>>(
       Kokkos::subview(v, kp0, Kokkos::ALL, Kokkos::ALL, Kokkos::ALL));
@@ -421,10 +413,8 @@ subview(const ViewLR<ST*****, Props...>& v,
         const int idim) {
   assert(v.data() != nullptr);
   assert(idim >= 0 && idim <= v.rank);
-  // NOTE: the final comparison is originally int <= long unsigned int
-  // the cast silences a warning, but may be unnecessary
   assert(kp0.first >= 0 && kp0.first < kp0.second
-         && kp0.second <= v.extent_int(idim));
+         && kp0.second < v.extent_int(idim));
   if (idim == 0) {
     return Unmanaged<ViewLS<ST*****,Props...>>(
       Kokkos::subview(v, kp0, Kokkos::ALL, Kokkos::ALL, Kokkos::ALL, Kokkos::ALL));
@@ -453,10 +443,8 @@ subview(const ViewLR<ST******, Props...>& v,
         const int idim) {
   assert(v.data() != nullptr);
   assert(idim >= 0 && idim <= v.rank);
-  // NOTE: the final comparison is originally int <= long unsigned int
-  // the cast silences a warning, but may be unnecessary
   assert(kp0.first >= 0 && kp0.first < kp0.second
-         && kp0.second <= v.extent_int(idim));
+         && kp0.second < v.extent_int(idim));
   if (idim == 0) {
     return Unmanaged<ViewLS<ST******,Props...>>(
       Kokkos::subview(v, kp0, Kokkos::ALL, Kokkos::ALL, Kokkos::ALL,
@@ -484,7 +472,6 @@ subview(const ViewLR<ST******, Props...>& v,
                       Kokkos::ALL, kp0));
   }
 }
-
 } // namespace ekat
 
 #endif // EKAT_SUBVIEW_UTILS_HPP
--- a/src/ekat/kokkos/ekat_subview_utils.hpp
+++ b/src/ekat/kokkos/ekat_subview_utils.hpp
@@ -93,18 +93,6 @@ subview(const ViewLR<ST****,Props...>& v,
       &v.impl_map().reference(i0, i1, i2, 0),v.extent(3));
 }
 
-// --- Rank4 multi-slice --- //
-template <typename ST, typename... Props>
-KOKKOS_INLINE_FUNCTION
-ViewLS<ST****, Props...>
-subview(const ViewLR<ST****, Props...>& v,
-        const Kokkos::pair<int, int> &kp0,
-        const int idim) {
-  assert(v.data() != nullptr);
-  auto sv = Kokkos::subview(v, kp0, Kokkos::ALL, Kokkos::ALL, Kokkos::ALL);
-  return Unmanaged<ViewLS<ST****,Props...>>(sv);
-}
-
 // --- Rank5 --- //
 template <typename ST, typename... Props>
 KOKKOS_INLINE_FUNCTION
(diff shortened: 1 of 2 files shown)


src/ekat/kokkos/ekat_subview_utils.hpp
```diff
@@ -316,6 +316,191 @@ subview_1(const ViewLR<ST******,Props...>& v,
...
+  assert(idim >= 0 && idim <= v.rank);
+  // NOTE: the final comparison is originally int <= long unsigned int
+  // the cast silences a warning, but may be unnecessary
+  assert(kp0.first >= 0 && kp0.first < kp0.second
+         && (unsigned int)kp0.second <= v.extent(idim));
+  if (idim == 0) {
+    return Unmanaged<ViewLS<ST**,Props...>>(Kokkos::subview(v, kp0, Kokkos::ALL));
+  } else if (idim == 1) {
+    return Unmanaged<ViewLS<ST**,Props...>>(Kokkos::subview(v, Kokkos::ALL, kp0));
+  } else {
```
jgfouca, line 362: I don't think we want this exact pattern since it will lead to warnings (code branch without a return statement). You've already asserted earlier that `idim >= 0 && idim <= v.rank`, so we should never hit the else. I would do it this way:
```
if (idim == 0) {
    return Unmanaged<ViewLS<ST**,Props...>>(Kokkos::subview(v, kp0, Kokkos::ALL));
} else {
    assert(idim == 1); // probably redundant but doesn't hurt anything
    return Unmanaged<ViewLS<ST**,Props...>>(Kokkos::subview(v, Kokkos::ALL, kp0));
}
```

jgfouca (changes requested):
Some minor comments. Also, please a quick unit test for these.

src/ekat/kokkos/ekat_subview_utils.hpp
```diff
@@ -267,7 +267,7 @@ subview_1(const ViewLR<ST****,Props...>& v,
   // Since we are keeping the first dimension, the stride is unchanged.
   auto vm = tmp.impl_map();
   vm.m_impl_offset.m_stride = v.impl_map().stride_0();
-  auto test =  Unmanaged<ViewLR<ST***,Props...>>(v.impl_track(),vm);
+  // auto test =  Unmanaged<ViewLR<ST***,Props...>>(v.impl_track(),vm);
```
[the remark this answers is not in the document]
mjs271: Ah, good catch. Was pretty sure it was extraneous--I'll nix it

src/ekat/kokkos/ekat_subview_utils.hpp
mjs271: I like that--thanks for the suggestion!

src/ekat/kokkos/ekat_subview_utils.hpp
```diff
@@ -316,6 +316,191 @@ subview_1(const ViewLR<ST******,Props...>& v,
...
+  assert(idim >= 0 && idim == v.rank);
+  // NOTE: the final comparison is originally int <= long unsigned int
+  // the cast silences a warning, but may be unnecessary
+  assert(kp0.first >= 0 && kp0.first < kp0.second
+         && (unsigned int)kp0.second <= v.extent(idim));
+  if (idim == 0) {
+    return Unmanaged<ViewLS<ST*,Props...>>(Kokkos::subview(v, kp0, Kokkos::ALL));
+  } else {
+    // FIXME: better way to do this? necessary?
+    assert(false);
```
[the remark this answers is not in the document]
mjs271: Good catch! Nixed it and confirmed with test

mjs271:
Added tests for `ekat::subview()` for the multi-sliced case. Note that these are _sort of_ redundant with the eamxx `f.subfield()` and `get_multi_sliced_view()` tests in `subfield_tests.cpp`, but it seemed best to add individual function tests to the respective code bases where they live.

With this, I will also convert https://github.com/E3SM-Project/scream/pull/2785 from draft to a full PR.

src/ekat/kokkos/ekat_subview_utils.hpp
```diff
@@ -316,6 +315,177 @@ subview_1(const ViewLR<ST******,Props...>& v,
...
+
+// --- Rank1 multi-slice --- //
+template <typename ST, typename... Props>
+KOKKOS_INLINE_FUNCTION
+Unmanaged<ViewLS<ST*, Props...>>
+subview(const ViewLR<ST*, Props...>& v,
+        const Kokkos::pair<int, int> &kp0,
+        const int idim) {
+  assert(v.data() != nullptr);
+  assert(idim == 0);
```
tcclevenger, line 331: I would remove `idim` argument for this case I think, since it is always 0. But ok if you or others disagree.

tests/kokkos/kokkos_utils_tests.cpp
```diff
@@ -544,4 +544,230 @@ TEST_CASE("subviews") {
...
+          }
+          // Make sure that our diffs counting strategy works
+          // by checking that two entries that should be different
+          // are indeed different.
+          if (v1_0(0) != v6(i0, i1, i2, i3, i4, 1))
+            ++ndiffs;
+        });
+    auto diffs_h = Kokkos::create_mirror_view(diffs);
+    Kokkos::deep_copy(diffs_h, diffs);
+    REQUIRE(diffs_h() == 1);
```
tcclevenger, line 770: Why are you expecting a diff?

src/ekat/kokkos/ekat_subview_utils.hpp
```diff
@@ -316,6 +315,177 @@ subview_1(const ViewLR<ST******,Props...>& v,
       v.impl_track(),vm);
 }
 
+// ================ Multi-sliced Subviews ======================= //
+// e.g., instead of a single-entry slice like v(:, 42, :), we slice over a range
+// of values, as in v(:, 27:42, :)
+// this means that the subview has the same rank as the "parent" view
```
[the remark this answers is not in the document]
mjs271: I meant to refer to the view from which the subview is being taken, but I agree it's best to avoid ambiguity. Will go with "source view" unless there's objections

src/ekat/kokkos/ekat_subview_utils.hpp
mjs271: I can't say I have strong opinions either way, other than a potential argument for consistent function signatures. Though I'm not sure that is much of a big deal

Any thought from others?

tests/kokkos/kokkos_utils_tests.cpp
mjs271: Haha, that one got me the first time, too.

This came from the original subview tests, and above, at the end of the `parallel_for()`, there's a sanity check to make sure the diffing workflow catches a known diff. So, the count ends up as identically 1 in the passing case

src/ekat/kokkos/ekat_subview_utils.hpp
bartgol: I would remove it. One rule I like when coding is "do not expose parameters/args unless the user is allowed to change/set them". If this parameter cannot be set by the user, then it should not be part of the function signature.

src/ekat/kokkos/ekat_subview_utils.hpp
```diff
@@ -316,6 +314,177 @@ subview_1(const ViewLR<ST******,Props...>& v,
...
+template <typename ST, typename... Props>
+KOKKOS_INLINE_FUNCTION
+Unmanaged<ViewLS<ST*, Props...>>
+subview(const ViewLR<ST*, Props...>& v,
+        const Kokkos::pair<int, int> &kp0,
+        const int idim) {
+  assert(v.data() != nullptr);
+  assert(idim == 0);
+  // NOTE: the final comparison is originally int <= long unsigned int
+  // the cast silences a warning, but may be unnecessary
```
bartgol, line 332: I think this comment is outdated. Perhaps we were using `extent` instead of `extent_int` at some point. But as of now, this comment is unnecessary. Let's remove it.

Edit: same goes for the same comment on the other overloads

src/ekat/kokkos/ekat_subview_utils.hpp
```diff
@@ -316,6 +314,177 @@ subview_1(const ViewLR<ST******,Props...>& v,
...
+
+// --- Rank2 multi-slice --- //
+template <typename ST, typename... Props>
+KOKKOS_INLINE_FUNCTION
+Unmanaged<ViewLS<ST**, Props...>>
+subview(const ViewLR<ST**, Props...>& v,
+        const Kokkos::pair<int, int> &kp0,
+        const int idim) {
+  assert(v.data() != nullptr);
+  assert(idim >= 0 && idim <= v.rank);
```
bartgol, line 347: I think the second check should have < rather than <=

Edit: same in the other overloads

src/ekat/kokkos/ekat_subview_utils.hpp
```diff
@@ -316,6 +314,177 @@ subview_1(const ViewLR<ST******,Props...>& v,
       v.impl_track(),vm);
 }
 
+// ================ Multi-sliced Subviews ======================= //
+// e.g., instead of a single-entry slice like v(:, 42, :), we slice over a range
+// of values, as in v(:, 27:42, :)
+// this means that the subview has the same rank as the source view
```
bartgol, line 320: I would add a comment starting that the pair of indices is to be interpreted as `[beg,end)`, so the second index is the first _after_ the desired ones (just like the end iterator in std containers).

src/ekat/kokkos/ekat_subview_utils.hpp
```diff
@@ -316,6 +314,177 @@ subview_1(const ViewLR<ST******,Props...>& v,
...
+subview(const ViewLR<ST**, Props...>& v,
+        const Kokkos::pair<int, int> &kp0,
+        const int idim) {
+  assert(v.data() != nullptr);
+  assert(idim >= 0 && idim <= v.rank);
+  // NOTE: the final comparison is originally int <= long unsigned int
+  // the cast silences a warning, but may be unnecessary
+  assert(kp0.first >= 0 && kp0.first < kp0.second
+         && kp0.second <= v.extent_int(idim));
+  if (idim == 0) {
```
bartgol, line 352: If you don't like these conditionals, we could bake the index in the function name. Someone like `subview_0` for subviewing along the host index vs `subview_1` for sunburn l subviewing asking the second. I _think_ in most (all?) use cases we will know which dimension we're subviewing anyways. @tcclevenger what do you think?

tests/kokkos/kokkos_utils_tests.cpp
```diff
@@ -544,4 +544,230 @@ TEST_CASE("subviews") {
   }
 }
 
+TEST_CASE("multi-slice subviews") {
```
bartgol, line 547: You should add some tests that catch errors. Since we used `assert` instead of exceptions, we can't use `REQUIRE_THROWS`. 

Perhaps we could write a separate test with the WILL_FAIL option, and run one bad op in there. We can't test more since the assert crashes at the first error.. but maybe we can put N bad calls and randomly choose one to execute. Over time, we'd have some sort of coverage. Of course, this test can only be added in debug mode, since asserts are not executed in release.

@jgfouca @tcclevenger do you have other ideas on how to test that asserts do indeed work?

bartgol (commented):
Looks good. I have some comments/suggestions.

tests/kokkos/kokkos_utils_tests.cpp
jgfouca: @bartgol , I recommend using the EKAT assert macros in ekat_assert.hpp. 
```
EKAT_ASSERT(false); // throws a logic_error
```

That would allow us to use REQUIRE_THROWS. This won't work for `EKAT_KERNEL_ASSERT` since that's just a Kokkos::abort.

src/ekat/kokkos/ekat_subview_utils.hpp
tcclevenger: My only issue is this would end up creating a lot of function, right? `subview_1()` would need 6 versions, `subview_2()` would have 5, etc. And then in the SCREAM-side PR we would end up having a long `switch(idim) case 1:, case 2:, ...` chain.

src/ekat/kokkos/ekat_subview_utils.hpp
mjs271: Good catch--thank you!

src/ekat/kokkos/ekat_subview_utils.hpp
mjs271: Unless there's an advantage to changing, I suppose I'm in favor of leaving it as-is, but could certainly be convinced to go with the opposite case

src/ekat/kokkos/ekat_subview_utils.hpp
mjs271: So... there's probably something I'm doing wrong, but getting rid of the `idim` argument caused issues with resolving the choice of overloaded function. I'm guessing it's due to handling it within `field_impl.hpp::get_multi_slice_view()` using an if block, i.e.,
```
if (idim == 0 && v_fullsize.rank == 1) {
  return DstView(ekat::subview_m1(v_fullsize,
                              Kokkos::make_pair<int, int>(k_beg, k_end)));
} else {
  return DstView(ekat::subview(
      v_fullsize, Kokkos::make_pair<int, int>(k_beg, k_end), idim));
}
```
In light of that, I'm happy to accept a suggestion for a fix that eliminates the extraneous argument, but in the mean time added a default value for the `idim` argument, and this makes the overloading straightforward again.

tcclevenger: Yeah, I see now that the way you had it allowed SCREAM side to not have any if-else. 

What is the `ekat::subview_m1()`, is that a typo? If you have an if-else on the SCREAM side, there should be no reason to have the idim input. What was the error you were getting?

If we keep the idim input, I think there needs to still be an assert that the user only uses idim=0, since they could use it thinking it does something else.

src/ekat/kokkos/ekat_subview_utils.hpp
tcclevenger: Unless @bartgol  has push back I'd say keep this how it is, and then later we may rethink all the `subview` routines.

src/ekat/kokkos/ekat_subview_utils.hpp
mjs271: Ah, it is a type... kind of. That was my attempt to differentiate the two choices by naming them differently. However, my interpretation is that the if/then results in a compile-time ambiguity about which function will be chosen by different sized views, and so doesn't build.

Also, if I'm recalling correctly, when it was just named `ekat::subview(const View<DT*> v, const kk::make_pair<int, int>(a, b))`, the compiler was considering the single-slice 1D function, `ekat::subview(const View<DT*> v, const int idx)`,  as a candidate, which some stranger on the internet seemed to be saying is because the compiler can't decide whether to prioritize choosing a 2-argument overloaded function based on type vs. const-ness 🤷‍♂️

All that being said... I will add that `assert()` back in 🙂

PR #334, opened 2024-04-15
Add cmake utility to create tests for asserts

bartgol:
## Motivation

The Catch2 framework does not offer a way to test that an assertion fails. It has `REQUIRE_THROWS`, but that only works with _exception_, not _assertions_. I found online a library that can test assertions via `setjmp` and the likes, but I'm not sure this would work correctly for device code. So I decided to roll our own testing mechanism. The idea is simple: wrap every "bad" call in `EKAT_TEST_ASSERT(...)`. This macro is never defined, but the cmake+python utility will use this special string to generate a test. In particular:
- we generate a separate unit test for each `EKAT_TEST_ASSERT (...)` line
- while testing the n-th occurrence of the `EKAT_TEST_ASSERT` string, remove all lines containing that string, except for the current one, where it will simply remove `EKAT_TEST_ASSERT`.
- since `EKAT_TEST_ASSERT` is simply removed from the line to test, it doesn't really matter where it appears in the line; it can even be a comment at the end:
```
   bad_call (); // EKAT_TEST_ASSERT
```
- the bad function call must fit in a single line, together with EKAT_TEST_ASSERT. Implementing a smarter multi-line parser (which would require a "macro-like" implementation, rather than the "comment" version in the point above) is harder, and I didn't want to spend more time on this.
- the utility can only accept a _single_ source cpp file as input
- for each line to test, we generate a separate src file in the build folder, and then we invoke the usual `EkatCreateUnitTest` with that file.
- the utility has no knowledge about NDEBUG. It's up to the user to use `EkatCreateAssertTest` only in debug builds.

## Testing

I added some dummy tests that verify this mechanism works.

## Additional Information
If we decide to merge this, I will later add some usage to existing tests, to ensure our existing `assert` calls work as expected.


tcclevenger (commented):
This looks good to me, but I'll give @jgfouca a chance to look since he's more familiar with the cmake.

cmake/EkatCreateUnitTestWithAsserts.cmake
```diff
@@ -0,0 +1,77 @@
+include(EkatCreateUnitTest)
+
+# This function creates unit tests to verify that calls to assert()
+# perform as expected. This function is not aware of whether asserts
+# are compiled away or not (i.e., whether the macro NDEBUG is present
```
jgfouca, line 5: CMake should know if the build is debug or not. We can use `EKAT_IS_DEBUG_BUILD` or `CMAKE_BUILD_TYPE`. We could maybe just have this function do nothing if it's not a debug build, that way the users of this function won't have to remember to guard the call.

jgfouca (changes requested):
One minor thought, otherwise looks good.

jgfouca (approved):

tcclevenger (approved):

PR #336, opened 2024-04-26
Another modification for EAMxx Multi-sliced Subfields

mjs271:
This is another small one. Just adds a couple of non-functional overloads to `get_subview()`-type functions that helps the compiler not crash. I'm not sure whether the `assert(false)` calls are the proper idea, so take a look and let me know--thanks!


bartgol:
I'm not super fond of this approach, to be honest. It makes no sense to call `subview_1` for a rank-1 view, so maybe we should try to add some `if constexpr` or `enable_if` magic at the call site, so that it doesn't call this function for rank-1 views. Same for `subview` of a rank-0 view.

mjs271:
Closing this because I found a better way to do this on the eamxx side of things.

PR #337, opened 2024-05-09
Refactor how Units class handles the string representation

bartgol:
## Motivation

In EAMxx, we use units such as `kg/kg`, to represent "mixing ratios". Since the `Units` class would interpret this as a non-dimensional unit, we allowed to set a string representation:
```c++
Units mix_ratio = kg/kg;
mix_ratio.set_string("kg/kg");
```
Over time, this has become a bit annoying. This PR changes the handling of the string representation altogether. Here's a summary of the mods:

- changed `to_string` from a free function to a method, for `Units`, `ScalingFactor`, and `RationalConstant`
- added `get_si_string` to get the representation of the `Units` obj in terms of fundamental SI units.
- the overload of `*`,`/`,`^` ops for `Units` will simply "compount" the string representation of the two terms.

For instance:
```c++
Units blah = m*kg/kg;
std::cout << blah.to_string(); // prints "m*(kg/kg)"
std::cout << blah.get_si_string(); // prints "m"
```
The user can still set funky string representation at construction time:
```c++
Units blah (kg/kg*kg/kg,"?!?");
std::cout << (m*blah/s).to_string(); // prints "(m*?!?)/s"
```

I added some testing for the string representatin

## Additional Information
This PR breaks current code, since the free functions `to_string(X)`, for X=`Units`, `ScalingFactor`, and `RationalConstant` has been removed in favor of a class method.

### DIFF
--- a/src/ekat/util/ekat_rational_constant.hpp
+++ b/src/ekat/util/ekat_rational_constant.hpp
@@ -3,6 +3,7 @@
 
 #include <stdexcept>
 #include <sstream>
+#include <array>
 
 #include "ekat/ekat_assert.hpp"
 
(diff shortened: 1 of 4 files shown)


bartgol:
@tcclevenger I forgot that the AT requires approvals from users with write access, so I'm pinging you as well.

tcclevenger (approved):

PR #340, opened 2024-05-21
Fix access to exec space concurrency

bartgol:
## Motivation

Kokkos allows to enable/disable code that was deprecated in version 4. Since we should push for _not_ enabling it, we need to remove all usage of deprecated code.

## Testing

Verified this compiles with `-D Kokkos_ENABLE_DEPRECATED_CODE_4:BOOL=OFF`


bartgol:
@jgfouca Once this makes it into EAMxx, we should start adding `Kokkos_ENABLE_DEPRECATED_CODE_4:BOOL=OFF` to all our standalone testing machine files (eventually also in E3SM, but that may take a bit longer.

jgfouca (approved):

PR #343, opened 2024-06-21
GetMpiDistributionName: Fix for cray MPI

jgfouca:
If the cray-specific one works, that should take precedence over the generic one if they both succeed.

## Motivation

The case SMS.ne30pg2_ne30pg2.F2010-SCREAMv1.pm-gpu_nvidiagpu was failing due to GetMpiDistributionName failing to probe the distribution. The reason is that it was trying to use the `-show` output instead of the `--cray-print-opts=cflags`. Both commands succeeded but the -show one took precedence due to being checked first. I think this is wrong and the cray-specific one should take highest precedence if it works.

## Testing

SMS.ne30pg2_ne30pg2.F2010-SCREAMv1.pm-gpu_nvidiagpu make it past cmake configure (with a macro fix).


ndkeen:
Yes this might be all that's needed -- but we could also try to detect if we are in cray-mpich world or not, then do the best one ?  The issue is the output from `-show` (at least with nvidia and mpich on pm) is long and doesnt contain the strings we are looking for.

ndkeen (approved):

bartgol (approved):

PR #344, opened 2024-06-25
Make copy constructor/assignment of ParameterList do a deep_copy

bartgol:
## Motivation

Since parameters are stored via ekat::any, and ekat::any uses shared_ptr internally, a default copy constructor/assignment would simply copy the shared pointers, which means that the new list will share data with the old list for any existing parameter (new parameter added after the copy will be local to each list though).

This PR makes the copy a deep one, so that changing the value of parameters in the original list will not be reflected in the new list (and vice versa). To allow sharing pointers, I added a `soft_copy` method, in case sharing data is desired. Notice though that a soft copy will not share the stored maps (string->params and string->sublist), but simply copy them (which, as explained above, means existing params will be shared).

## E3SM Stakeholder Feedback

I kind of need this to allow easier usage of pyeamxx in SCREAM repo

## Testing

I added a test to verify that the copy constructor/assignment does not cause to share data with the original parameter list.

### DIFF
--- a/src/ekat/ekat_parameter_list.cpp
+++ b/src/ekat/ekat_parameter_list.cpp
@@ -4,6 +4,45 @@
 
 namespace ekat {
 
+ParameterList::
+ParameterList (const ParameterList& src)
+{
+  m_name = src.m_name;
+  for (const auto& it : src.m_params) {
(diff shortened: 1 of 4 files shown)


jgfouca (approved):
LGTM

PR #346, opened 2024-07-25
Fix RepackType when packing const scalar data

bartgol:
## Motivation

Calling `repack` passing a view of const scalar data type (not a `const Pack<T,N>`) in LinInterp::setup was causing an error, since the output of `repack` was a `View<const Pack<const T,N>,...>`, which is not compatible with what `setup_impl` expected, which is `View<const Pack<T,N>>`. The fix is very quick, though, and I manually verified that I can now call LinInterp::setup passing views of const doubles.

Note: everything was already working fine passing views of const packs, since the meta utility `RepackType` was already handling const correctly for pack types.

### DIFF
--- a/src/ekat/ekat_pack_kokkos.hpp
+++ b/src/ekat/ekat_pack_kokkos.hpp
@@ -233,7 +233,7 @@ template<int N, typename OldType>
 struct RepackType {
   using type =
     typename std::conditional<std::is_const<OldType>::value,
-                              const Pack<OldType,N>,
+                              const Pack<std::remove_const_t<OldType>,N>,
                               Pack<OldType,N>>::type;
 };
 
--- a/tests/pack/pack_kokkos_tests.cpp
+++ b/tests/pack/pack_kokkos_tests.cpp
@@ -218,47 +218,59 @@ TEST_CASE("repack", "ekat::pack") {
   using ekat::Pack;
   using ekat::repack;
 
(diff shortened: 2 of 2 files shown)


tcclevenger (approved):

tcclevenger (changes requested):
Before this merges, I'll revoke my review just to ask if it would be simple to add this case to the Linear interp tests?

Edit: maybe you already addressed this in your last paragraph..

tcclevenger (approved):
I'll reaprove and take off automerge. You can add it back if a test is unnecessary.

bartgol:
> Before this merges, I'll revoke my review just to ask if it would be simple to add this case to the Linear interp tests?
> 
> Edit: maybe you already addressed this in your last paragraph..

The repack function cannot work for, say, `View<const double*>` unless we add this fix. We virtually never use `Pack<const T,N>`, but prefer to use `const Pack<T,N>` everywhere, so it makes sense to me to ensure that RepackType<...>::type _never_ returns a `Pack<const T, N>`.

bartgol:
@tcclevenger sorry, I think I misread your comment. I thought you wanted to move the fix into the LinInterp class, which is impossible. Now I understand you meant to add a test for repacking view of const data. I can add that in the repack unit tests.

tcclevenger (approved):

PR #348, opened 2024-08-16
Avoid running compute-sanitizer in very large kernel

tcclevenger:
## Motivation
Trying to shorten the time taken by compute sanitizer nightlies for EAMxx. The WSM unit test contains a kernel over a large range that would take the racecheck tool ~25 min to complete. I don't think there is any value in running that kernel through the sanitizer since the other kernels in that unit test will run through all the WSM src code, and take on the order of a few seconds. This is the only ekat test that takes over a few seconds.


tcclevenger:
We have a SCREAM-based solution. This is not an issue for EKAT since the AT does not contain tests with compute-sanitizer.

PR #349, opened 2024-08-23
fix unused warnings in release builds

mahf708:
## Motivation

reduce warnings in release builds

## E3SM Stakeholder Feedback

n/a

## Testing

n/a

---

This was previously supposed to have been fixed but it was present on pm-gpu. See https://github.com/E3SM-Project/EKAT/pull/345/commits/d1f98cfba86b5a1e93315033b344af192af885cd. The sz_out is the one I had in my standalone shoc build.


bartgol (approved):
The ~lord~ compiler works in mysterious ways...

PR #354, opened 2024-11-27
Improve EKAT_REQUIRE macro to allow passing exception type

bartgol:
## Motivation
When doing unit tests, I often check that exceptions are raised when appropriate, via something like
```c++
  REQUIRE_THROWS (some_call_which_throws());
```
But if the function can throw at several points in its execution, this check may not verify the desired behavior. E.g., consdider this
```c++
void func (int i, int j) {
  EKAT_REQUIRE_MSG (i>0, "wrong i");
  EKAT_REQUIRE_MSG (j>0, "wrong j");
}
TEST_CASE ("my-test")
{
  int i=2, j=3;
  func(i,j); // This should work
  
  i = -1;
  REQUIRE_THROWS (func(i,j)); // should fail b/c i<=0
  j = -1;
  REQUIRE_THROWS (func(i,j)); // should fail b/c j<=0
}
```
The second `REQUIRE_THROWS` _will_ throw, but NOT b/c `j<0`, bur rather b/c `i<0`, since we did not change its value back to 2.

With this PR, I can do
```c++
class ExceptionA : public std::runtime_error {};
class ExceptionB : public std::logic_error {};
void func (int i, int j) {
  EKAT_REQUIRE_MSG (i>0, "wrong i",ExceptionA);
  EKAT_REQUIRE_MSG (j>0, "wrong j",ExceptionB);
}
TEST_CASE ("my-test")
{
  int i=2, j=3;
  func(i,j); // This should work
  
  i = -1;
  REQUIRE_THROWS_AS (func(i,j),ExceptionA); // should fail b/c i<=0
  j = -1;
  REQUIRE_THROWS_AS (func(i,j),ExceptionB); // should fail b/c j<=0
}
```
Now, the test will fail, since the 2nd `REQUIRE_THROWS` will still throw `ExceptionA`.

Note: this change is backward compatible.

## Testing
I added a unit test that verifies we can now throw a user-defined exception type.

### DIFF
--- a/src/ekat/ekat_assert.hpp
+++ b/src/ekat/ekat_assert.hpp
@@ -2,6 +2,7 @@
 #define EKAT_ASSERT_HPP
 
 #include <sstream>
+#include <iostream>
 #include <exception>
 #include <assert.h>
 #include <stdexcept>  // For std::logic_error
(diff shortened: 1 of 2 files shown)


src/ekat/ekat_assert.hpp
```diff
@@ -25,16 +26,24 @@
...
+  do {                                                                            \
+    if ( ! (condition) ) {                                                        \
+      std::stringstream _ss_;                                                     \
+      _ss_ << "\n FAIL:\n" << #condition  << "\n";                                \
+      _ss_ << EKAT_BACKTRACE;                                                     \
+      _ss_ << "\n" << msg;                                                        \
+      if constexpr (std::is_constructible<exception_type, std::string>::value) {  \
+        throw exception_type(_ss_.str());                                         \
+      } else {                                                                    \
+        std::cerr << _ss_;                                                        \
```
bartgol, line 43: This prints the message to screen before raising the exception.

This case allows to handle most (if not all) std library exception types. User defined exceptions MUST fit one of these two constructors (empty or taking a `std::string` or `const char*`) in order for code to compile.

jgfouca:
@bartgol , I'm confused by your example:

```
  int i=2, j=3;
  func(i,j); // This should work
  
  i = -1;
  REQUIRE_THROWS_AS (func(i,j),ExceptionA); // should fail b/c i<=0
  j = -1;
  REQUIRE_THROWS_AS (func(i,j),ExceptionB); // should fail b/c j<=0
```

It looks to me that i must get set back to 2 for the second check to work. How would the program ever make it to the line
`EKAT_REQUIRE_MSG (j>0, "wrong j",ExceptionB);` if the line before it is always throwing an exception?

In any case, the PR itself looks fine.

jgfouca (dismissed):

jgfouca:
I see now what you're getting at. The unit test itself had mistakes.

bartgol:
> I see now what you're getting at. The unit test itself had mistakes.

Yeah, the macro was fine. But always throwing the same exception does not allow us to differentiate in case we need to throw multiple exceptions in the same call. Besides, it is helpful to be able to throw a specific type of exception, in case the calling code is able to handle SOME type of exceptions but nothers (e.g., a bad cast can be handled, but a bad alloc can't)

bartgol:
Tests pass on mappy.

jgfouca (approved):

PR #355, opened 2024-12-10
Upgrades to yaml writer

bartgol:
## Motivation

There are mainly two upgrades:
- allow long multiline strings values: the yaml writer automatically chooses a multiline format if the string contains the '\n' char
- allow inline sequences: the yaml writer automatically chooses inline format if the overall sequence content won't be larger than 100 characters

For clarity, the following are, respectively, inline vs block formats for sequences:
```yaml
list_1: [one, two, three]
list_2:
  - one
  - two
  -three
```
If the overall line length is "short", inline lists make the file much easier to parse visually.

## Testing

Manually verified the behavior is as expected, by using this branch in EAMxx

### DIFF
--- a/src/ekat/io/ekat_yaml.cpp
+++ b/src/ekat/io/ekat_yaml.cpp
@@ -285,6 +285,7 @@ std::string write_param (const T& t) {
   s << std::boolalpha << t;
   return s.str();
 }
+
 template<>
 std::string write_param<double> (const double& t) {
   std::stringstream ss;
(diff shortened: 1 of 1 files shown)


src/ekat/io/ekat_yaml.cpp
```diff
@@ -303,47 +304,88 @@ std::string write_param<char> (const char& t) {
...
-      using vtype = decltype(t);
-      if (params.isType<vtype>(pname)) {
-        auto v = params.get<vtype>(pname);
-        out << write_param(v) << '\n';
-        found = true;
-      }
-      return found;
-    });
-    return found;
+    if (params.isType<int>(pname)) {
```
bartgol, line 310: The above syntax with `TypeListFor` seemed overly complicated, especially considering we know a priori the types, and there are only 4 of them...

src/ekat/io/ekat_yaml.cpp
```diff
@@ -303,47 +304,88 @@ std::string write_param<char> (const char& t) {
   return (t ? "true" : "false");
 }
 
-void write_parameter_list (const ParameterList& params, std::ostream& out, int indent) {
```
bartgol, line 306: Using a `YAML::Emitter`, we don't have to keep track of indentation and other shenanigans. It's the right thing to do: let the library take care of the YAML formatting.

jgfouca (dismissed):
Looks great!

src/ekat/io/ekat_yaml.cpp
```diff
@@ -303,47 +304,88 @@ std::string write_param<char> (const char& t) {
...
+      auto v = params.get<bool>(pname);
+      out << YAML::Key << pname << YAML::Value << v;
+    } else if (params.isType<double>(pname)) {
+      auto v = params.get<double>(pname);
+      out << YAML::Key << pname << YAML::Value << v;
+    } else if (params.isType<std::string>(pname)) {
+      auto v = params.get<std::string>(pname);
+      out << YAML::Key << pname << YAML::Value;
+      if (v.find('\n') != std::string::npos)
+        out << YAML::Literal; // Use Literal for multiline
```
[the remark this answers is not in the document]
bartgol: It's ironic to me that not using "Literal" makes the emitter treat \n literally (as a sequence of chars), while using "Literal" makes it interpret its meaning...

bartgol:
I manually ran the jenkins job on mappy, and the tests pass. @jgfouca when you've had a chance to review the last couple of commits, can you please merge?

PR #356, opened 2024-12-12
Fix variadic EKAT_REQUIRE macro

bartgol:
## Motivation

The implementation added in PR #354 was wrong, and it wasn't detected until I checked out that branch in EAMxx. Namely, the error was due to how `__VA_ARGS__` expands. Instead of passing it to `IMPL_THROW`, we should JUST use it to detect which macro to call. The new impl does just that, with three shorthand versions of EKAT_REQUIRE.

Also, I changed the default exception type to runtime_error, since it's more general than logic_error.

## Testing

I added a call to the three versions of EKAT_REQUIRE in a test, to ensure they all compile correctly.

### DIFF
--- a/src/ekat/ekat_assert.hpp
+++ b/src/ekat/ekat_assert.hpp
@@ -5,7 +5,7 @@
 #include <iostream>
 #include <exception>
 #include <assert.h>
-#include <stdexcept>  // For std::logic_error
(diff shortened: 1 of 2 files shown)


jgfouca (approved):

bartgol:
I fired up a jenkins build on mappy, and it passed. Merging.

PR #358, opened 2025-01-13
Adjacent difference utility

bartgol:
## Motivation

The function `index_and_shift` is often used simply to compute something like `v(i+1)-v(i)`. This can make the code hide a bit the meaning.

This PR adds the utility `adj_diff`, which, given a view (of either packs or scalars), computes ONE adjacent difference, at a given index. If the view is of scalars, it will return a single scalar (the difference at that index), while if the view is of packs it will return a pack of differences (starting at that index). There are two modes: forward and backward. In the scalar case, forward returns `v(index+1)-v(index)`, while backward returns `v(index)-v(index-1)`. In the packed case, `v(index+1)` and `v(index-1)` are replaced with the proper shift (left/right) of `v(index)`, padded with the correct value at the right/left of the pack.

The handling of the boundaries is rather simple, and assumes that the user will correctly discard the meaningless returned values. For instance, assuming a scalar view:
- Forward: for index=v.size()-1 (the last entry of the view) there is no "index+1" entry, so we use 0, which makes the fcn return simply `-v(last)`
- Backward: for index=0 (the first entry of the view) there is no "index-1" entry, so we use 0, which makes the fcn return simply `v(0)`

## E3SM Stakeholder Feedback

It could simplify some code in EAMxx

## Testing

Added a unit test for both scalar/packed views, also testing both the case where the view length is a multiple of the pack size or not.

## Additional Information
I thought about making this utility act on all the entries of the view. While this may simplify some aspect of it (e.g., index refers to the _SCALAR_ entry in the scalar version, and to the _PACKED_ entry in the pack version), it would also add more complexity at the call site, having to make arrangements for a temp view. Not to mention the fact that the user would have to pass a range struct or a team obj, so that the function can dispatch the proper parallel_for inside. All in all, I think the "single-entry difference" version ("single" meaning a single scalar or a single pack) is the most flexible.

### DIFF
--- a/tests/pack/pack_kokkos_tests.cpp
+++ b/tests/pack/pack_kokkos_tests.cpp
@@ -13,6 +13,13 @@ namespace {
 template <typename View, int rank, typename T = void>
 using OnlyRank = typename std::enable_if<View::rank == rank, T>::type;
 
+template<typename ViewT>
+typename ViewT::HostMirror cmvc(const ViewT& v) {
+  auto vh = Kokkos::create_mirror_view(v);
+  Kokkos::deep_copy(vh,v);
+  return vh;
+}
+
 template <typename View>
 void fill(const View& a)
 {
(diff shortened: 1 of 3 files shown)


tests/pack/pack_kokkos_tests.cpp
```diff
@@ -13,6 +13,13 @@ namespace {
 template <typename View, int rank, typename T = void>
 using OnlyRank = typename std::enable_if<View::rank == rank, T>::type;
 
+template<typename ViewT>
+typename ViewT::HostMirror cmvc(const ViewT& v) {
+  auto vh = Kokkos::create_mirror_view(v);
+  Kokkos::deep_copy(vh,v);
+  return vh;
```
tcclevenger, line 20: Why not use `Kokkos::create_mirror_view_and_copy()`?

tcclevenger: I guess it may be more characters because you need to define the space

tcclevenger (approved):

bartgol:
The new tests pass on weaver and my laptop. Merging.

PR #361, opened 2025-01-30
Allow to use where_expression inside min/max

bartgol:
## Motivation

In view of using `where_expression` in EAMxx, we need to support min/max operation inside reduction loops. This PR allows to do `max(w,s)` where `w` is a `where_expression` and `s` is the scalar type associated to the the value type of the expression. Namely, for `where_expression<MaskT,ValueT>`, `s` is of type `ekat::ScalarTraits<ValueT>::scalar_type`. If `ValueT` is a built-in scalar, that's the same as `ValueT`, but for pack types, it returns the type of the pack entries. With this modification, we can do something like
```c++
using value_t = decltype(my_view(0));
using scalar_t = ekat::ScalarTraits<value_t>::scalar_type;
Kokkos::parallel_reduce (tvr, KOKKOS_LAMBDA(int k, scalar_t& lmax)
{
   w = where(my_view(k) > 1e-10);
   lmax = max(w,lmax);
},Kokkos::Max<scalar_t>(output));
```
This will work regardless of whether the input view stores packs or built-in scalars, allowing a) code that is more readable, and b) the possibility to change scalar type (e.g. pack vs double) without having to change the code.

## Testing

Added a test in the "where" unit tests.

## Additional Information
It may be interesting to implement overloads such as
```c++
template<typename M, typename V>
typename where_expression<M,V>::scalar_t
operator+= (typename where_expression<M,V>::scalar_t& lhs, const where_expression<M,V>& rhs);
```
Notice that this is different from doing `where += scalar`, in which the user almost certainly meant "add scalar to all active entries of the where expression". In `scalar += where`, there aren't really many ways to interpret the output. And it is probably useful in case one wants to do a masked sum reduction:
```c++
using value_t = decltype(my_view(0));
using scalar_t = ekat::ScalarTraits<value_t>::scalar_type;
scalar_t sum;
Kokkos::parallel_reduce (tvr, KOKKOS_LAMBDA(int k, scalar_t& lsum)
{
   lsum += where(my_view(k)>1e-10);
},sum);
```

@tcclevenger I think I may add that to this PR, since it's still a reduction of a where_expression with a scalar starting point, just a different reduction type (`+=` (or other op) instead of `max`/`min`). But you may start taking a look if you want.

### DIFF
--- a/src/ekat/util/ekat_where.hpp
+++ b/src/ekat/util/ekat_where.hpp
@@ -17,6 +17,7 @@ class where_expression<bool,V>
 public:
   using mask_t = bool;
   using value_t = V;
+  using scalar_t = typename ekat::ScalarTraits<value_t>::scalar_type;
 
   where_expression (const bool& m, V& v)
    : m_mask (m) , m_value (v)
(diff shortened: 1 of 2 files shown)


bartgol:
@tcclevenger adding support for `scalar op= where_expression` reduction was quick, so I went ahead and added it. I only test += and *=, but since the impl is via a macro for all op, I figured that's enough...

tcclevenger (approved):
Looks good.

bartgol:
Tests pass on mappy. Merging.

PR #362, opened 2025-02-04
Changes for Kokkos 4.5 update and Aurora support

oksanaguba:


ambrad:
Do we all think this PR is ready to be merged into master? I do, unless something new comes up.

bartgol:
> Do we all think this PR is ready to be merged into master? I do, unless something new comes up.

We did fix all the deprecated code usage, right? I need to still run the CI CUDA build manually, to make sure there's nothing in EKAT that needs to be broken up (e.g., maybe a file to break in two). I can do that now.

ambrad:
> did fix all the deprecated code usage, right?

I fixed the atomic call. I don't think there was anything else, but I didn't think to examine the warnings closely. Aren't there CI builds on E3SM that would have failed if any remained?

bartgol:
> Aren't there CI builds on E3SM that would have failed in the build phase if any remained?

Yes. But the CUDA build times out, and shows no log, so it's hard to tell what happened there. It could indeed be a build issue...

ambrad:
Oh, is that the only build that disallows deprecated usage?

bartgol:
> Oh, is that the only build that disallows deprecated usage?

No, they all do. But I wonder if the debug symbols trigger something like in Homme (the sp build does not build homme), like a functor too big, and a test hangs...

bartgol:
After some digging on the E3SM CUDA debug build in the CI container, I decided we should be good to merge the ekat PR. The issue is the very large code that ptxas ends up building for SHOC with the single monolithic shoc_main function. We can handle this on the E3SM side, by doing some small mods to our CI build setting, to use small kernels for the CUDA debug build.

@ambrad @oksanaguba I am comfortable merging this PR. I'll do one more quick review, then we can merge (maybe Monday morning/).

cmake/machine-files/kokkos/intel-gen.cmake
```diff
@@ -0,0 +1,9 @@
+include (${CMAKE_CURRENT_LIST_DIR}/generic.cmake)
+
+option(Kokkos_ARCH_INTEL_GEN "" ON)
+set(Kokkos_ENABLE_DEPRECATED_CODE_3 TRUE CACHE BOOL "")
```
bartgol, line 4: This variable is unused. Code that was deprecated in Kokkos 3.X has been removed by now.

bartgol (approved):
Ok, this looks good. There's one pointless cmake var we could remove, since it should be unused by kokkos, but not a big deal.

PR #367, opened 2025-03-29
Deprecate std_meta utility classes

bartgol:
## Motivation

The `ekat::any` and `ekat::enable_shared_from_this` classes were introduced out of necessity when we were still requiring only C++14. The former did not exist at all in C++14, while the latter was missing the important member `weak_from_this`. A while ago we have started requiring C++17 (namely, when Kokkos started mandating it), where these two classes are present, and offer all the functionalities we need.

At this point, there really isn't a reason for these classes to still be shipped by ekat. We are deprecating them (rather than removing them), to give people time to transition away. We should plan an official ekat release (maybe with some documentation) in the next few months. I don't think we have a release or tag anywhere, so it's time to do that anyways.

### DIFF
--- a/src/ekat/ekat_parameter_list.cpp
+++ b/src/ekat/ekat_parameter_list.cpp
@@ -33,17 +33,6 @@ ParameterList& ParameterList::operator= (const ParameterList& src)
   return *this;
 }
 
-ParameterList ParameterList::soft_copy () const
-{
-  ParameterList p;
-
-  p.m_name = m_name;
-  p.m_params = m_params;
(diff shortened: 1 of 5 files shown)


tcclevenger (dismissed):
Feel free to assign some doc sections to me. Will this live in the E3SM docs? Or will ekat have it's own?

bartgol:
> Feel free to assign some doc sections to me. Will this live in the E3SM docs? Or will ekat have it's own?

I was thinking of adding wiki pages within this gh repo.

tcclevenger (approved):

PR #368, opened 2025-04-15
Update blake cmake settings

jgfouca:
## Motivation

Make it more convenient to run kokkos stuff on Blake.

## Testing

Tested a cmake build using these machine files.


bartgol (changes requested):
Can we change the mach file to `blake-gpu`? We may reinstate blake as a CPU mach too (since it also has Intel CPU nodes).

Also, we prob need a Blake entry in machine_specs.py, no?

jgfouca:
@bartgol , machine_specs.py is in the E3SM repo but I can change the name to blake-gpu.

bartgol:
> @bartgol , machine_specs.py is in the E3SM repo but I can change the name to blake-gpu.

Ah rats, I didn't notice the repo was EKAT :)

bartgol (approved):

PR #375, opened 2025-07-24
host_to_device/device_to_host: add support for unpacked data

jgfouca:
These functions make it convenient to ship lots of data back and forth between host pointers and device views. This can be tedious to do by hand, even if the data is unpacked (although it's even more tedious with packed, which is why this capability was added to EKAT ages ago). This PR adds support for moving around unpacked data. Since we already did support PackSize=1, which is BFB identical to Real* when just looking at the raw data, I just used reinterpret casts to pretend the incoming Views were PackSize=1. I also fixed some const correctness stuff.

### DIFF
--- a/src/pack/ekat_pack_kokkos.hpp
+++ b/src/pack/ekat_pack_kokkos.hpp
@@ -613,39 +613,39 @@ host_to_device(const std::vector<typename ViewT::value_type::scalar const*>& dat
 }
 
 // Sugar for when size is uniform (1d)
-template <typename SizeT, typename ViewT>
+template <typename ViewT>
 typename std::enable_if<IsPack<typename ViewT::value_type>::value, void>::type // void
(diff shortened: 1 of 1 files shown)


bartgol (approved):

jgfouca:
@bartgol , I discovered that I needed one more round of enhancements to cover what I want to do in gwd.

bartgol (approved):

bartgol:
@jgfouca Just so you know: we cannot update ekat in e3sm until the ekat-update PR goes in. That's b/c we are changing the organization of files and eamxx as well as mam4xx and haero all have to be updated to the new system. I am almost done with haero, then I have a PR ready to be open for mam4xx. Once all of that works, I can finally make the ekat-update PR move forward.

While you work on gw, you may need to use a branch of ekat forking off an old master, and with your host2dev branches merged in.

PR #376, opened 2025-07-29
Add all_libs target to the export list

bartgol:
## Motivation

A customer of EKAT (Haero) is linking against ekat::AllLibs, but that means that the target (ekat_all_libs) must be added to the EXPORT list.

## Testing

Tested with haero and its downstream customer mam4xx


bartgol:
The INTERFACE include directories of ekat::XYZ do not seem to propagate to ekat::AllLibs.

tcclevenger (approved):

PR #386, opened 2025-09-18
WorkspaceManager: Make constant a CPP define to make it easier to link

jgfouca:
pm-gpu + nvidia cannot link WSM in its prior form.

## Testing
Tested a branch @ndkeen gave me. It builds.


jgfouca:
I tried changing the constant to a `static inline`. That did not help. There's some weirdness going on with the constant being used as a default argument value.

bartgol:
How about a constexpr function?

bartgol:
Another thought: while at it, we could prob add a check that the factor is >0? Something down the road may crash anyways, but at least we could stop early with a meaningful message...

jgfouca:
@bartgol , good points. Done.

PR #392, opened 2025-10-22
Simplify and upgrade implementation of repack utility

bartgol:
## Motivation

I randomly came across this fcn and randomly thought of a way to greatly simplify its implementation. Furthermore, it can now also be used to pack a view of scalars, while the original impl of repack required the input view's value type to already be a Pack.

## Testing

No further testing needed.


jgfouca (approved):
Way smaller! Looks gtm

tcclevenger (approved):

PR #399, opened 2025-11-07
Some simplifications to the expression templates design

bartgol:
## Motivation

This PR cleans up a few things, making the design a bit more clear (hopefully). Namely:

- remove the [CRTP](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern) design: it wasn't really used, and it kind of complicated things in a few places, for no real gain.
- move all fcns/operators overloads to a "helper" header. The user can now include only `ekat_expression_helpers.hpp` (for expression creation) and `ekat_expression_eval.hpp` (for the evaluation fcn).
- added a macro to create binary math fcn expressions, and use it to define pow, max, min (the latter are no longer valid BinaryOp values)

## Additional Information
These mods came while I was trying to use this framework for EAMxx. Namely, I implemented the two overloads of `eamxx::Field::scale` to use expressions. One overloads scales the field by a constant, the other by another field. While for the first overload the expression implementation was faster (on both CPU and GPU), for the latter i was getting good times on CPU but terrible times on GPU. So I started to try a variety of options, and in the process did some simplifications/fixes to the API, reaching the state of this branch.

Incidentally, compared to master, this branche makes the bad impl of the Field::scale method mentioned above go about 25% faster. Although it still remain slower than the original Field::scale impl, it may point to some template code bloat in the original impl. More testing is needed for better diagnose the core issue of the GPU performance.

@tcclevenger it's probably best if you review the PR one commit at a time, so that each change becomes logically contained.

### DIFF
--- a/src/expression/ekat_expression_math.hpp
+++ b/src/expression/ekat_expression_math.hpp
@@ -22,9 +22,7 @@ namespace ekat {
                                                                                         \
     using eval_arg1_t = eval_return_t<EArg1>;                                           \
     using eval_arg2_t = eval_return_t<EArg2>;                                           \
-    using eval_t =                                                                      \
-      decltype(Kokkos::impl(std::declval<eval_arg1_t>(),                                \
-                            std::declval<eval_arg2_t>()));                              \
+    using eval_t = std::common_type_t<eval_arg1_t,eval_arg2_t>;                         \
                                                                                         \
     name##Expression (const EArg1& arg1, const EArg2& arg2)                             \
       : m_arg1(arg1)                                                                    \


src/expression/ekat_expression_binary_op.hpp
```diff
@@ -52,24 +55,27 @@ class BinaryExpression : public Expression<BinaryExpression<ELeft,ERight,OP>>{
     }
   }
 
-  static auto ret_type () { return ELeft::ret_type() + ERight::ret_type(); }
+  static auto ret_type () {
+    return std::common_type_t<ret_left,ret_right>(0);
+  }
```
tcclevenger: I'm still confused about this `ret_type` function. Is it just returning a default element? Why not just define a public type 
```
using ret_type = std::common_type_t<ret_left,ret_right>;
```
Maybe I'm missing where it's being used..

tcclevenger (approved):

src/expression/ekat_expression_binary_op.hpp
tcclevenger: Sorry, I thought it deleted my comment. That was from looking at one of the first commits lol

tcclevenger (approved):

PR #403, opened 2025-12-15
Fix a couple of constexpr issue in Units

bartgol:
## Motivation

The fcn `operator*(ScalingFactor,Units)` was not really constexpr, as there were non-constexpr fcn calls involved. Also, when creating "invalid" units, the multiplication between `ScalingFactor(-infty)` and `Units::nondimensional()` was giving overflow warnings, so I simplified the fcn.


bartgol:
The pb was indeed elsewhere...

PR #407, opened 2026-01-29
Avoid clash of temp var name in IMPL_THROW

bartgol:
## Motivation

The temp var used inside IMPL_THROW has a name that is TOO likely to shadow other var names. In particular, if someone stuffs an error msg in a stringstream var called `ss`, and calls `EKAT_ERROR_MSG(ss.str())`, the CPP will blindly expand it so that the string `msg` is replaced with `ss.str()`. But since we just declared a LOCAL `ss` var (incidentally, of the same type), shadowing the one at the macro call site, `ss.str()` evaluates to the content of the newly created stringstream, which is empty.

The solution is to use a name that is HIGHLY unlikely to clash with any var that is currently defined at the call site.

## E3SM Stakeholder Feedback

This was spotted in E3SM, as reported in E3SM-Project/E3SM#8017.

## Testing

Added a new test that checks that the user-provided message is not lost.

### DIFF
--- a/src/core/ekat_assert.hpp
+++ b/src/core/ekat_assert.hpp
@@ -48,15 +48,15 @@ void throw_exception(const std::string& msg)
 
 // Internal do not call directly.
 #define IMPL_THROW(condition, msg, exception_type)  \
-  do {                                                                \
-    if ( ! (condition) ) {                                            \
-      std::stringstream ss;                                           \
-      ss << msg;                                                      \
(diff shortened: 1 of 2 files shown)


tcclevenger (approved):

bartgol:
Manually tested on my workstation (both SERIAL and CUDA), and all tests pass. I'll merge.

PR #415, opened 2026-03-13
Use login shells in snl testing workflow

bartgol:
## Motivation

Recent changes in the SNL testing container image broke our testing, since we were using a non-login shell, which does not load all scripts in /etc/profile.d/, one of which contained the necessary instructions to load the correct env (e.g. load modules).

### DIFF
--- a/.github/workflows/ghci-snl-testing.yml
+++ b/.github/workflows/ghci-snl-testing.yml
@@ -87,7 +87,7 @@ jobs:
           submodules: recursive
       - name: Ensure CACTS is installed
         run: |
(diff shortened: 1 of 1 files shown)


bartgol:
I am patching the SNL container images so that the venv folder is world writable. While this seem like a security risk, the fact that our CI containers are ephemeral means we can allow the user to install any sort of stuff in that folder, without compromising security (the filesystem mod will all disappear when the container is stopped/removed).

I'll re-trigger testing next week (the AT2 workflows need to patch our images, and they likely already ran tonight).

PR #428, opened 2026-04-09
Adding support for Apple Silicon (without MPI)

jeff-cohere:
This PR stubs out floating point environment logic for Apple Silicon (whose floating point environment is different and complicated and possibly something to be studied in the distant future).

Apple products are supported only without MPI, because some hwloc-related features aren't supported on Apple silicon, and this can produce failures if special steps are not taken. We can follow this up with a PR that adds support for MPI on Macs if there is interest.

Accordingly, I've also disabled the EKAT comm unit test when MPI is disabled.

## Motivation
Some folks like to develop on their laptops. This allows EKAT to be built in a restricted way to support this kind of development. At the very least, mam4xx developers will be able to work on mam4xx's standalone configuration on Macs.

## Testing

The PR doesn't change any testing logic, aside from disabling an MPI-related test when MPI is disabled.

### DIFF
--- a/src/core/ekat_fpe.cpp
+++ b/src/core/ekat_fpe.cpp
@@ -3,6 +3,26 @@
 #include <cfenv>
 
 #ifndef EKAT_HAVE_FEENABLEEXCEPT
+#ifdef __APPLE__
+// Apple Silicon has a very different floating point environment, which we leave for
+// future exploration.
+
+inline int fegetexcept (void)
+{
+  return 0;
+}
+
+inline int feenableexcept (int excepts)
+{
+  return 0;
+}
+
+inline int fedisableexcept (int excepts)
+{
+  return 0;
+}
+
+#else
 
 // Drop-in replacement of some useful GNU utils (needed on Apple platforms)
 inline int fegetexcept (void)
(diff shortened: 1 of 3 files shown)


src/core/ekat_fpe.cpp
```diff
@@ -3,6 +3,26 @@
...
+}
+
+inline int fedisableexcept (int excepts)
+{
+  return 0;
+}
+
+#else
 
 // Drop-in replacement of some useful GNU utils (needed on Apple platforms)
```
jeff-cohere, line 27: It occurs to me that this logic was intended for Intel Macs, which are pretty old at this point. We could delete this code block if it's not used for anything else.

bartgol (approved):
@jeff-cohere If supporting APPLE devices is important, we can add a job of the cpu-gcc workflow that runs on macos-latest...

jeff-cohere:
Thanks, Luca. I'm not sure how to add `macos-latest` to the existing GitHub workflow, but even without it we'll have people testing the Mac builds with mam4xx soon. Is it okay to merge, or do you think we should add Mac builds to our GitHub CI first?

bartgol:
Since APPLE is not for production, it's prob ok to live without a CI. If someone is _really_ invested in ekat to work on APPLE, they can submit a PR with the new workflow...

PR #429, opened 2026-04-10
Standardize package found logic for tpls

bartgol:
## Motivation

As shown in the failed CI runs for `eamxx-v1` workflow in E3SM-Project/E3SM#8258, there is an issue with how we decide whether to fetch and install yaml-cpp. Namely, if the pkg is found, and only provides the (nicely) scoped `yaml-cpp::yaml-cpp` target, the if condition that establish whether to call FetchContent is wrong. This PR changes it to be more robust, for all the three packages: ensure that <pkg>_FOUND is always set, and use that to decide whether to use FetchContent.

Side fix: if the yaml-cpp pkg is found but does not define the scoped tgt, we create the scoped one ourselves. But we were not propagating this to the installed EkatConfig.cmake, so downstream tgt would not know what yaml-cpp::yaml-cpp was (again, only for the case where a pre-installed yaml-cpp did NOT define the scoped tgt). To fix this, I added the same logic to our EkatConfig.cmake.in file, so that the scoped target is guaranteed to exist for downstream customers.


src/logging/CMakeLists.txt
```diff
@@ -12,9 +12,11 @@ if (NOT EKAT_SKIP_FIND_SPDLOG)
   else()
     message (STATUS "Looking for spdlog ... NOT FOUND")
   endif()
+else ()
+  set (spdlog_FOUND FALSE)
 endif()
 
-if (NOT TARGET spdlog)
```
bartgol, line 17: This was also a bit subtle: the found pkg may as well only define `spdlog::spdlog` (which is what we link against anyways). So we may have found spdlog, but this if statement would proceed as if we had not...

bartgol:
The last version should be extra-extra safe:

1. check if a target (scoped/unscoped) exists. If so, use that
2. if no tgt exists, use find_package
3. if either of the above succeeds, ensure scoped tgt is defined. If not, create an alias scoped tgt (pointing to the unscoped one)
4. if tgt or pkg was not found, proceed to use FetchContent.
5. in EkatConfig.cmake.in, replicate similar logic, to ensure we honor any existing tgt in the host project.

src/parser/CMakeLists.txt
```diff
@@ -17,19 +23,22 @@ if (NOT EKAT_SKIP_FIND_YAML_CPP)
...
+else ()
+  set (yaml-cpp_FOUND FALSE)
+endif()
+
+# If we found yaml-cpp (via find_package or as an existing tgt)
+# ensure it provides the scoped target
+if (yaml-cpp_FOUND)
+  if (TARGET yaml-cpp AND NOT TARGET yaml-cpp::yaml-cpp)
+    add_library(yaml-cpp::yaml-cpp ALIAS yaml-cpp)
+  endif()
```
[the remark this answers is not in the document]
bartgol: This is starting to look a lot like having to write our own FindYamlCpp.cmake... Gemini seems to agree with me:

> Copilot is technically correct, but it’s suffering from "Perfectionist's Bloat."
> In the world of CMake, there is a point of diminishing returns where you transition from "robust build system" to "writing a full-blown package manager." Copilot wants you to reach the latter.

The current impl would work for 99.99% of the cases. I think we're safe enough for now. If someone loses an hour of work b/c this issue fires back, I'll gladly buy them a few rounds of drinks to make up for it.

bartgol:
I built e3sm with this version of ekat on my workstation, and it builds fine.

src/parser/CMakeLists.txt
```diff
@@ -17,19 +23,22 @@ if (NOT EKAT_SKIP_FIND_YAML_CPP)
   if (yaml-cpp_FOUND)
     message (STATUS "Looking for yaml-cpp ... FOUND")
     message (STATUS "  yaml-cpp_DIR: ${yaml-cpp_DIR}")
-    # It is possible that the installation provides the yaml-cpp target, but not the yaml-cpp::yaml-cpp target. If so, define the alias target
-    if (NOT TARGET yaml-cpp::yaml-cpp)
-      if (NOT TARGET yaml-cpp)
-        message (FATAL_ERROR "Neither yaml-cpp nor yaml-cpp::yaml-cpp targets were found")
-      endif()
-      add_library (yaml-cpp::yaml-cpp ALIAS yaml-cpp)
+    if (NOT TARGET yaml-cpp AND NOT TARGET yaml-cpp::yaml-cpp)
```
jeff-cohere: What does this `if` test do?

jeff-cohere (dismissed):
I have one question about an empty `if` block. But the rest looks good to me.

src/parser/CMakeLists.txt
bartgol: whoops! good catch. Will remove.

jgfouca (approved):

PR #430, opened 2026-04-13
Add common angle/solid-angle units to units namespace

bartgol:
## Motivation

`rad` is used a lot in E3SM, and users need to inject code like
```cpp
Units rad (ekat::units::Units::nondimensional(),"rad")
```
which is not super-clear. It'd be clearer to just use `ekat::units::rad` directly.

Also, added a `rename` method, so one can do
```cpp
auto m2 = (m*m).rename("m2");
```
which I think it's clearer (in intent) than
```cpp
Units m2(m*m,"m2");
```
Finally, i removed `dyn` and `rem`, as they are non-SI units, but unlike `g` (which is used) they seem to be quite niche.

### DIFF
--- a/src/core/ekat_units.hpp
+++ b/src/core/ekat_units.hpp
@@ -98,6 +98,7 @@ public:
 
   constexpr Units& operator= (const Units&) = default;
 
+  [[deprecated("Use the 'ekat::units::none' constant instead.")]]
   static constexpr Units nondimensional () {
     Units u(ScalingFactor::one());
     u.m_string_repr[0] = '1';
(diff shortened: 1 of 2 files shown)


bartgol:
This PR turned into a more general cleanup of the Units class and units namespace:

- add `rad`, `sr`, and `none` constants
- deprecate `Units::nondimensional()` (use constant instead.
- deprecated `rem` and `dyn` (non-SI and VERY niche). Btw, dyn's numerical value was wrong, fixed it.
- remove global array of basic units symbols
- use string_view when possible
- added `is_scaled` (parallel method to `is_dimensionless`).

mahf708:
shouldn't the derived units belong to the downstream app/customer?

bartgol:
> shouldn't the derived units belong to the downstream app/customer?

That's one way to approach the issue. In general, yes, but I find it "reasonable" to provide some "common" ones (like N, Pa, J, W, etc).

mahf708 (approved):

PR #432, opened 2026-04-16
Improve fetch content utility for parallel builds

bartgol:
## Motivation

The current logic was still not robust enough in parallel builds. The core issue (after digging with a bot) seemed to be the fact that, while we did skip the clone in FC's MakeAvailable by setting `FETCHCONTENT_SOURCE_DIR_<PKG>` for bulds 2,3,..., we were still calling `FetchContent_MakeAvailable`, which does end up touching more files, and messing with the generation phase (somehow).

The new impl only uses the Declare/Populate features of FC, and does NOT use MakeAvailable, but rather simply adds the subdirectory.

### DIFF
--- a/cmake/EkatUtils.cmake
+++ b/cmake/EkatUtils.cmake
@@ -148,37 +148,90 @@ function(separate_cut_arguments prefix options oneValueArgs multiValueArgs retur
 endfunction(separate_cut_arguments)
 
 # Utility to avoid race conditions in FetchContent_MakeAvailable
-# when 2+ builds are configuring at the same time
-include(FetchContent)
-function(ekat_make_available NAME GIT_TAG_VALUE)
+# when 2+ builds are configuring at the same time. Only the 1st one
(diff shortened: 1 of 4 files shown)


bartgol:
I think this is now in a very good state after some copilot iterations. Summary

- E3SM now builds correctly the e3sm_atm_developer suite (in parallel) on mappy
- Move the whole fetch content logic inside a cmake utility function
- Make the fcn use `git rev-parse` to check ref of currently checked out tpl in tpl source dir (if non-empty)
- Use FetchContent only to declare/populate the folder, and manually use `add_subdirectory` if/when the src dir is populated (this seemed to be key to avoid subtle fs race conditions in MakeAvailable)

jeff-cohere (approved):
Well, it's cool that CMake gives you the elements you need to implement your own version of `FetchContent_MakeAvailable`! This is quite involved, but it just might work! Worth a shot, anyway.

PR #435, opened 2026-04-20
Resolve expired storage issue in Units

bartgol:
## Motivation

The last two PRs (that were attempting to fix the error that certain Intel compilers give) completely misunderstood the core of the issue. After another back-n-forth with a bot, I finally understood the core issue (hopefully), and this PR should address it.

Namely, the utility method that concats string representations was creating a temporary, and we are not allowed to access that temporary in a ctor body, as it is technically expired by then (for constexpr evaluations), while it is still alive while the initializer list is parsed.


bartgol:
We've manually tested this version with e3sm/next on compy, and the build error finally goes away!

jeff-cohere (approved):
Nice find!

PR #436, opened 2026-04-29
Adding support for kokkos-tools

jeff-cohere:
## Motivation

At the 2026 ALCF hackathon, I've been using [kokkos-tools](https://github.com/kokkos/kokkos-tools) (specifically the [simple kernel timer](https://github.com/kokkos/kokkos-tools/tree/develop/profiling/simple-kernel-timer)) to understand performance bottlenecks in mam4xx. Kokkos-tools is easy to build and takes advantage of Kokkos instrumentation already compiled into an application, but it's a bit inconvenient to use because you have to set an environment variable to use it.

This PR introduces kokkos-tools as an optional feature for EKAT, and sets the environment variable appropriately in its bundled test session and catch2 main.

## Testing

This is a build-system-only change--no new code features are added. I'm testing this in its feature branch at the hackathon. When I've verified that it works to my satisfaction, I'll mark this PR as ready for review.

## Addenda

Apparently, kokkos-tools needs C++20, so this PR depends on #400.


jeff-cohere:
I guess now we have another reason to move to C++20:

```
CMake Error at /home/jnjohnson/hackathon/E3SM/externals/ekat/extern/kokkos-tools/CMakeLists.txt:12 (message):
  KokkosTools requires C++20
```

src/testing-support/CMakeLists.txt
```diff
@@ -34,17 +52,23 @@ else()
...
+# Write out a configuration header for testing support.
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/ekat_testing_support.hpp.in
+               ${CMAKE_CURRENT_BINARY_DIR}/ekat_testing_support.hpp)
+
 # A small lib for Catch2 main executable, so we don't rebuild it
 # for every single unit test
 add_library(ekat_catchmain ekat_catch_main.cpp)
 target_link_libraries(ekat_catchmain PUBLIC ekat::Core)
 target_link_libraries(ekat_catchmain PUBLIC Catch2::Catch2)
+target_link_libraries(ekat_catchmain PUBLIC ekat_testsession)
```
bartgol: We cannot link ekat_testsession to ekat_catchmain. EKAT's customers may ship their own test session (a cpp file where the fcns `ekat_initialize_test_session(int,char**,bool)` and `ekat_finalize_test_session()` are defined), where they may do different things to init/finalize the testing session.

src/testing-support/CMakeLists.txt
```diff
@@ -34,17 +52,23 @@ else()
   set (EKAT_BUILDS_CATCH2 OFF CACHE BOOL "Whether Ekat builds Catch2" FORCE)
 endif()
 
+# Write out a configuration header for testing support.
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/ekat_testing_support.hpp.in
```
bartgol: Instead of a header to configure, how about calling `target_compile_definitions(..)` on the test session and catch main targets (perhaps even with PRIVATE visibility)? It seems easier and more scoped than having to configure, ship, and include a header...

bartgol (changes requested):
My biggest concern is linking the test session tgt to the catchmain tgt. I think we should avoid that. The other comment may be more debatable..

src/testing-support/CMakeLists.txt
jeff-cohere: Yeah, I did this to alleviate a linking error on my Mac, because Apple is very particular about chasing down all dependencies in dynamically linked libraries, which is not as important as what you're mentioning.

src/testing-support/CMakeLists.txt
jeff-cohere: Very good idea.

bartgol (approved):

PR #442, opened 2026-05-11
Use an alias for Kokkos' default view memory trait

tcclevenger:
Add an alias for Kokkos' default memory trait (managed memory, no other specializations) since Kokkos deprecated `using MemoryManaged = MemoryTraits<0>`.

### DIFF
--- a/src/kokkos/ekat_kokkos_types.hpp
+++ b/src/kokkos/ekat_kokkos_types.hpp
@@ -44,6 +44,9 @@ using DefaultDevice = Kokkos::Device<Kokkos::DefaultExecutionSpace, Kokkos::Defa
 // A device type to force host execution
 using HostDevice = Kokkos::Device<Kokkos::DefaultHostExecutionSpace, Kokkos::DefaultHostExecutionSpace::memory_space>;
 
+// Alias for Kokkos' default memory trait for views (managed memory, no special properties)
+using ManagedMemoryTrait = Kokkos::MemoryTraits<0>;
+
 template<typename DT, typename... Props>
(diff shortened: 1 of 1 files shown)


src/kokkos/ekat_kokkos_types.hpp
```diff
@@ -44,6 +44,9 @@ using DefaultDevice = Kokkos::Device<Kokkos::DefaultExecutionSpace, Kokkos::Defa
 // A device type to force host execution
 using HostDevice = Kokkos::Device<Kokkos::DefaultHostExecutionSpace, Kokkos::DefaultHostExecutionSpace::memory_space>;
 
+// Alias for Kokkos' default memory trait for views (managed memory, no special properties)
+using ManagedMemoryTrait = Kokkos::MemoryTraits<0>;
```
tcclevenger, line 48: @bartgol I didn't add this to KokkosTypes since it isn't device dependent. Also, I added "Trait" since I felt "MemoryManaged" wasn't very descriptive. I also considered `KokkosManagedMemoryTrait` or even `KokkosDefaultMemoryTrait`. Not sure if you have opinions on these.

bartgol: I think this is fine. I hope this won't get used much anyways in ekat and customers (in favor of KokkosTypes' typedefs), so this is fine.

Does kokkos still have the unmanaged trait typedef in 5.x?

bartgol (approved):

src/kokkos/ekat_kokkos_types.hpp
tcclevenger: Yes, Unmanaged still exists

PR #444, opened 2026-05-14
Upgrade error macros to handle throw inside stack unwind

bartgol:
## Motivation

When an exception is thrown while the stack is being unwound, the runtime calls std::terminate. This causes the stack (and message) from the original exception to be lost, making debug hard. For instance, during stack unwind, some structures may be cleaned in a way that does not meet "usual" requirements (e.g., all IO files have all been closed). But this should NOT prevent the original error from reaching the screen. The building is on fire, so who cares if a window broke?

This PR adds some logic to the internal `throw_exception` routine, that prevents the exception from being thrown if we can detect that there is already an uncaught exception. It also prints a warning with the curr exception msg (without the stacktrace), but prevents more exception to be printed, to keep the screen poliution to a minimum. The idea is "ok, we're likely about to crash, so I'll let you know ONCE that other stuff went south, just so you know, but won't interfere with the current execution".

## E3SM Stakeholder Feedback

I think this will drastically improve debugging experience in E3SM, where the typical error msg in the logs was
```
[EAMxx] Finalize ...
terminate called after throwing an instance of 'std::runtime_error'
  what():  Error! ScorpioSession::finalize called, but a file is still in use elsewhere.
 - filename: /home/lbertag/workdir/e3sm/e3sm-data/inputdata/atm/cam/topo/USGS-gtopo30_ne2np4pg2_x6t_20230331.nc

FAILED CONDITION: 'it.second.num_customers==0'
```
which is thrown by the shutdown of the scorpio session which we do during the driver cleanup, and which prevented users from seeing the REAL issue that sparked the first exception.

## Testing

I added a unit test, and also verified in eamxx that with this branch I _do_ see the original error (instead of the secondary one) when the code crashes.

### DIFF
--- a/src/core/ekat_assert.hpp
+++ b/src/core/ekat_assert.hpp
@@ -3,6 +3,7 @@
 
 #include <sstream>
 #include <iostream>
+#include <exception>  // For std::uncaught_exceptions
 #include <stdexcept>  // For std::runtime_error
 #include <type_traits>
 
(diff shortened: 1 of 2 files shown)


jgfouca (approved):
Looks great!

PR #448, opened 2026-06-01
Add support for testing on oneapi ghci-snl runner

bartgol:
## Motivation

We need to be able to test with oneapi compilers.

## E3SM Stakeholder Feedback

This needs to be integrated before we can enable oneapi testing in E3SM.

### DIFF
--- a/bin/test-launcher
+++ b/bin/test-launcher
@@ -159,7 +159,6 @@ def run_test(buffer_output,print_omp_affinity,exec_args):
             # variable TEST_LAUNCHER_ON_GPU to either True or False
             if ${TEST_LAUNCHER_ON_GPU}:
                 expect (len(my_ids)==1, "Error! For GPU runs, each rank should use only one device.")
-                exec_args.append(" --kokkos-device-id={}".format(ids[0]))
             else:
                 # Start placing on the correct cpu set
                 cmd += " taskset -c " + ",".join(my_ids)
--- a/cacts.yaml
+++ b/cacts.yaml
@@ -76,6 +76,8 @@ machines:
 
   ghci-snl-cpu: {}
 
+  ghci-snl-oneapi: {}
+
   ghci-snl-cuda:
     gpu_arch: "cuda"
     num_run_res: 2
--- a/cacts.yaml
+++ b/cacts.yaml
@@ -37,6 +37,8 @@
 
 project:
   name: EKAT
+  ctest_error_exceptions: ["error_handler"]
+
   # NOTE: CACTS will also set project.root_dir at runtime, so you can actually use
   # ${project.root_dir} in the machines/configurations sections
 
--- a/cacts.yaml
+++ b/cacts.yaml
@@ -98,8 +98,6 @@ configurations:
     cmake_args:
       CMAKE_CXX_STANDARD: 20
       EKAT_ENABLE_ALL_PACKAGES: True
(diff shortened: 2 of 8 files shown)


cacts.yaml
```diff
@@ -94,8 +98,6 @@ configurations:
     cmake_args:
       CMAKE_CXX_STANDARD: 20
       EKAT_ENABLE_ALL_PACKAGES: True
-      EKAT_TEST_THREAD_INC: ${2 if machine.gpu_arch is None else 1}
```
bartgol, line 97: Removing these exposed some cascade bugs.

1. the GPU tests were getting omp16 threads, since EKAT_ENABLE_GPU was NOT set in the cache, so downstream code would not see it
2. fixing EKAT_ENABLE_GPU exposed a bug in test-launcher. Since EKAT_ENABLE_GPU was not set in the cache, when configuring the launcher we were NOT setting `TEST_LAUNCHER_ON_GPU=ON`, so an if statement was never triggering. After fixing EKAT_ENABLE_GPU, it did trigger, revealing we were still attempting to add the flag `--kokkos-device-id`, which was removed a while ago.

bin/test-launcher
```diff
@@ -159,7 +159,6 @@ def run_test(buffer_output,print_omp_affinity,exec_args):
             # variable TEST_LAUNCHER_ON_GPU to either True or False
             if ${TEST_LAUNCHER_ON_GPU}:
```
bartgol, line 160: This if statement does very little now. I am tempted to remove it altogether, which would simplify a bit the logic here, as well as the cmake logic in the repo.

tcclevenger: Yeah, I support getting rid of it.

cmake/machine-files/ghci-snl-cpu.cmake
```diff
@@ -3,3 +3,4 @@ include(${CMAKE_CURRENT_LIST_DIR}/ghci-snl.cmake)
 
 # Currently, we have 32 cores for each ghci-snl-cpu instance, but 4 ranks is enough
 set(EKAT_TEST_MAX_RANKS 4 CACHE STRING "Upper limit on ranks for mpi tests")
+set(EKAT_TEST_MAX_THREADS 8 CACHE STRING "Upper limit on ranks for mpi tests")
```
tcclevenger: I think this comment needs to be updated?

cmake/machine-files/ghci-snl-oneapi.cmake
```diff
@@ -0,0 +1,9 @@
+# Common settings for our ghci images
+include(${CMAKE_CURRENT_LIST_DIR}/ghci-snl.cmake)
+
+# Set kokkos arch
+option (Kokkos_ARCH_SPR "" ON)
+
+# Currently, we have 192 cores on blake, but testing 4 ranks is probably enough
+set(EKAT_TEST_MAX_RANKS 4 CACHE STRING "Upper limit on ranks for mpi tests")
+set(EKAT_TEST_MAX_THREADS 8 CACHE STRING "Upper limit on ranks for mpi tests")
```
tcclevenger: Same comment about the comment.

tcclevenger (dismissed):

bin/test-launcher
bartgol: It'll be a follow up one then.

PR #453, opened 2026-06-13
Fix issue when using zip to modify containers

bartgol:
## Motivation

The `zip` function returns a temporary object which is later dereferenced by `operator*` into a temporary tuple. As such, using `auto&` violates C++ rules, as it binds an lvalue ref (the lhs tuple) to an rvalue ref (the rhs tuple). The gold standard solution is to use `auto&&`, together with a minor modification to the `operator*` implementation (which is what happens here).

 The key mod is to ensure that the reference in the tuple is forwarded correctly to the resulting type, as `make_tuple` and the plain `auto` return type would silently cause to return a tuple of T's rather than a tuple of T&'s.
 
 More details: `forward_as_tuple` allows the tuple created in the lambda inside `operator*` to return an actual tuple of T& rather than a tuple of T. The decltype in the function return type prevents this to be (quite misteriously, I agree) decay the generated tuple<T&>&& to tuple<T>. Why I don't quite grasp this last bit myself, I followed gemini suggestion, and verified that without `decltype(auto)`, some aptly crafted containers _can_ fail to return a tuple of refs.

## Testing

I verified that this fixes the following code that did not compile in eamxx:
```
for (auto& [name,dim] : ekat::zip(names,dims)) {...}
```
where `names` and `dims` where nonconst std vectors. Both the `auto&&` mod in the customer code and this PR are needed to fix the compiler error.

## Additional Information

This is ONLY needed for loops using non-const structured binding. Usage along the line of `for (const auto& [a,b] : ekat::zip(C1,C2))` was already correct (standard compliant, not just "gcc-lucky").

### DIFF
--- a/tests/algorithm/CMakeLists.txt
+++ b/tests/algorithm/CMakeLists.txt
@@ -5,14 +5,12 @@ if (EKAT_TEST_DOUBLE_PRECISION)
   EkatCreateUnitTest(lin_interp${DP_POSTFIX}
     SOURCES lin_interp_test.cpp
     LIBS ekat::Algorithm
-    COMPILER_DEFS EKAT_TEST_DOUBLE_PRECISION
     THREADS 1 ${EKAT_TEST_MAX_THREADS} ${EKAT_TEST_THREAD_INC})
 endif()
 if (EKAT_TEST_SINGLE_PRECISION)
   EkatCreateUnitTest(lin_interp${SP_POSTFIX}
     SOURCES lin_interp_test.cpp
     LIBS ekat::Algorithm
-    COMPILER_DEFS EKAT_TEST_SINGLE_PRECISION
     THREADS 1 ${EKAT_TEST_MAX_THREADS} ${EKAT_TEST_THREAD_INC})
 endif()
 
@@ -23,7 +21,6 @@ if (EKAT_TEST_DOUBLE_PRECISION)
             tridiag_tests_correctness.cpp
             tridiag_tests_performance.cpp
     LIBS ekat::Algorithm
-    COMPILER_DEFS EKAT_TEST_DOUBLE_PRECISION
     THREADS ${EKAT_TEST_MAX_THREADS}
     EXCLUDE_MAIN_CPP)
 endif()
@@ -33,7 +30,6 @@ if (EKAT_TEST_SINGLE_PRECISION)
             tridiag_tests_correctness.cpp
             tridiag_tests_performance.cpp
     LIBS ekat::Algorithm
-    COMPILER_DEFS EKAT_TEST_SINGLE_PRECISION
     THREADS ${EKAT_TEST_MAX_THREADS}
     EXCLUDE_MAIN_CPP)
 endif()
@@ -57,12 +53,10 @@ endif()
 if (EKAT_TEST_DOUBLE_PRECISION)
   EkatCreateUnitTest(reduction${DP_POSTFIX}
     SOURCES reduction_tests.cpp
-    COMPILER_DEFS EKAT_TEST_DOUBLE_PRECISION
     LIBS ekat::Algorithm)
 endif()
 if (EKAT_TEST_SINGLE_PRECISION)
   EkatCreateUnitTest(reduction${SP_POSTFIX}
     SOURCES reduction_tests.cpp
-    COMPILER_DEFS EKAT_TEST_SINGLE_PRECISION
     LIBS ekat::Algorithm)
 endif()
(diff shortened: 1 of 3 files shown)


tests/algorithm/CMakeLists.txt
```diff
@@ -5,14 +5,12 @@ if (EKAT_TEST_DOUBLE_PRECISION)
   EkatCreateUnitTest(lin_interp${DP_POSTFIX}
     SOURCES lin_interp_test.cpp
     LIBS ekat::Algorithm
-    COMPILER_DEFS EKAT_TEST_DOUBLE_PRECISION
```
jeff-cohere, line 8: Any relevant significance to removing these definitions? Or is this just code cleanup?

jeff-cohere: Ah, I guess you're just addressing a compiler warning. :-)

jeff-cohere (approved):

tests/algorithm/CMakeLists.txt
bartgol: yeah, I think the warning is what caused oneapi build to fail (CTest looks for certain strings in the build log output and deems the build a fail if it finds them). I'm not 100% sure, but I figured I fixed them and see what happens...

bartgol: Looks like the build passed now, so maybe that was it..

bartgol: It did fix it. And for fun, I asked gemini to find out what triggered it. The ctest regex list for errors is:
```cpp
static char const* cmCTestErrorMatches[] = {
  "^[Bb]us [Ee]rror", // noqa: spellcheck disable-line
  "^[Ss]egmentation [Vv]iolation",
  "^[Ss]egmentation [Ff]ault",
  ":.*[Pp]ermission [Dd]enied",
  "([^ :]+):([0-9]+): ([^ \\t])",
  "([^:]+): error[ \\t]*[0-9]+[ \\t]*:",
  "^Error ([0-9]+):",
  "^Fatal",
  "^Error: ",
  "^Error ",
  "[0-9] ERROR: ",
  R"(^"[^"]+", line [0-9]+: [^Ww])",
  "^cc[^C]*CC: ERROR File = ([^,]+), Line = ([0-9]+)",
  "^ld([^:])*:([ \\t])*ERROR([^:])*:",
  R"(^ild:([ \t])*\(undefined symbol\))",
  "([^ :]+) : (error|fatal error|catastrophic error)",
  "([^:]+): (Error:|error|undefined reference|multiply defined)",
  R"(([^:]+)\(([^\)]+)\) ?: (error|fatal error|catastrophic error))",
  "^fatal error C[0-9]+:",
  ": syntax error ",
  "^collect2: ld returned 1 exit status",
  "ld terminated with signal",
  "Unsatisfied symbol",
  "^Unresolved:",
  "Undefined symbol",
  "^Undefined[ \\t]+first referenced",
  "^CMake Error.*:",
  ":[ \\t]cannot find",
  ":[ \\t]can't find",
  R"(: \*\*\* No rule to make target [`'].*\'.  Stop)",
  R"(: \*\*\* No targets specified and no makefile found)",
  ": Invalid loader fixup for symbol",
  ": Invalid fixups exist",
  ": Can't find library for",
  ": internal link edit command failed",
  ": Unrecognized option [`'].*\\'",
  R"(", line [0-9]+\.[0-9]+: [0-9]+-[0-9]+ \([^WI]\))",
  "ld: 0706-006 Cannot find or open library file: -l ",
  "ild: \\(argument error\\) can't find library argument ::",
  "^could not be found and will not be loaded.",
  "s:616 string too big",
  "make: Fatal error: ",
  "ld: 0711-993 Error occurred while writing to the output file:",
  "ld: fatal: ",
  "final link failed:",
  R"(make: \*\*\*.*Error)",
  R"(make\[.*\]: \*\*\*.*Error)",
  R"(\*\*\* Error code)",
  "nternal error:",
  R"(Makefile:[0-9]+: \*\*\* .*  Stop\.)",
  ": No such file or directory",
  ": Invalid argument",
  "^The project cannot be built\\.",
  "^\\[ERROR\\]",
  "^Command .* failed with exit code",
  "lcc: \"([^\"]+)\", (line|строка) ([0-9]+): (error|ошибка)",
  nullptr
};
```
The one that triggered is `"([^ :]+):([0-9]+): ([^ \\t])"`, which means "any non-space and non-colon char, followed by :, followed by digits, followed by :, followed by a space, followed by any nonspace char". The line
```
<command line>:5:9: note: previous definition is here
```
matches it, when you consider the portion `5:9: n`.

Fun times.

PR #457, opened 2026-07-14
Fix runner labels for cuda/oneapi jobs

bartgol:
## Motivation

Our runners got updated a few weeks ago, and their labels changed a bit. The CUDA/OneAPI jobs in our workflow were not updated accordingly. This should fix it.

@tcclevenger Once this is merged, you can rebase your C++20 PR on master, and it should finally get tested on those runners too.


tcclevenger (approved):
Awesome!

PR #459, opened 2026-07-31
Fix Pack copy constructors: remove constexpr for GCC v14+ OpenMP simd issue

ndkeen:
GCC 14 and later reject OpenMP directives (#pragma omp simd, used by vector_simd) inside constexpr function bodies, whereas GCC 13 and earlier accepted this. Drop constexpr from the three Pack copy constructors in ekat_pack.hpp that contain vector_simd loops so that EKAT builds cleanly with newer GCC versions.

### DIFF
--- a/src/pack/ekat_pack.hpp
+++ b/src/pack/ekat_pack.hpp
@@ -240,21 +240,26 @@ struct Pack {
   }
 
   // Init this Pack from another one.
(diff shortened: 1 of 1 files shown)


bartgol (approved):
W need to verify that we did not use Pack inside any constant expression. If so, we need some c++20 trickery to keep the best of both worldd. However, I think for now it's safe to do these changes.

